PR #4, opened 2016-11-22
Remove v0 workflow instances API

rouzwawi:
This removes all references to the Bigtable `execution_info` table, including v0 Workflows `/instances` API endpoints which still used the old executions table.


bergman:
👍

PR #10, opened 2016-11-30
Introduce Queued State

rouzwawi:
This is mainly a refactor of existing behaviour into a setup that makes it easier to implement further scheduling behaviour.

The change also renames the `AWAITING_RETRY` state to a more generic `QUEUED` state.

```
> STYX_CLI_HOST=localhost:8080 styx ls -c example-component
  WORKFLOW INSTANCE    STATE            LAST EXECUTION ID                               PREVIOUS EXECUTION INFO

example-component bihourly-job
  2016-12-01           QUEUED           styx-run-bc83472a-f4e8-4bab-a86a-070bf6649054   Exit code: 20
  2016-12-02           QUEUED           styx-run-212ef3f0-8ae3-4f56-b969-75319ef93cda   Exit code: 20
  2016-12-03           QUEUED           styx-run-6a03f8de-8b69-42e0-8615-d00371ae5afe   Exit code: 20
```


styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -212,7 +216,7 @@ public StyxScheduler build() {
           scheduleSources,
           statsFactory,
           executorFactory,
-          publisherFactory);
+          publisherFactory, retryUtil);
```
fabriziodemaria: Should have its own new line :)

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
```diff
@@ -87,22 +77,11 @@ private void checkRetry(RunState state) {
...
     }
   }
-
-  private Duration calculateDelay(RunState state) {
-    final int tries = (state.data().tries() < maxExponent) ? state.data().tries() : maxExponent;
-    final int multiplier = Math.max(1, RANDOM.nextInt(1 << tries));
-    final Duration delay = baseDelay.multipliedBy(multiplier);
-
-    final String instanceKey = state.workflowInstance().toKey();
-    LOG.info("{} scheduling retry #{} in {}", instanceKey, state.data().tries(), delay);
```
fabriziodemaria: Is it intended to leave out this logging?

styx-scheduler-service/src/main/java/com/spotify/styx/state/StateManager.java
```diff
@@ -58,6 +59,11 @@
   void receive(Event event) throws IsClosed;
 
   /**
+   * Get a map of all active {@link WorkflowInstance} states.
+   */
+  ImmutableMap<WorkflowInstance, RunState> activeStates();
```
bergman: Why ImmutableMap and not just Map? Would anyone think they can mutate the return value of this and somehow that would be persisted?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -0,0 +1,106 @@
...
+
+    final Instant now = time.get();
+    final Instant deadline = Instant
+        .ofEpochMilli(runState.timestamp())
+        .plus(ttls.ttlOf(runState.state()));
+
+    return isAtOrAfter(now, deadline);
+  }
+
+  private boolean isAtOrAfter(Instant now, Instant deadline) {
```
bergman: I find this function confusing, I'm not sure which value is at or after compared to the other and isn't this the same as `!foo.isAfter(bar)`?

bergman:
LGTM so far but I guess we need to redirect the old events (`triggerExec`, `submit`) to this `QUEUED` state?

styx-common/src/main/java/com/spotify/styx/state/RunState.java
```diff
@@ -298,7 +306,7 @@ public RunState retry(WorkflowInstance workflowInstance) {
       switch (state()) {
         case TERMINATED: // for backwards compatibility
         case FAILED:     // for backwards compatibility
-        case AWAITING_RETRY:
+        case QUEUED:
```
fabriziodemaria: `// for backwards compatibility`

rouzwawi:
@bergman I've deferred redirecting `triggerExecution` and `submit` into `QUEUED` to se later PR.

styx-common/src/main/java/com/spotify/styx/state/RunState.java
```diff
@@ -286,7 +290,11 @@ public RunState retryAfter(WorkflowInstance workflowInstance, long delayMillis)
       switch (state()) {
         case TERMINATED:
         case FAILED:
```
rouzwawi: I removed the `enqueue` and `dequeue` events from this PR, so it's mainly focused on just introducing the `QUEUED` state.

bergman:
👍

PR #16, opened 2016-12-03
Increase local file watcher sensitivity

rouzwawi:
This reduces latency for detecting changes, which also reduces the
run-time for the tests from about 50 seconds to 10 seconds (on Mac OSX).


bergman:
👍

PR #18, opened 2016-12-04
Implement offset/limit for workflow instance executions

rouzwawi:


styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
```diff
@@ -107,29 +109,46 @@ void writeEvent(SequenceEvent sequenceEvent) throws IOException {
...
+          break;
+        }
+
+        result = scanner.next();
+      }
     }
-    workflowInstanceDataList.sort(WorkflowInstanceExecutionData.COMPARATOR);
 
-    return workflowInstanceDataList;
+    return workflowInstancesSet.parallelStream()
```
bergman: did the `parallelStream()` make any discernible difference?

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -196,12 +197,18 @@ public WorkflowResource(Storage storage) {
...
 
-  private Response<List<WorkflowInstanceExecutionData>> instances(String componentId, String endpointId) {
+  private Response<List<WorkflowInstanceExecutionData>> instances(
+      Request request,
+      String componentId,
+      String endpointId) {
+
     final WorkflowId workflowId = WorkflowId.create(componentId, endpointId);
-    final List<WorkflowInstanceExecutionData> data;
+    final String offset = request.parameter("offset").orElse("");
```
bergman: This is just semantics and I can't really find anything useful on google right now, but isn't this more of a `cursor` than an `offset`? At least in the SQL sense, an `offset` to me is the number of items to skip while this is the element to start from.

styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
```diff
@@ -48,8 +56,12 @@
 import com.spotify.styx.model.Workflow;
 import com.spotify.styx.model.WorkflowInstance;
 import com.spotify.styx.model.WorkflowState;
-import com.spotify.styx.storage.InMemStorage;
```
bergman: 🎉

styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
```diff
@@ -382,6 +426,26 @@ public void shouldReturnWorkflowInstanceData() throws Exception {
     assertJson(response, "triggers.[0].executions.[0].statuses.[1].timestamp", is("2016-08-10T07:00:02Z"));
   }
 
+  @Test
+  public void shouldPaginateWorkflowInstancesData() throws Exception {
```
bergman: Could you add a test where the `offset` parameter is set as well? I kind of get the feeling that there could be an off-by-one error in there.

```java
scan.setStartRow(Bytes.toBytes(offsetInstance.toKey() + '#'));
```

styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
```diff
@@ -107,29 +109,46 @@ void writeEvent(SequenceEvent sequenceEvent) throws IOException {
...
 
-    final Set<WorkflowInstance> workflowInstancesSet = Sets.newHashSet();
-    for (Result result : eventsTable.getScanner(scan)) {
-      final String key = new String(result.getRow());
-      final int lastHash = key.lastIndexOf('#');
-      final WorkflowInstance wfi = WorkflowInstance.parseKey(key.substring(0, lastHash));
-      workflowInstancesSet.add(wfi);
+    if (!Strings.isNullOrEmpty(offset)) {
+      final WorkflowInstance offsetInstance = WorkflowInstance.create(workflowId, offset);
+      scan.setStartRow(Bytes.toBytes(offsetInstance.toKey() + '#'));
```
bergman: Will this include the item named in `offset` or start from the next one? It's not immediately obvious to me.

styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
```diff
@@ -107,29 +109,46 @@ void writeEvent(SequenceEvent sequenceEvent) throws IOException {
...
+        result = scanner.next();
+      }
     }
-    workflowInstanceDataList.sort(WorkflowInstanceExecutionData.COMPARATOR);
 
-    return workflowInstanceDataList;
+    return workflowInstancesSet.parallelStream()
+        .map(workflowInstance -> {
+          try {
+            return executionData(workflowInstance);
```
bergman: This is only partly related since it wasn't changed in this commit, but this function will call `readEvents` which in turns reads from the same table we just read from to get this list. Is there a point to that? Why not just use that data in the first scan?

styx-common/src/main/java/com/spotify/styx/storage/Storage.java
```diff
@@ -73,7 +73,7 @@
   /**
    * Removes a workflow definition.
    *
-   * @param workflowId The workflowid to remove.
+   * @param workflowId The workflow id to remove.
```
bergman: Let's use the type instead?

styx-common/src/main/java/com/spotify/styx/storage/Storage.java
```diff
@@ -103,10 +103,15 @@
   /**
    * Get execution information for all the {@link WorkflowInstance} of the specified {@link WorkflowId}.
    *
+   * <p>Results can be paginated based on a offset {@link WorkflowInstance#parameter()} and a limit.
+   *
    * @param workflowId  The workflowId to get execution information for
+   * @param offset      The offset parameter
```
bergman: The id of the last item from previous fetch. Or something to that effect.

styx-common/src/test/java/com/spotify/styx/storage/BigTableStorageTest.java
```diff
@@ -113,7 +118,34 @@ public void shouldReturnExecutionDataForWorkflow() throws Exception {
...
 
+  @Test
+  public void shouldPaginateExecutionDataForWorkflow() throws Exception {
+    setUp(0);
+    storage.writeEvent(SequenceEvent.create(Event.triggerExecution(WFI1, "triggerId1"), 0L, 0L));
+    storage.writeEvent(SequenceEvent.create(Event.triggerExecution(WFI2, "triggerId2"), 0L, 3L));
+    storage.writeEvent(SequenceEvent.create(Event.triggerExecution(WFI3, "triggerId3"), 0L, 3L));
+
+    List<WorkflowInstanceExecutionData> workflowInstanceExecutionData =
+        storage.executionData(WORKFLOW_ID1, WFI2.parameter(), 100);
```
bergman: Ah! So the `offset` parameter is inclusive. This is unexpected to me and not how pagination works to me. You would have to guess the next ID to not get duplicates between pages.

styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
rouzwawi: Yes, on a local run on request that took 9 seconds it went down to about 3.

styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
rouzwawi: The `startRow` is inclusive

rouzwawi:
Discussion offline: we'll merge this for now and iterate on the details.

PR #19, opened 2016-12-04
Handle IOExceptions in TriggerManager

rouzwawi:
And don't use InMemStorage for tests

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
@@ -50,6 +50,7 @@ public class TriggerManager {
 
   private static final Logger LOG = LoggerFactory.getLogger(TriggerManager.class);
 
+  private static final String NATURAL_TRIGGER = "natural-trigger";
   private static final int INITIAL_DELAY_SECONDS = 1;
   private static final int TICK_INTERVAL_SECONDS = 1;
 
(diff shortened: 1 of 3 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -87,28 +88,28 @@ void tick() {
...
         try {
-          triggerListener.event(
-              workflow,
-              "natural-trigger",
-              decrementInstant(naturalTrigger, partitioning));
+          triggerListener.event(workflow, NATURAL_TRIGGER, decrementInstant(next, partitioning));
         } catch (AlreadyInitializedException e) {
           LOG.warn("{}", e.getMessage());
+        } catch (Throwable e) {
+          LOG.warn("Triggering {} threw exception", workflow.id(), e);
```
rouzwawi: So that we don't miss a trigger in the case the trigger listener failed.

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -77,8 +78,8 @@ public void start() {
   }
 
   void tick() {
-    Map<Workflow, Optional<Instant>> map;
-    Set<WorkflowId> enabled;
+    final Map<Workflow, Optional<Instant>> map;
```
bergman: These `final`s… is there a rule to them? They clutter PRs when done like this.

rouzwawi:
I follow
- use final fields for all production code (unless it really has to be non-final)
- don't care for test code (I usually don't use final fields)

PR #22, opened 2016-12-08
Reject creating a pod if referenced secret does not exist

### LINKED ISSUE #12: No feedback when Workflow references non-existing secret
If the `secret` field of a Workflow Schedule references a secret volume that does not exist in k8s, Pods will fail to start but Styx does not detect the failure. The result is that the execution hangs in `SUBMITTED` for the TTL period and then times out. No useful feedback about the cause of the issue is surfaced anywhere.

rouzwawi:
fixes #12


bergman:
👍

bergman:
👍

PR #23, opened 2016-12-08
Resources

bergman:
Missing:
 - [ ] cli
 - [x] tests

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/CliResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/CliResource.java
@@ -26,8 +26,6 @@ import static com.spotify.styx.util.StreamUtil.cat;
 
 import com.google.api.client.util.Lists;
 import com.google.common.base.Throwables;
-import com.spotify.apollo.Client;
-import com.spotify.apollo.Request;
 import com.spotify.apollo.RequestContext;
 import com.spotify.apollo.Response;
 import com.spotify.apollo.entity.EntityMiddleware;
@@ -96,11 +94,21 @@ public class CliResource {
 
     final List<Route<AsyncHandler<Response<ByteString>>>> proxies = Arrays.asList(
         Route.async(
-            "POST", BASE + "/events",
-            this::injectEventProxy),
+            "GET", BASE + "/<endpoint:path>",
+            rc -> proxy("/" + arg("endpoint", rc), rc)),
         Route.async(
-            "POST", BASE + "/trigger",
-            this::triggerWorkflowInstanceProxy));
+            "POST", BASE + "/<endpoint:path>",
+            rc -> proxy("/" + arg("endpoint", rc), rc)),
+        Route.async(
+            "DELETE", BASE + "/<endpoint:path>",
+            rc -> proxy("/" + arg("endpoint", rc), rc)),
+        Route.async(
+            "PATCH", BASE + "/<endpoint:path>",
+            rc -> proxy("/" + arg("endpoint", rc), rc)),
+        Route.async(
+            "PUT", BASE + "/<endpoint:path>",
+            rc -> proxy("/" + arg("endpoint", rc), rc))
+    );
 
     return cat(
         routes.stream().map(r -> r.withPrefix(Api.Version.V0.prefix())),
@@ -108,6 +116,12 @@ public class CliResource {
         proxies.stream().map(r -> r.withPrefix(Api.Version.V0.prefix())),
         proxies.stream().map(r -> r.withPrefix(Api.Version.V1.prefix()))
     );
+  }
+
+  private CompletionStage<Response<ByteString>> proxy(String endpoint, RequestContext rc) {
+    return rc.requestScopedClient()
+        .send(rc.request()
+            .withUri(schedulerServiceBaseUrl + SCHEDULER_BASE_PATH + endpoint));
   }
 
   private static String arg(String name, RequestContext rc) {
@@ -178,22 +192,6 @@ public class CliResource {
     } catch (IOException e) {
       throw Throwables.propagate(e);
     }
-  }
-
-  private CompletionStage<Response<ByteString>> injectEventProxy(RequestContext requestContext) {
-    final Client client = requestContext.requestScopedClient();
-    final Request proxyRequest = requestContext.request()
-        .withUri(schedulerServiceBaseUrl + SCHEDULER_BASE_PATH + "/events");
-
-    return client.send(proxyRequest);
-  }
-
-  private CompletionStage<Response<ByteString>> triggerWorkflowInstanceProxy(RequestContext requestContext) {
-    final Client client = requestContext.requestScopedClient();
-    final Request proxyRequest = requestContext.request()
-        .withUri(schedulerServiceBaseUrl + SCHEDULER_BASE_PATH + "/trigger");
-
-    return client.send(proxyRequest);
   }
 
   private class LastExecutionEventVisitor implements EventVisitor<Boolean> {
--- a/styx-api-service/src/main/java/com/spotify/styx/api/ResourceResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/ResourceResource.java
@@ -1,0 +1,138 @@
+/*-
+ * -\-\-
+ * Spotify Styx API Service
+ * --
+ * Copyright (C) 2016 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.styx.api;
+
+import com.google.common.base.Throwables;
+import com.spotify.apollo.RequestContext;
+import com.spotify.apollo.Response;
+import com.spotify.apollo.Status;
+import com.spotify.apollo.entity.EntityMiddleware;
+import com.spotify.apollo.entity.JacksonEntityCodec;
+import com.spotify.apollo.route.AsyncHandler;
+import com.spotify.apollo.route.Middleware;
+import com.spotify.apollo.route.Route;
+import com.spotify.styx.model.Resource;
+import com.spotify.styx.storage.Storage;
+import java.io.IOException;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Stream;
+import okio.ByteString;
+
+public final class ResourceResource {
+
+  static final String BASE = "/resources";
+
+  private final Storage storage;
+
+  public ResourceResource(Storage storage) {
+    this.storage = Objects.requireNonNull(storage);
+  }
+
+  public Stream<? extends Route<? extends AsyncHandler<? extends Response<ByteString>>>> routes() {
+    final EntityMiddleware em =
+        EntityMiddleware.forCodec(JacksonEntityCodec.forMapper(Middlewares.OBJECT_MAPPER));
+
+    return Stream.of(
+        Route.with(
+            em.serializerDirect(ResourcesPayload.class),
+            "GET", BASE,
+            rc -> getResources()),
+        Route.with(
+            em.direct(Resource.class),
+            "POST", BASE,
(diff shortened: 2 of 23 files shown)


styx-api-service/src/main/java/com/spotify/styx/api/CliResource.java
```diff
@@ -96,11 +94,21 @@ public CliResource(String schedulerServiceBaseUrl, EventStorage eventStorage) {
...
+            rc -> proxy("/" + arg("endpoint", rc), rc)),
+        Route.async(
+            "GET", BASE + "/<endpoint>/<id>",
+            rc -> proxy("/" + arg("endpoint", rc) + "/" + arg("id", rc), rc)),
+        Route.async(
+            "DELETE", BASE + "/<endpoint>/<id>",
+            rc -> proxy("/" + arg("endpoint", rc) + "/" + arg("id", rc), rc)),
+        Route.async(
+            "PATCH", BASE + "/<endpoint>/<id>",
+            rc -> proxy("/" + arg("endpoint", rc) + "/" + arg("id", rc), rc))
```
rouzwawi: You might want to replace these all with just a `/<endpoint:path>` parameter that will match the rest of the uri. see https://github.com/spotify/apollo/blob/master/apollo-api/docs/routing-engine.md#path-parameters

styx-api-service/src/main/java/com/spotify/styx/api/ResourceResource.java
```diff
@@ -0,0 +1,171 @@
...
+    if (!payload.isPresent()) {
+      return Response.forStatus(Status.BAD_REQUEST.withReasonPhrase("Missing payload."));
+    }
+
+    Resource resource;
+    try {
+      resource = OBJECT_MAPPER.readValue(payload.get().toByteArray(), Resource.class);
+    } catch (IOException e) {
+      return Response.forStatus(Status.BAD_REQUEST.withReasonPhrase("Invalid payload."));
+    }
```
rouzwawi: Use the EntityMiddleware to avoid having to do manual parsing and error handling.

styx-common/src/main/java/com/spotify/styx/model/DataEndpoint.java
```diff
@@ -49,15 +49,19 @@
...
+
   @JsonCreator
   public static DataEndpoint create(
       @JsonProperty("id") String id,
       @JsonProperty("partitioning") Partitioning partitioning,
       @JsonProperty("docker_image") Optional<String> dockerImage,
       @JsonProperty("docker_args") Optional<List<String>> dockerArgs,
-      @JsonProperty("secret") Optional<Secret> secret) {
+      @JsonProperty("secret") Optional<Secret> secret,
+      @JsonProperty("resources") Optional<List<String>> resources) {
```
rouzwawi: I think jackson can handle parsing a missing key directly into an empty list, without having to wrap it in an optional. For the `resources` field, a missing key and an empty list should have the same meaning.

styx-common/src/main/java/com/spotify/styx/model/Resource.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+@AutoValue
+@JsonIgnoreProperties(ignoreUnknown = true)
+public abstract class Resource {
+
+  @JsonProperty
+  public abstract String id();
+
+  @JsonProperty
+  public abstract long concurrency();
```
rouzwawi: Do we want to call it `concurrency` or `limit`? "Resource limit" sounds more natural to me than a "Resource concurrency".

styx-api-service/src/main/java/com/spotify/styx/api/ResourceResource.java
```diff
@@ -0,0 +1,171 @@
...
+            json(), "POST", BASE,
+            rc -> postResource(rc.request())),
+        Route.with(
+            json(), "GET", BASE + "/<rid>",
+            rc -> getResource(arg("rid", rc))),
+        Route.with(
+            json(), "DELETE", BASE + "/<rid>",
+            rc -> deleteResource(arg("rid", rc))),
+        Route.with(
+            json(), "PATCH", BASE + "/<rid>",
```
rouzwawi: Shouldn't this be PUT?

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -451,7 +451,8 @@ private TriggerListener trigger(
 
     return (workflow, triggerId, instant) -> {
       try {
-        if (!storage.enabled(workflow.id()) || !storage.globalEnabled()) {
+        final Set<WorkflowId> enabledWorkflows = storage.enabledWorkflows();
```
rouzwawi: Unrelated change? Why get all the enabled workflows instead of just checking for a specific one?

rouzwawi: If it was changed because `enabled(WorkflowId)` is deprecated, then use `storage.workflowState` instead.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -477,4 +482,39 @@ private static DateTime instantToDatetime(Instant instant) {
   private static Instant datetimeToInstant(DateTime dateTime) {
     return dateTime.toDate().toInstant();
   }
+
+  Resource getResource(String id) {
+    return entityToResource(
+        datastore.get(datastore.newKeyFactory().kind(KIND_RESOURCE).newKey(id)));
```
rouzwawi: Potential NPE if the lookup does not find a result.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -477,4 +482,39 @@ private static DateTime instantToDatetime(Instant instant) {
...
+  }
+
+  void postResource(Resource resource) {
+    datastore.put(resourceToEntity(resource));
```
rouzwawi: Wrap in retries?

styx-common/src/main/java/com/spotify/styx/storage/InMemStorage.java
```diff
@@ -20,9 +20,11 @@
 
 package com.spotify.styx.storage;
 
+import com.google.bigtable.repackaged.com.google.common.collect.ImmutableList;
```
rouzwawi: Shaded import

styx-common/src/main/java/com/spotify/styx/storage/Storage.java
```diff
@@ -116,14 +117,14 @@
    * @return true if the queried workflow is enabled
    */
   @Deprecated
-  boolean enabled(WorkflowId workflowId) throws IOException;
+  boolean isWorkflowEnabled(WorkflowId workflowId) throws IOException;
```
rouzwawi: These were named with the argument type in mind, like `store(Workflow)`, `delete(WorkflowId)`, `enabled(WorkflowId)`.

The names are all repetitive now with `storeWorkflow(Workflow)`.

I can see that the last one is be better as `isEnabled(WorkflowId)`, though.

If we want to rename them, then maybe the metric names in `MeteredStorage` should be aligned too.

styx-api-service/src/main/java/com/spotify/styx/api/CliResource.java
bergman: Thanks, was thinking of that but didn't know how to do it :)

styx-common/src/main/java/com/spotify/styx/model/Resource.java
bergman: The word `limit` might not convey that it gets replenished after use. Maybe we can look into what wording thread pools use?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
bergman: Thanks!

styx-common/src/main/java/com/spotify/styx/storage/InMemStorage.java
bergman: Thanks!

styx-common/src/main/java/com/spotify/styx/storage/Storage.java
bergman: Ah, forgot about the strings in `MeteredStorage`. Will have another pass at this.

bergman: Will keep `enabledWorkflows` and `isEnabled` but revert the others

bergman: Will revert all names :)

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -477,4 +482,43 @@ private static DateTime instantToDatetime(Instant instant) {
...
+
+  Optional<Resource> getResource(String id) {
+    Entity entity = datastore.get(datastore.newKeyFactory().kind(KIND_RESOURCE).newKey(id));
+    if (entity == null) {
+      return Optional.empty();
+    }
+    return Optional.of(entityToResource(entity));
+  }
+
+  void postResource(Resource resource) throws IOException {
```
rouzwawi: No transaction needed when only doing one operation

bergman: Fixed, thanks!

rouzwawi:
@bergman should we wait for cli and tests or are you doing them as separate prs?

bergman:
Let's do CLI and tests in a separate PR if you need this now. Just know that this code has never been actually run against datastore (mocked or otherwise).

rouzwawi:
Then we should at least do tests in this PR.

styx-api-service/src/test/java/com/spotify/styx/api/ResourceResourceTest.java
```diff
@@ -0,0 +1,257 @@
...
+      throws InterruptedException, java.util.concurrent.ExecutionException,
+             java.util.concurrent.TimeoutException {
+    return completionStage.toCompletableFuture().get(5, TimeUnit.SECONDS);
+  }
+
+  private String path(String path) {
+    return Api.Version.V1.prefix() + ResourceResource.BASE + path;
+  }
+
+  private Connection setupBigTableMockTable() {
```
rouzwawi: Do we have to set up mocked behaviour for bigtable for this test? I think it would be enough just to send in a plain mock, since none of the paths will interact with bigtable.

styx-api-service/src/test/java/com/spotify/styx/api/ResourceResourceTest.java
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  @Test
+  public void shouldListMultipleResources() throws Exception {
+    storage.storeResource(RESOURCE_2);
+    Response<ByteString> response =
+        awaitResponse(serviceHelper.request("GET", path("")));
+
+    assertThat(response, hasStatus(belongsToFamily(StatusType.Family.SUCCESSFUL)));
+    System.out.println("response.payload().get().utf8() = " + response.payload().get().utf8());
```
rouzwawi: println

styx-api-service/src/test/java/com/spotify/styx/api/ResourceResourceTest.java
```diff
@@ -0,0 +1,257 @@
...
+        ImmutableMap.of("id", "resource1", "concurrency", 1),
+        ImmutableMap.of("id", "resource2", "concurrency", 2)
+    ));
+
+    // change resource2
+    Response<ByteString> putResponse =
+        awaitResponse(serviceHelper.request("PUT", path("/resource2"),
+            ByteString.encodeUtf8("{\"id\": \"resource2\", \"concurrency\": 3}")));
+    assertThat(putResponse, hasStatus(belongsToFamily(StatusType.Family.SUCCESSFUL)));
+    assertJson(putResponse, "id", equalTo("resource2"));
```
rouzwawi: Try not to assert things that are not essential for the test. Checking the put response object is already done in the `shouldUpdateResource` test case.

bergman: yup, same for first listing i guess, it's the same case as `shouldListMultipleResources` except that one doesn't `POST` but uses datastore directly

rouzwawi: Yeah I was thinking about that but then thought that it makes sense to test that the changes are actually reflected in the api. But checking the put response is not directly related.

rouzwawi:
LGTM 👍  :shipit:

PR #25, opened 2016-12-08
Store triggerId in state

rouzwawi:
Preparation for injecting the trigger id into the k8s pod, and other upcoming features.


fabriziodemaria:
We can now add trigger checks in StateInitializingTriggerTest (https://github.com/spotify/styx/blob/master/styx-scheduler-service/src/test/java/com/spotify/styx/StateInitializingTriggerTest.java#L89)

rouzwawi:
@fabriziodemaria good catch, fixed that test.

rouzwawi:
@fabriziodemaria merge?

PR #26, opened 2016-12-09
Triggers transition into QUEUED state

fabriziodemaria:


styx-common/src/main/java/com/spotify/styx/state/RunState.java
```diff
@@ -172,8 +172,20 @@ public RunState created(WorkflowInstance workflowInstance, String executionId, S
...
+
+        default:
+          throw illegalTransition("dequeue");
+      }
+    }
+
+    @Override
     public RunState submit(WorkflowInstance workflowInstance, ExecutionDescription executionDescription) {
       switch (state()) {
+        case QUEUED: // for backwards compatibility
```
rouzwawi: This probably also have to be added for the legacy `created` event.

styx-common/src/test/java/com/spotify/styx/state/RunStateTest.java
```diff
@@ -101,6 +101,7 @@ public void testRunErrorOnCreating() throws Exception {
   public void testSetExecutionId() throws Exception {
     transitioner.initialize(RunState.fresh(WORKFLOW_INSTANCE, this::record));
     transitioner.receive(eventFactory.triggerExecution("trig"));
+    transitioner.receive(eventFactory.dequeue());
```
rouzwawi: I think these tests should not be modified to add a `dequeue` event but instead the state machine has to be backwards compatible with `triggerExecution -> created` sequences. See comment above.

styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
```diff
@@ -87,8 +87,19 @@ public void shouldNotTransitionIfNotTimedOut() throws Exception {
...
+
+    assertThat(stateManager.get(INSTANCE).state(), is(State.PREPARE));
+  }
+
+  @Test
+  public void shouldExecuteNewTriggers() throws Exception {
+    StateData stateData = StateData.builder().tries(0).build();
     setUp(20, RunState.create(INSTANCE, State.QUEUED, stateData, time));
 
     now = now.plus(15, ChronoUnit.SECONDS);
```
rouzwawi: Maybe this line should be removed so that we ensure that it dequeues the state immediately?

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
```diff
@@ -279,21 +289,23 @@ public void restoresActiveStatesFromBigtableRetriggered() throws Exception {
...
-    givenStoredEvent(Event.retryAfter(workflowInstance, 30000),                                7L);
-    givenStoredEvent(Event.halt(workflowInstance),                                             8L);
-    givenStoredEvent(Event.triggerExecution(workflowInstance, "trig2"),                        9L);
-    givenStoredEvent(Event.created(workflowInstance, TEST_EXECUTION_ID_1, TEST_DOCKER_IMAGE), 10L);
-    givenStoredEvent(Event.started(workflowInstance),                                         11L);
-    givenStoredEvent(Event.terminate(workflowInstance, 30),                                   12L);
-    givenStoredEvent(Event.retryAfter(workflowInstance, 30000),                               13L);
-    givenActiveStateAtSequenceCount(workflowInstance,                                         13L);
+    givenStoredEvent(Event.triggerExecution(workflowInstance, "trig1"),                       0L);
+    givenStoredEvent(Event.dequeue(workflowInstance),                                         1L);
```
rouzwawi: Same comment here as in the RunState test, this test should pass without modification.

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
```diff
@@ -64,23 +64,29 @@ public void shouldCatchUpWithNaturalTriggers() throws Exception {
     givenNextNaturalTrigger(HOURLY_WORKFLOW, "2016-03-14T13:00:00Z");
 
     styxStarts();
-    timePasses(1, SECONDS);
+    tickTriggerManager();
+    Thread.sleep(100);
```
rouzwawi: Can this be implemented without a sleep? Perhaps use awaitility to wait for the condition you had in mind.

styx-common/src/main/java/com/spotify/styx/state/RunState.java
fabriziodemaria: Since we added the QUEUED state after deprecating `created`, we should not have in storage a `created` transition that was executed on top of a state machine that includes the QUEUED state. Do we really need to allow for that transition in RunState?

fabriziodemaria: Never mind! Modifying `triggerExecution` behaviour indeed means we need to allow for the `created` transition from QUEUED.

rouzwawi:
I think this looks good.

It was a bit unclear why some of the tests were modified to add the `dequeue` event. I had to checkout out the code locally and remove those changes to make sure that they were not added to fix a failing test, as opposed to just aligning the tests to use the new event sequence.

Seems all of the ones I tried was the latter, so it's ok. For future changes, make it clear if a test was changed to align or to fix failures.

fabriziodemaria:
Thanks for pointing that out. I will make my intentions clear with comments the next time tests are changed this way.

rouzwawi:
👍

Issue #24, opened 2016-12-08
Changing Workflow partitioning causes trigger time offsets


rouzwawi:
When changing a Workflows partitioning configuration, the previously stored `nextNaturalTrigger` time is kept, which potentially leads to an offset trigger time.

Reproducing
- found an hourly Workflow that was enabled
- nextNaturalTrigger: 2016-12-08 (19:00:00.000) UTC
- changed to daily
- it was triggered at 2016-12-08 (19:00:00.000) UTC
- nextNaturalTrigger: 2016-12-09 (19:00:00.000) UTC

The workflow is now offset to 19:00 UTC instead of 00:00 which a daily workflow would trigger at normally.

PR #27, opened 2016-12-13
Next natural trigger bug fix

TinaRanic:
Fixes https://github.com/spotify/styx/issues/24

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/model/WorkflowState.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/WorkflowState.java
@@ -20,12 +20,11 @@
 
 package com.spotify.styx.model;
 
-import static java.util.Optional.of;
-
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
+import java.time.Instant;
 import java.util.Optional;
 
 /**
@@ -44,27 +43,53 @@ public abstract class WorkflowState {
   @JsonProperty
   public abstract Optional<String> commitSha();
 
+  @JsonProperty
+  public abstract Optional<Instant> nextNaturalTrigger();
+
+  public abstract Builder toBuilder();
+
+  public static Builder builder() {
+    return new AutoValue_WorkflowState.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    public abstract Builder enabled(boolean enabled);
+    public abstract Builder dockerImage(String dockerImage);
+    public abstract Builder commitSha(String commitSha);
+    public abstract Builder nextNaturalTrigger(Instant nextNaturalTrigger);
+
+    public abstract WorkflowState build();
+  }
+
   @JsonCreator
   public static WorkflowState create(
       @JsonProperty("enabled") Optional<Boolean> enabled,
       @JsonProperty("docker_image") Optional<String> dockerImage,
-      @JsonProperty("commit_sha") Optional<String> commitSha) {
-    return new AutoValue_WorkflowState(enabled, dockerImage, commitSha);
+      @JsonProperty("commit_sha") Optional<String> commitSha,
+      @JsonProperty("next_natural_trigger") Optional<Instant> nextNaturalTrigger) {
+    Builder builder = builder();
+    enabled.ifPresent(e -> builder.enabled(e));
+    dockerImage.ifPresent(di -> builder.dockerImage(di));
+    commitSha.ifPresent(cs -> builder.commitSha(cs));
+    nextNaturalTrigger.ifPresent(nnt -> builder.nextNaturalTrigger(nnt));
+    return builder.build();
   }
 
   public static WorkflowState all(boolean enabled, String dockerImage, String commitSha) {
-    return create(of(enabled), of(dockerImage), of(commitSha));
+    return builder().enabled(enabled).dockerImage(dockerImage).commitSha(commitSha).build();
   }
 
   public static WorkflowState empty() {
-    return create(Optional.empty(), Optional.empty(), Optional.empty());
+    return builder().build();
   }
 
   public static WorkflowState patchDockerImage(String dockerImage) {
-    return create(Optional.empty(), of(dockerImage), Optional.empty());
+    return builder().dockerImage(dockerImage).build();
   }
 
   public static WorkflowState patchEnabled(boolean enabled) {
-    return create(of(enabled), Optional.empty(), Optional.empty());
+    return builder().enabled(enabled).build();
   }
 }
--- a/styx-common/src/main/java/com/spotify/styx/model/WorkflowState.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/WorkflowState.java
@@ -70,10 +70,10 @@ public abstract class WorkflowState {
       @JsonProperty("commit_sha") Optional<String> commitSha,
       @JsonProperty("next_natural_trigger") Optional<Instant> nextNaturalTrigger) {
     Builder builder = builder();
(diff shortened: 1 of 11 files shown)


styx-common/src/main/java/com/spotify/styx/util/WorkflowStateUtil.java
```diff
@@ -30,11 +30,17 @@ private WorkflowStateUtil() { }
...
       final WorkflowState patch) {
+
     return originalWorkflowState.map(
-        o -> WorkflowState.create(
-            Optional.of(patch.enabled().orElse(o.enabled().orElse(false))),
-            patch.dockerImage().isPresent() ? patch.dockerImage() : o.dockerImage(),
-            patch.commitSha().isPresent() ? patch.commitSha() : o.commitSha())
+        o -> {
+          WorkflowState.Builder builder = o.toBuilder()
+              .enabled(patch.enabled().orElse(originalWorkflowState.get().enabled().orElse(false)));
```
rouzwawi: This line should use `o` instead of `originalWorkflowState`

styx-common/src/main/java/com/spotify/styx/model/WorkflowState.java
```diff
@@ -44,27 +43,53 @@
...
-      @JsonProperty("commit_sha") Optional<String> commitSha) {
-    return new AutoValue_WorkflowState(enabled, dockerImage, commitSha);
+      @JsonProperty("commit_sha") Optional<String> commitSha,
+      @JsonProperty("next_natural_trigger") Optional<Instant> nextNaturalTrigger) {
+    Builder builder = builder();
+    enabled.ifPresent(e -> builder.enabled(e));
+    dockerImage.ifPresent(di -> builder.dockerImage(di));
+    commitSha.ifPresent(cs -> builder.commitSha(cs));
+    nextNaturalTrigger.ifPresent(nnt -> builder.nextNaturalTrigger(nnt));
+    return builder.build();
```
rouzwawi: Why not just call the `AutoValue_WorkflowState` constructor?

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
```diff
@@ -114,6 +124,93 @@ public void removedEnabledWorkflowWontGetScheduled() throws Exception {
...
+    awaitNumberOfDockerRuns(1);
+
+    WorkflowInstance workflowInstance = dockerRuns.get(0)._1;
+    RunSpec runSpec = dockerRuns.get(0)._2;
+    assertThat(workflowInstance.workflowId(), is(DAILY_WORKFLOW.id()));
+    assertThat(runSpec, is(RunSpec.simple("busybox", "--hour", "2016-03-13")));
+
+    injectEvent(Event.started(workflowInstance));
+    injectEvent(Event.terminate(workflowInstance, 0));
+    Thread.sleep(1000);
```
rouzwawi: Try using `awaitWorkflowInstanceState` instead of sleeping. I believe this line is waiting for the state to be in `AWAITING_RETRY`.

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
```diff
@@ -114,6 +124,93 @@ public void removedEnabledWorkflowWontGetScheduled() throws Exception {
...
+
+    workflowChanges(Workflow.create(DAILY_WORKFLOW.componentId(),
+        DAILY_WORKFLOW.componentUri(),
+        DataEndpoint.create(DATA_ENDPOINT_DAILY.id(), DATA_ENDPOINT_DAILY.partitioning(),
+            Optional.of("freebox"), DATA_ENDPOINT_DAILY.dockerArgs(),
+            DATA_ENDPOINT_DAILY.secret())));
+    timePasses(StyxScheduler.SCHEDULER_TICK_INTERVAL_SECONDS, SECONDS);
+
+    Thread.sleep(1000);
```
rouzwawi: Same here, avoid sleeps.

rouzwawi:
Some comments, but looks good overall.

The tests need some changes, plus a rebase.

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
TinaRanic: This test verifies that after a wf has executed successfully with a given parameter, a change of its run specification wont re-trigger the WFI with the same parameter.

I want to check if the state is `DONE`, but using the awaitility didn't work because the state has been already processed (is not anymore in `activeStates`), so `getState` returns null. But I can just remove the sleeping assuming that the state will always reach `DONE` before starting a new statement?

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
TinaRanic: I didn't find `Awaitility` way to check that something has not changed.

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
rouzwawi: How about adding a way to await that a workflow instance completes? e.g. `awaitWorkflowInstanceCompletion(wfi)` that will wait until the state manager returns a null state'

styx-common/src/main/java/com/spotify/styx/model/WorkflowState.java
rouzwawi: ok

rouzwawi:
👍

styx-common/src/main/java/com/spotify/styx/model/WorkflowState.java
bergman: I guess all of these could be use method references rather than lambdas?

```java
enabled.ifPresent(builder::enabled);
```

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -480,7 +485,7 @@ private TriggerListener trigger(
     };
   }
 
-  private static Consumer<Workflow> workflowChanged(
+  private Consumer<Workflow> workflowChanged(
```
bergman: Most other methods in this class seem to be static and get passed everything needed. Did you consider passing `time` into this method instead of making it an instance method?

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerServiceFixture.java
```diff
@@ -187,6 +187,18 @@ void workflowDeleted(Workflow workflow) {
     workflowRemoveListener.accept(workflow);
   }
 
+  /**
+   * Fast forwards the time without any execution in-between.
+   */
+  void timeJumps(int n, TimeUnit unit) {
```
bergman: Do you think it's somehow possible to have the distinction even clearer between `timePasses` and `timeJumps` without having to read the javadoc? `jump` compared to `pass` is in the right direction but I would definitely need to read the javadoc to know for sure.

Not a big deal at all, just thought it was unclear.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -354,11 +355,10 @@ void patchState(String componentId, WorkflowState state) throws IOException {
   }
 
   public WorkflowState workflowState(WorkflowId workflowId) throws IOException {
-    return
-        WorkflowState.create(
-            Optional.of(enabled(workflowId)),
-            getDockerImage(workflowId),
-            getCommitSha(workflowId));
+    WorkflowState.Builder builder = WorkflowState.builder().enabled(enabled(workflowId));
+    getDockerImage(workflowId).ifPresent(dockerImage -> builder.dockerImage(dockerImage));
```
bergman: method references

bergman:
👍

PR #28, opened 2016-12-13
[cli] Fallback to json inspection for unknown events

rouzwawi:
We keep breaking the CLI each time we add a new event, because of how we use Jackson.

This will make the cli more forgiving and fallback to just displaying the event name in case it can not parse it into the static types.


TinaRanic:
nice! 💯 👍

bergman:
👍

PR #29, opened 2016-12-14
Limit Concurrency

rouzwawi:
todo

- [x] tests
- [x] documentation

depends on #23

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -20,16 +20,41 @@
 
 package com.spotify.styx;
 
+import static java.util.Collections.emptySet;
+import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.counting;
+import static java.util.stream.Collectors.groupingByConcurrent;
+import static java.util.stream.Collectors.toCollection;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toMap;
+import static java.util.stream.Collectors.toSet;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 import com.spotify.styx.model.Event;
+import com.spotify.styx.model.Resource;
+import com.spotify.styx.model.Workflow;
+import com.spotify.styx.model.WorkflowId;
 import com.spotify.styx.model.WorkflowInstance;
 import com.spotify.styx.state.RunState;
+import com.spotify.styx.state.RunState.State;
 import com.spotify.styx.state.StateData;
 import com.spotify.styx.state.StateManager;
 import com.spotify.styx.state.TimeoutConfig;
+import com.spotify.styx.storage.Storage;
 import com.spotify.styx.util.Time;
+import java.io.IOException;
 import java.time.Instant;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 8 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -50,32 +74,115 @@
...
+              return usage < resource.concurrency();
+            });
+
+        if (canRun) {
+          resourceRefs.forEach(id -> currentResourceUsage.computeIfAbsent(id, id_ -> 0L));
+          resourceRefs.forEach(id -> currentResourceUsage.compute(id, (id_, l) -> l + 1));
+          sendDequeue(instance);
         } else {
-          LOG.info("{} executing retry #{}", key.toKey(), state.data().tries());
+          // todo: send info event, could not run because resource limit reached
```
bergman: will this come in this PR?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -50,32 +74,115 @@
...
+        .map(resource -> ResourceWithInstance.create(resource, instanceState));
+  }
+
+  private Set<String> workflowResources(WorkflowId workflowId) {
+    final Optional<Workflow> workflowOpt = workflowCache.workflow(workflowId);
+    if (!workflowOpt.isPresent()) {
+      return emptySet();
+    }
+
+    return Sets.newHashSet(workflowOpt.get().schedule().resources());
```
bergman: why is this preferred over `new HashSet<>()`? I can see the use in pre-jdk7 but not since `<>` was ok.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
rouzwawi: We have a separate ticket for it, so I'll leave the todos.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -50,32 +74,115 @@
...
+        activeStates.parallelStream()
+            .map(InstanceState::workflowInstance)
+            .map(WorkflowInstance::workflowId)
+            .distinct()
+            .collect(toMap(workflowId -> workflowId, this::workflowResources));
+
+    final Map<String, Long> currentResourceUsage =
+        activeStates.parallelStream()
+            .filter(entry -> !timedOutInstances.contains(entry.workflowInstance()))
+            .filter(entry -> entry.runState().state() != RunState.State.QUEUED)
```
TinaRanic: After we have filtered out the instances about to time out, shouldn't we then filter only instances whose state is `RUNNING`?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -50,32 +74,115 @@
...
+                ConcurrentHashMap::new,
+                counting()));
+
+    final List<InstanceState> eligibleInstances =
+        activeStates.parallelStream()
+            .filter(entry -> !timedOutInstances.contains(entry.workflowInstance()))
+            .filter(entry -> shouldExecute(entry.runState()))
+            .collect(toCollection(Lists::newArrayList));
+    Collections.shuffle(eligibleInstances);
+
```
TinaRanic: Does it make sense the `Resource` related methods to move them to something like `ResourceManager`?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
rouzwawi: That was what I did in my first iteration, but it causes a race.

Example: The scheduler will dequeue some instances on tick 1. These go into submitting (not yet running). The scheduler ticks again and sees free slots in the same resources and dequeues even more instances.

By looking at instances not in queued, we avoid this condition.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -50,32 +74,115 @@
...
+        activeStates.parallelStream()
+            .filter(entry -> hasTimedOut(entry.runState()))
+            .map(InstanceState::workflowInstance)
+            .collect(toSet());
+
+    final Map<WorkflowId, Set<String>> workflowResourceReferences =
+        activeStates.parallelStream()
+            .map(InstanceState::workflowInstance)
+            .map(WorkflowInstance::workflowId)
+            .distinct()
```
bergman: is this really needed given you put them in a map in the next step?

bergman: ah, these parallel streams… so one could use `.collect(toConcurrentMap())` instead

bergman:
👍

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
rouzwawi: That makes sense, but I would like to do it at a later point when the patterns have settled a bit.

PR #31, opened 2016-12-18
General cleanup

rouzwawi:
- Merges `EventStorage` and `Storage` interfaces
- Adds utility method for generating versioned routes - `Api.prefixRoutes(routes, V0, V1, ...)`
- Moves versioned api test boilerplate into `VersionedApiTest` base class
- Simplifies some transformation logic in the cli module

---

This PR as a prerequisite for the [`state-info`](https://github.com/spotify/styx/tree/state-info) branch.

PR #32, opened 2016-12-18
State Info

rouzwawi:
Merge after #31

This adds a general info messaging capability to the `StateData` and refactors the CLI to use it for displaying information about the active states.

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/CliResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/CliResource.java
@@ -29,7 +29,6 @@ import static java.util.stream.Collectors.toList;
 
 import com.google.api.client.util.Lists;
 import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableSet;
 import com.spotify.apollo.RequestContext;
 import com.spotify.apollo.Response;
 import com.spotify.apollo.entity.EntityMiddleware;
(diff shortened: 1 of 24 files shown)


styx-common/src/main/java/com/spotify/styx/api/cli/ActiveStatesPayload.java
```diff
@@ -52,20 +56,33 @@
     public abstract String state();
 
     @JsonProperty
+    public abstract StateData stateData();
```
rouzwawi: This addition will break older CLI artifacts, even though it's a backwards compatible change.

That is because the `ActiveState` type was missing the `@JsonIgnoreProperties(ignoreUnknown = true)` annotation which I added now.

bergman:
rebased after merge of #31

styx-api-service/src/main/java/com/spotify/styx/api/CliResource.java
```diff
@@ -152,11 +152,13 @@ private ActiveStatesPayload activeStates(RequestContext requestContext) {
     return ActiveStatesPayload.ActiveState.create(
         state.workflowInstance(),
         state.state().toString(),
-        state.data().executionId().orElse("<no execution id>"),
+        state.data(),
+        state.data().executionId().orElse("<no execution id>"), // deprecated
```
bergman: if we're breaking cli anyway with this release, why keep this?

bergman:
👍

PR #33, opened 2016-12-19
Display number of tries in cli

rouzwawi:
screen shot 2016-12-19 at 15 04 12


bergman:
moving number of tries incrementing to start of try instead of end of try?

rouzwawi:
Yes. The number was a bit confusing as it was before since it would stay at 0 until termination or failure. But incrementing it with 1 in the cli code caused it to say tries 2 when QUEUED for the second run.

bergman:
👍

PR #35, opened 2016-12-21
Add error event when pod is missing

TinaRanic:
I might remove the isntanceStates() method on StateManager. I liked using instanceState directly, but we can derive it from the activeStates().

@rouz @jocke @fdema

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -21,6 +21,8 @@
 package com.spotify.styx.docker;
 
 import static com.spotify.styx.docker.KubernetesPodEventTranslator.translate;
+import static com.spotify.styx.state.RunState.State.RUNNING;
+import static java.util.stream.Collectors.toSet;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
@@ -52,6 +54,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -190,9 +193,31 @@ class KubernetesDockerRunner implements DockerRunner {
         .watch(new PodWatcher(Integer.parseInt(resourceVersion)));
   }
 
-  private void pollPods() {
+  void examineRunningWFISandAssociatedPods(PodList podList) {
+    final Set<WorkflowInstance> runningWorkflowInstances = stateManager.activeStates()
+        .values()
+        .stream()
+        .filter(runState -> runState.state().equals(RUNNING))
+        .map(RunState::workflowInstance)
+        .collect(toSet());
+
+    final Set<WorkflowInstance> workflowInstancesForPods = podList.getItems().stream()
+        .filter(pod -> pod.getMetadata().getAnnotations()
+            .containsKey(STYX_WORKFLOW_INSTANCE_ANNOTATION))
+        .map(pod -> WorkflowInstance
+            .parseKey(pod.getMetadata().getAnnotations().get(STYX_WORKFLOW_INSTANCE_ANNOTATION)))
+        .collect(toSet());
+
+    runningWorkflowInstances.removeAll(workflowInstancesForPods);
+    runningWorkflowInstances.forEach(workflowInstance -> stateManager.receiveIgnoreClosed(
+        Event.runError(workflowInstance, "No pod associated with this instance")));
+  }
+
+  @VisibleForTesting
+  void pollPods() {
     try {
       final PodList list = client.pods().list();
+      examineRunningWFISandAssociatedPods(list);
       for (Pod pod : list.getItems()) {
         inspectPod(Watcher.Action.MODIFIED, pod);
       }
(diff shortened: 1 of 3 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -258,11 +258,11 @@ private void sendTimeout(WorkflowInstance workflowInstance) {
   }
 
   @AutoValue
-  abstract static class InstanceState {
-    abstract WorkflowInstance workflowInstance();
-    abstract RunState runState();
+  public abstract static class InstanceState {
```
rouzwawi: I'd prefer keeping this private.

In `KubernetesDockerRunner`, you can get the `WorkflowInstance` from the `RunState`, no need to spread this class across different component.

TinaRanic: yeah true, I commented on the PR itself it was unnecessary job

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -190,9 +193,31 @@ public void init() {
...
+            .containsKey(STYX_WORKFLOW_INSTANCE_ANNOTATION))
+        .map(pod -> pod.getMetadata().getAnnotations().get(STYX_WORKFLOW_INSTANCE_ANNOTATION))
+        .map(workflowInstanceKey -> WorkflowInstance.parseKey(workflowInstanceKey))
+        .collect(toSet());
+
+    activeWorkflowInstances.removeAll(collect);
+    activeWorkflowInstances.parallelStream()
+        .forEach(workflowInstance -> stateManager.receiveIgnoreClosed(
+            Event.runError(workflowInstance, "No pod associated with this instance")));
+  }
```
rouzwawi: This method is not that time sensitive and I think the collections that are worked on are pretty small that the `paralellStream` use is unnecessary. It might actually be more overhead to dispatch on the fork-join-pool rather than just iterating sequentially.

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesDockerRunnerPodPollerTest.java
```diff
@@ -0,0 +1,142 @@
...
+  public void shouldNotSendRunErrorWhenPodForRunningWFIExists() throws Exception {
+    Pod createdPod = KubernetesDockerRunner.createPod(WORKFLOW_INSTANCE, RUN_SPEC);
+    Pod createdPod2 = KubernetesDockerRunner.createPod(WORKFLOW_INSTANCE_2, RUN_SPEC);
+    podList.setItems(Arrays.asList(createdPod, createdPod2));
+    when(k8sClient.pods().list()).thenReturn(podList);
+    setupActiveInstances();
+
+    kdr.pollPods();
+
+    verify(stateManager, never()).receiveIgnoreClosed(
```
fabriziodemaria: I like we don't depend on something like `SyncStateManager` 👍

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -190,9 +194,32 @@ public void init() {
         .watch(new PodWatcher(Integer.parseInt(resourceVersion)));
   }
 
-  private void pollPods() {
+  void examineRunningWFISandAssociatedPods(PodList podList) {
+    final Set<WorkflowInstance> activeWorkflowInstances = stateManager.activeStates().entrySet()
```
fabriziodemaria: Since RunState includes the WorkflowInstance, we can just use `.values()` followed by something like:
```
.filter(runState -> !runState.equals(RUNNING))
.map(RunState::workflowInstance) 
```

styx-scheduler-service/src/test/java/com/spotify/styx/state/SyncStateManager.java
```diff
@@ -71,6 +75,13 @@ public void receive(Event event) {
   }
 
   @Override
+  public Set<Scheduler.InstanceState> instanceStates() {
```
fabriziodemaria: This is not needed anymore.

fabriziodemaria:
@ranic can we add a test that involves an active state that is not in RUNNING state?

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesDockerRunnerPodPollerTest.java
```diff
@@ -0,0 +1,153 @@
...
+  @Test
+  public void shouldSendRunErrorWhenPodForRunningWFIDoesntExist() throws Exception {
+    Pod createdPod = KubernetesDockerRunner.createPod(WORKFLOW_INSTANCE, RUN_SPEC);
+    podList.setItems(Arrays.asList(createdPod));
+    when(k8sClient.pods().list()).thenReturn(podList);
+    setupActiveInstances(RunState.State.RUNNING);
+
+    kdr.pollPods();
+
+    verify(stateManager, times(1)).receiveIgnoreClosed(
```
fabriziodemaria: Can we also add a `verify` for `WORKFLOW_INSTANCE` to check that that was never called?

TinaRanic: I wouldn't like to reverify for something that there is already a unit test `shouldNotSendRunErrorWhenPodForRunningWFIExists`. I don't think it adds value to improving test coverage.

fabriziodemaria:
👍

PR #36, opened 2016-12-23
add version info to cli

bergman:


styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -325,7 +325,8 @@ private static WorkflowInstance getWorkflowInstance(Namespace namespace) {
   private static class StyxCliParser {
 
     final ArgumentParser parser = ArgumentParsers.newArgumentParser("styx")
-        .description("Styx CLI");
+        .description("Styx CLI")
+        .version("styx " + Main.class.getPackage().getImplementationVersion());
```
rouzwawi: This should maybe say `styx-cli` so we can differentiate it from the server version in case we add that too.

bergman: Not sure about the dash. I think it should either match the program name (`$0`) or be apparent that it's not the same. `Styx CLI 0.2.2`?

rouzwawi:
:+1:

PR #38, opened 2017-01-12
Add setter for RetryUtil in scheduler builder

rouzwawi:
Also lower the max backoff exponent to 4, which gives us a max retry time of 2^4*3m = 48m.


bergman:
what is the plan for pluggable retryutil?

i-maravic:
+1

rouzwawi:
@bergman to make it easier to make configuration changes without having to release a new version.

bergman:
👍

PR #41, opened 2017-01-18
Backfills

bergman:


styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -205,6 +215,64 @@ private void evaluateResourcesForDequeue(
...
+      final Workflow workflow = workflowOpt.get();
+      final int active = backfillStates.getOrDefault(backfill.id(), 0L).intValue();
+      final int needed = backfill.concurrency() - active;
+      if (needed <= 0) {
+        return;
+      }
+
+      final Instant start = backfill.lastTrigger().orElse(backfill.start());
+      final List<Instant> partitions =
+          ParameterUtil.rangeOfInstants(start, backfill.end(), workflow.schedule().partitioning())
```
rouzwawi: I think it would be safest to persist the partitioning in the Backfill entity rather than reading it form the workflow. If the partitioning changes during a backfill, this will lead to surprising behaviour.

bergman: sure

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -205,6 +215,64 @@ private void evaluateResourcesForDequeue(
...
+              .stream()
+              .limit(needed + 1)
+              .collect(toList());
+
+      if (partitions.isEmpty()) {
+        return;
+      }
+
+      final int numTrigger = Math.min(Math.min(needed, partitions.size()), backfill.concurrency());
```
rouzwawi: `needed` is always smaller than `backfill.concurrency()` so the outer min should not be needed?

bergman: Thanks! I need to take another pass at this section, it's very messy right now.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -205,6 +215,64 @@ private void evaluateResourcesForDequeue(
         .map(resource -> ResourceWithInstance.create(resource, instanceState));
   }
 
+  @VisibleForTesting
+  void triggerBackfills(Collection<InstanceState> activeStates) {
+    final List<Backfill> backfills;
+    try {
+      backfills = storage.backfills();
+    } catch (IOException e) {
+      LOG.warn("Failed to get resource limits", e);
```
fabriziodemaria: This should say `Failed to get backfills data` or similar.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -205,6 +215,64 @@ private void evaluateResourcesForDequeue(
         .map(resource -> ResourceWithInstance.create(resource, instanceState));
   }
 
+  @VisibleForTesting
+  void triggerBackfills(Collection<InstanceState> activeStates) {
+    final List<Backfill> backfills;
```
fabriziodemaria: Can we filter out backfills that are already completed at this preliminary stage? So we won't scan the cache for very old Workflows belonging to very old backfills which might not be in the cache anymore.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -205,6 +215,64 @@ private void evaluateResourcesForDequeue(
...
+              .collect(toList());
+
+      if (parameters.isEmpty()) {
+        return;
+      }
+
+      final List<Instant> triggers = parameters.subList(0, Math.min(needed, parameters.size()));
+
+      triggers.forEach(
+          partition -> triggerListener.event(workflow, backfill.id(), partition));
```
fabriziodemaria: This will throw runtime `AlreadyInitializedException` if already active, we should probably catch that, logging the event internally and proceed. `--force-halt` is out of scope for this PR, right?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -205,6 +216,64 @@ private void evaluateResourcesForDequeue(
...
+              .collect(toList());
+
+      if (parameters.isEmpty()) {
+        return;
+      }
+
+      final List<Instant> triggers = parameters.subList(0, Math.min(needed, parameters.size()));
+
+      triggers.forEach(
+          partition -> triggerListener.event(workflow, Trigger.backfill(backfill.id()), partition));
```
fabriziodemaria: This will throw runtime `AlreadyInitializedException` if already active, we should probably catch that, logging the event internally and proceed. `--force-halt` is out of scope for this PR, right?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -205,6 +216,64 @@ private void evaluateResourcesForDequeue(
...
+        .collect(groupingBy(
+            state -> state.runState().data().triggerId().orElse("UNKNOWN"),
+            HashMap::new,
+            counting()));
+
+    backfills.forEach(backfill -> {
+      final Optional<Workflow> workflowOpt = workflowCache.workflow(backfill.workflowId());
+
+      if (!workflowOpt.isPresent()) {
+        LOG.warn("workflow not found for backfill: {}", backfill);
```
fabriziodemaria: This doesn't seem like something that will be disappear in the next tick. Should we advance `nextTrigger` in this case? In the future we can store more information about the status of each trigger to explicitly point out to users that the issue with this was missed information from the schedule source.

bergman: good point

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -0,0 +1,288 @@
...
+            : Optional.empty()))
+        .collect(toList());
+    return BackfillsPayload.create(backfillPayloads);
+  }
+
+  private Response<BackfillPayload> getBackfill(String id) {
+    try {
+      Optional<Backfill> backfillOpt = storage.backfill(id);
+      if (backfillOpt.isPresent()) {
+        List<RunStateData> statues = retrieveBackfillStatuses(backfillOpt.get());
```
bergman: typo 🗽

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -0,0 +1,288 @@
...
+    )
+        .map(r -> r.withMiddleware(Middleware::syncToAsync))
+        .collect(toList());
+
+    return Api.prefixRoutes(routes, V1);
+  }
+
+  private BackfillsPayload getBackfills(RequestContext requestContext) {
+    final Optional<String> componentOpt = requestContext.request().parameter("component");
+    final Optional<String> statusesFlagOpt = requestContext.request().parameter("status");
```
bergman: this flag could be more explicit, `include_status` or something, just the word `status` doesn't really say enough: is it filtering, inclusion of status, what values are possible for the parameter, would `?status=false` or `?status=0` disable the statuses?

bergman:
👍

bergman:
🍾

PR #42, opened 2017-01-20
warm up cache from storage

honnix:
During starting up, sheduler may start reading cache before it is
populated from storage, which will lead to workflow not pairing
with resource.

Before starting scheduler and sources, read existing workflows from
storage and warm up the cache.


rouzwawi:
👍

PR #43, opened 2017-01-23
Add Trigger ADT

fabriziodemaria:
The first usage of this will be to filter active states in the StateManager whose trigger is of kind `backfill` (related to #41).

For now, this PR will just set all new triggers to `unknown` type and parse old trigger data to `unknown` type.


styx-common/src/main/java/com/spotify/styx/model/TriggerVisitor.java
```diff
@@ -0,0 +1,34 @@
...

+ * See the License for the specific language governing permissions and

+ * limitations under the License.

+ * -/-/-

+ */
+
+package com.spotify.styx.model;
```
rouzwawi: Shouldn't this rather be in the `com.spotify.styx.state` package?

fabriziodemaria:
The test that replays the active states using our stored data in production succeeded.

rouzwawi:
If I understand correctly, some of the classes (Execution, Trigger) were renamed to Data because they collided with the new Trigger class, and to keep the naming uniform. But since it's in the new package, should those renames be reverted?

styx-common/src/test/java/com/spotify/styx/model/data/WorkflowInstanceExecutionDataTest.java
```diff
@@ -176,18 +184,24 @@ public void shouldDeserializeExecutionData() throws Exception {
...
                         Arrays.asList(
-                            ExecStatus.create(Instant.parse("2016-08-03T10:56:03.607Z"), "STARTED"),
-                            ExecStatus.create(Instant.parse("2016-08-03T10:57:03.607Z"), "RUNNING"),
-                            ExecStatus.create(Instant.parse("2016-08-03T10:58:03.607Z"), "SUCCESS")
+                            ExecStatus
+                                .create(Instant.parse("2016-08-03T10:56:03.607Z"), "STARTED"),
+                            ExecStatus
+                                .create(Instant.parse("2016-08-03T10:57:03.607Z"), "RUNNING"),
+                            ExecStatus
+                                .create(Instant.parse("2016-08-03T10:58:03.607Z"), "SUCCESS")
```
rouzwawi: It would be nice if we could avoid whitespace changes like this in the future.

rouzwawi:
LGTM 👍

PR #45, opened 2017-01-25
Define Trigger type for trigger events

fabriziodemaria:
This PR also adds `backfillActiveStates` method to the StateManager.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/state/TriggerSerializer.java
+++ b/styx-common/src/main/java/com/spotify/styx/state/TriggerSerializer.java
@@ -59,32 +59,32 @@ public class TriggerSerializer {
   private static class TriggerSerializerVisitor implements TriggerVisitor<PersistentTrigger> {
 
     @Override
-    public PersistentTrigger natural(String triggerId) {
-      return new PersistentTrigger("natural", triggerId);
+    public PersistentTrigger natural() {
+      return new PersistentTrigger("natural");
     }
 
     @Override
     public PersistentTrigger adhoc(String triggerId) {
-      return new PersistentTrigger("adhoc", triggerId);
+      return new PersistentTrigger.Adhoc(triggerId);
     }
 
     @Override
     public PersistentTrigger backfill(String triggerId) {
-      return new PersistentTrigger("backfill", triggerId);
+      return new PersistentTrigger.Backfill(triggerId);
     }
 
     @Override
     public PersistentTrigger unknown(String triggerId) {
-      return new PersistentTrigger("unknown", triggerId);
+      return new PersistentTrigger.Unknown(triggerId);
     }
   }
 
   @JsonTypeInfo(use = Id.NAME, visible = true)
   @JsonSubTypes({
       @JsonSubTypes.Type(value = PersistentTrigger.class, name = "natural"),
-      @JsonSubTypes.Type(value = PersistentTrigger.class, name = "adhoc"),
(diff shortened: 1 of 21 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -84,7 +85,10 @@ void tick() {
 
       if (enabled.contains(workflow.id())) {
         try {
-          triggerListener.event(workflow, NATURAL_TRIGGER, decrementInstant(next, partitioning));
+          triggerListener.event(
+              workflow,
+              Trigger.natural(NATURAL_TRIGGER),
```
rouzwawi: It seems like the `natural` trigger should not have any `triggerId` field. It's always going to be this constant.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -142,6 +144,15 @@ public void receive(Event event) throws IsClosed {
   }
 
   @Override
+  public Map<WorkflowInstance, RunState> backfillActiveStates() {
+    final ImmutableMap.Builder<WorkflowInstance, RunState> builder = ImmutableMap.builder();
+    states.entrySet().stream()
+        .filter(entry -> isBackfill(entry.getValue().runState.data().trigger().get().toTrigger()))
```
rouzwawi: Potential exception thrown here because of the unchecked `.get()` on the optional.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -142,6 +144,15 @@ public void receive(Event event) throws IsClosed {
...
+        .forEach(entry -> builder.put(entry.getKey(), entry.getValue().runState));
```
rouzwawi: Collecto to a map rather than side-effecting foreach to builder.

styx-scheduler-service/src/main/java/com/spotify/styx/state/StateManager.java
```diff
@@ -64,6 +64,12 @@
   Map<WorkflowInstance, RunState> activeStates();
 
   /**
+   * Get a map of all active {@link WorkflowInstance} states whose issuing {@link Trigger} is of
+   * kind "backfill".
+   */
+  Map<WorkflowInstance, RunState> backfillActiveStates();
```
rouzwawi: I'm not sure this is the state managers concern? Should it be internal to the scheduler?

styx-common/src/main/java/com/spotify/styx/state/TriggerSerializer.java
```diff
@@ -59,8 +60,8 @@ public Trigger deserialize(ByteString json) {
   private static class TriggerSerializerVisitor implements TriggerVisitor<PersistentTrigger> {
 
     @Override
-    public PersistentTrigger natural(String triggerId) {
-      return new PersistentTrigger("natural", triggerId);
+    public PersistentTrigger natural() {
+      return new PersistentTrigger("natural", NATURAL_TRIGGER_ID);
```
rouzwawi: The trigger id is ignored in this case right?

The json produced for the natural trigger should just be `{'@type':'natural'}`.

bergman:
👍

PR #46, opened 2017-01-26
Add activeStates metrics per state

fabriziodemaria:

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -89,6 +89,7 @@ import java.io.IOException;
 import java.security.GeneralSecurityException;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -435,7 +436,6 @@ public class StyxScheduler implements AppInit {
       Stats stats) {
 
     final Gauge<Long> queuedEventsCount = stateManager::getQueuedEventsCount;
-    final Gauge<Long> activeStatesCount = stateManager::getActiveStatesCount;
     final Gauge<Long> allWorkflowsCount = () -> workflowCache.all().stream().count();
     final Gauge<Long> configuredWorkflowsCount = () -> workflowCache.all().stream()
         .filter(WorkflowValidator::hasDockerConfiguration)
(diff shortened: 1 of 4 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -453,8 +453,11 @@ private void setupMetrics(
       }
     };
 
+    Arrays.stream(RunState.State.values()).forEach(state -> stats.registerActiveStates(
```
bergman: doesn't `values()` have a `forEach`?

`RunState.State.values().foreach(state -> ...`

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -435,7 +436,6 @@ private void setupMetrics(
       Stats stats) {
 
     final Gauge<Long> queuedEventsCount = stateManager::getQueuedEventsCount;
-    final Gauge<Long> activeStatesCount = stateManager::getActiveStatesCount;
```
bergman: I would keep the old one for now, we can remove it later

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
fabriziodemaria: That doesn't seem to work with enum...

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
fabriziodemaria: The same graph achieved by the old one can still be constructed from the new one I believe

rouzwawi: Agree

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
bergman: ah, didn't notice it was an array

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -122,8 +125,13 @@ public void registerQueuedEvents(Gauge<Long> queuedEventsCount) {
   }
 
   @Override
-  public void registerActiveStates(Gauge<Long> activeStatesCount) {
-    registry.register(ACTIVE_STATES, activeStatesCount);
+  public void registerActiveStates(RunState.State state, Gauge<Long> activeStatesCount) {
+    registry.register(ACTIVE_STATES_PER_RUNSTATE, activeStatesCount);
```
rouzwawi: This line will register different gauges for the same metricId. That will most likely lead to confusing results.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -86,6 +87,7 @@
   private final ConcurrentMap<String, Histogram> dockerOperationHistograms;
   private final ConcurrentMap<String, Meter> dockerOperationMeters;
   private final ConcurrentHashMap<WorkflowId, Gauge> activeStatesPerWorkflowGauges;
+  private final ConcurrentHashMap<RunState.State, Gauge> activeStatesPerRunStateGauges;
```
rouzwawi: This map is not needed as the set of states is always static and we set up gauges on startup

rouzwawi:
LGTM

PR #47, opened 2017-01-26
Simplify and isolate jackson serialization types

rouzwawi:


fabriziodemaria:
Nice! 👍

bergman:
omg

rouzwawi:
The main thing here is `TypeWrapperModule` which enabled `Event` and `Trigger` to directly be (de)serialized without having to use `Persistent(Event|Trigger)` in the rest of the codebase.

bergman:
👍

PR #48, opened 2017-01-26
check global enabled status

honnix:
check global enabled status before triggering all workflows, instead of
checking for each workflow.

this will also force ad-hoc triggering regardless of enabled or not.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -312,7 +312,8 @@ public class StyxScheduler implements AppInit {
     final StateFactory stateFactory =
         (workflowInstance) -> RunState.fresh(workflowInstance, time, outputHandlers);
 
-    final TriggerListener trigger = trigger(storage, stateFactory, stateManager);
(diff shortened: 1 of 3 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -460,28 +461,6 @@ private void setupMetrics(
...
-  private TriggerListener trigger(
-      Storage storage,
-      StateFactory stateFactory,
-      StateManager stateManager) {
-    final TriggerListener stateInitializingTrigger =
-        new StateInitializingTrigger(stateFactory, stateManager, storage);
-
-    return (workflow, trigger, instant) -> {
-      try {
-        final boolean enabled = storage.workflowState(workflow.id()).enabled().orElse(false);
```
rouzwawi: Looks like this per-workflow enable check was dropped? This would lead to disabled workflows triggering from the TriggerManager.

honnix:
As I understood TriggerManager does the "enabled" checking already before signaling TriggerListener.

Best regards,
Hongxin

rouzwawi:
Ah, you're right. Then disregard my comment.

PR #49, opened 2017-01-26
Replace metering decorators with proxy

rouzwawi:


styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MeteredProxy.java
```diff
@@ -0,0 +1,76 @@
...
+
+    return (T) Proxy.newProxyInstance(
+        clazz.getClassLoader(),
+        new Class[]{clazz},
+        handler);
+  }
+
+  @Override
+  public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+    // todo: special case for store(Workflow workflow) -> storeWorkflow
```
bergman: so, rename that function or pass argument type list to instumentation?

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MeteredProxy.java
```diff
@@ -0,0 +1,76 @@
...
+    final String operation = method.getName();
+
+    final Instant t0 = time.get();
+    final Object ret = method.invoke(delegate, args);
+    final long durationMillis = t0.until(time.get(), ChronoUnit.MILLIS);
+
+    if (delegate instanceof Storage) {
```
honnix: Any thoughts subclassing `MetricsStats` so we don't need to do `instanceof` here?

rouzwawi: I think this check is lightweight enough over having to split up those methods into different classes.

honnix: True, good enough now.

PR #51, opened 2017-02-02
Log workflow instance together with pod status

honnix:
For debugging purpose we might need to know on which host the workflow
instance is running. Host IP is included in pod status.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -291,14 +291,7 @@ class KubernetesDockerRunner implements DockerRunner {
         return;
       }
 
-      final String podName = pod.getMetadata().getName();
-      LOG.info("Pod event for {} at resource version {}", podName, lastResourceVersion);
-      LOG.info("Action: {}", action);
-      try {
-        LOG.info("Status: {}", OBJECT_MAPPER.writeValueAsString(pod.getStatus()));
-      } catch (JsonProcessingException e) {
-        LOG.info("Status: {}", pod.getStatus());
-      }
+      logEvent(action, pod);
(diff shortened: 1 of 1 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -309,6 +302,20 @@ public void eventReceived(Action action, Pod pod) {
       }
     }
 
+    private void logEvent(Action action, Pod pod) {
+      final String podName = pod.getMetadata().getName();
+      LOG.info("Pod event for {} at resource version {}", podName, lastResourceVersion);
+      LOG.info("Action: {}", action);
+      String workflowInstance = pod.getMetadata().getAnnotations()
```
bergman: final

TinaRanic:
minor thing, otherwise LGTM 👍

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -309,6 +302,20 @@ public void eventReceived(Action action, Pod pod) {
       }
     }
 
+    private void logEvent(Action action, Pod pod) {
+      final String podName = pod.getMetadata().getName();
+      LOG.info("Pod event for {} at resource version {}", podName, lastResourceVersion);
+      LOG.info("Action: {}", action);
+      final String workflowInstance = pod.getMetadata().getAnnotations()
+              .getOrDefault(KubernetesDockerRunner.STYX_WORKFLOW_INSTANCE_ANNOTATION, "N/A");
+      LOG.info("Workflow instance: {}", workflowInstance);
```
honnix: Thanks! The "Status" is already quite messy now, so I thought it would be better having them separated.

PR #52, opened 2017-02-10
[WAITING] remove apollo from cli

bergman:
- saves about 5 seconds on every call
- synchronous calls makes the code much easier to follow IMHO

### DIFF
--- a/styx-cli/pom.xml
+++ b/styx-cli/pom.xml
@@ -18,19 +18,6 @@
     </dependency>
 
     <dependency>
-      <groupId>com.spotify</groupId>
-      <artifactId>apollo-core</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.spotify</groupId>
-      <artifactId>apollo-environment</artifactId>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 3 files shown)


rouzwawi:
Cool. Out of curiosity, what made the calls take 5 seconds longer? It's surprising given that the underlying Apollo Client uses the same HTTP client.

bergman:
Yes, the 5 second delay was probably due to some other problem and is not usually there. I still kind of like the synchronous model better than the other one for these simple cases where it's just one call.

rouzwawi:
@bergman Can' that be achieved with just?

```
Response<ByteString> response = client.send(...).toCompletableFuture().get()
```

bergman:
You would still have the signaller and service thing which just feels odd in a temporary client.

rouzwawi:
The signaller was only needed because the calls where async. If the main thread blocks on the calls, the top level try-with-resource block does not have to block on `instance.waitForShutdown()`, so we can remove all the `Signaller` code.

bergman:
You're right, will have another pass at this later.

PR #53, opened 2017-02-12
Inspect Pod status for missing PodIP

fabriziodemaria:


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesPodEventTranslator.java
```diff
@@ -141,6 +141,10 @@ private KubernetesPodEventTranslator() {
 
       case "Succeeded":
       case "Failed":
+        if ("".equals(pod.getStatus().getPodIP())) {
```
honnix: Is it an empty string or null in this case?

rouzwawi: Maybe better to check with `Strings.isNullOrEmpty()` from Guava

honnix:
Shall we close this PR?

PR #56, opened 2017-02-16
Global resource

bergman:


styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
```diff
@@ -337,6 +337,22 @@ public void shouldFailWhenUnknownResourceReference() throws Exception {
   }
 
   @Test
+  public void shouldIssueInfoIfGlobalResourceDepleted() throws Exception {
```
TinaRanic: shall we also test the fallback, when the global resource doesn't exist in the storage?

bergman: good suggestion, done and pushed :)

TinaRanic:
👍 LGTM

bergman:
I thought a bit and am not so sure anymore that the default case should be to limit if not set. I think I would rather keep the original behaviour of running without any resource if not set.

styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
bergman: went back on this, no fallback, so no test.

fabriziodemaria:
Close this? @bergman

PR #58, opened 2017-02-22
quote message to ease print out processing

honnix:


honnix:
No fix. Things can be done by counting column, plus quoting looks confusing.

Issue #6, opened 2016-11-24
Inconsistent patchState behaviour


fabriziodemaria:
If a user patches the state for a Workflow/Component calling the related API, the following can happen:

- At first, the user patches the Docker image for the specific pair `Component/Workflow`;
- Afterwards, the user patches the Docker image for the `Component`, expecting it to be picked up by all the Workflows within the Component;

The Docker image at Workflow level is always returned first when preparing executions, so the new configuration at Component level will not be picked up.

ukarlsson:
We have a use case in bq-runner (https://ghe.spotify.net/bigquery/bigquery-runner) that is very similar to bq-load and in this use case `docker_image` in `data-info.yaml` is very useful so we are a bit worried about the deprecation of this field. In addition, we would also like to have a way of clearing the HTTP posted docker image since it overrides `data-info.yaml`. I actually think it would make more sense to have `data-info.yaml` override the HTTP post, and if the precedence was changed in this manner, we would not need to remove the HTTP posted version.

fabriziodemaria:
After ##318 #296 this issue should be now resolved. The only way to set a Docker image is now via the API call for a specific workflow.

PR #60, opened 2017-02-22
re-prioritize docker image specs

TinaRanic:
This PR changes the order in which a docker image specification takes precedence:

old way:
1. docker image as part of state in workflow instance
2. docker image as part of state in component
3. docker image specified in workflow definition (yaml)

new way:
1. docker image specified in workflow definition (yaml)
2. docker image as part of state in workflow instance
3. docker image as part of state in component

The reason for this is that many customers find it more intuitive that a docker image specification in the yaml file should override everything else.

ping @kanterov @Xeago @fabriziodemaria @bergman @honnix

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -359,19 +359,20 @@ class DatastoreStorage {
   }
 
   Optional<String> getDockerImage(WorkflowId workflowId) throws IOException {
+    Optional<String> dockerImage =
+        workflow(workflowId).flatMap(wf -> wf.schedule().dockerImage());
+    if (dockerImage.isPresent()) {
+      return dockerImage;
+    }
+
     final Key workflowKey = workflowKey(workflowId);
-    Optional<String> dockerImage = getOptStringProperty(datastore, workflowKey, PROPERTY_DOCKER_IMAGE);
+    dockerImage = getOptStringProperty(datastore, workflowKey, PROPERTY_DOCKER_IMAGE);
     if (dockerImage.isPresent()) {
       return dockerImage;
     }
 
     final Key componentKey = componentKeyFactory.newKey(workflowId.componentId());
-    dockerImage = getOptStringProperty(datastore, componentKey, PROPERTY_DOCKER_IMAGE);
-    if (dockerImage.isPresent()) {
-      return dockerImage;
-    }
-
-    return workflow(workflowId).flatMap(wf -> wf.schedule().dockerImage());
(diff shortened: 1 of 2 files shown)


styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -359,19 +359,20 @@ void patchState(String componentId, WorkflowState state) throws IOException {
   }
 
   Optional<String> getDockerImage(WorkflowId workflowId) throws IOException {
-    final Key workflowKey = workflowKey(workflowId);
-    Optional<String> dockerImage = getOptStringProperty(datastore, workflowKey, PROPERTY_DOCKER_IMAGE);
+    Optional<String> dockerImage =
```
bergman: Could this variable be named such that it reflects where the value comes from? Something like

```java
final Optional<String> workflowDefinitionDockerImage = ...
```

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -359,19 +359,20 @@ void patchState(String componentId, WorkflowState state) throws IOException {
...
+        workflow(workflowId).flatMap(wf -> wf.schedule().dockerImage());
     if (dockerImage.isPresent()) {
       return dockerImage;
     }
 
-    final Key componentKey = componentKeyFactory.newKey(workflowId.componentId());
-    dockerImage = getOptStringProperty(datastore, componentKey, PROPERTY_DOCKER_IMAGE);
+    final Key workflowKey = workflowKey(workflowId);
+    dockerImage = getOptStringProperty(datastore, workflowKey, PROPERTY_DOCKER_IMAGE);
```
bergman: Same as above

fabriziodemaria:
Regarding the case in which the docker image is not specified in the workflow definition, we still have a case in which a user starts by posting a docker image per workflow via the API, and later decides to post it for the entire component: in this case, workflows for which an image was previously posted won't pick up the new one posted at the component level. That might not be the wanted behaviour.

I believe we should improve the API to be able to get what has been posted and at which level for a workflow/component, as well as giving the possibility to delete a posted image for maximum flexibility (also via the CLI).

Just to note this down: this change is not backwards compatible.

Fixes #6.

honnix:
Agree with @fabriziodemaria , this is not backward compatible. We will need to communicate in a good way.

ulzha:
It looks like we'll be able to clean the design up and have one single way of setting Docker image, that is, the workflow level API endpoint: https://github.com/spotify/styx/pull/243. So the entire concept of "prioritization" in Styx will disappear. It was accidental complexity.

PR #62, opened 2017-02-24
use parallelStream

honnix:
sequential execution is too slow for backfill with many partitions

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
@@ -126,7 +126,7 @@ public final class BackfillResource {
           .filter(backfill -> backfill.workflowId().endpointId().equals(workflow));
     }
 
-    final List<BackfillPayload> backfillPayloads = backfills
+    final List<BackfillPayload> backfillPayloads = backfills.parallel()
         .map(backfill -> BackfillPayload.create(
             backfill,
             "true".equals(statusesFlagOpt.orElse("false"))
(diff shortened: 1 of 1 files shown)


styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -109,7 +109,7 @@ private BackfillsPayload getBackfills(RequestContext requestContext) {
 
     Stream<Backfill> backfills;
     try {
-      backfills = storage.backfills().stream();
+      backfills = storage.backfills().parallelStream();
```
bergman: Could this be moved to where the intensive part is? It doesn't really matter for the filtering, right? We're even moving that to datastore at some point.

honnix: It's definitely not needed for filtering, but it looks both component and workflow are optional, so in the worst case we might end up with processing all backfills.

The other change is for sure the intensive part as we have identified.

fabriziodemaria:
LGTM

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
bergman: I mean that you could move `.parallelStream()` to just before line 130. I don't know, perhaps I'm being overly cautious about parallel streams.

honnix: OK I got it. Good point! I will fix it.

PR #65, opened 2017-02-28
halt workflow instances when halting backfill

TinaRanic:


styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -168,6 +183,30 @@ private BackfillsPayload getBackfills(RequestContext requestContext) {
     return Response.ok();
   }
 
+  private void haltActiveInstances(Backfill backfill, RequestContext rc) throws IOException {
```
fabriziodemaria: More specifically we can name this something like `haltActiveBackfillInstances`

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -168,6 +183,30 @@ private BackfillsPayload getBackfills(RequestContext requestContext) {
...
+        .filter(activeWorkflowInstances::contains)
+        .forEach(workflowInstance -> haltActiveInstance(workflowInstance, client));
+  }
+
+  private void haltActiveInstance(WorkflowInstance workflowInstance, Client client) {
+    try {
+      final ByteString payload = serialize(Event.halt(workflowInstance));
+      Request request = Request.forUri(schedulerApiUrl("events"), "POST")
+          .withPayload(payload);
+      client.send(request);
```
fabriziodemaria: This seems to be an asynchronous call and we don't track the outcome of the request? We should notify if some of the `halt` requests failed in the process

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -168,6 +183,30 @@ private BackfillsPayload getBackfills(RequestContext requestContext) {
...
+  private void haltActiveInstances(Backfill backfill, RequestContext rc) throws IOException {
+    final Set<WorkflowInstance> activeWorkflowInstances =
+        storage.readActiveWorkflowInstances(backfill.workflowId().componentId()).keySet();
+
+    final Client client = rc.requestScopedClient();
+
+    rangeOfInstants(backfill.start(), backfill.nextTrigger(), backfill.partitioning()).stream()
+        .map(instant -> WorkflowInstance
+            .create(backfill.workflowId(), toParameter(backfill.partitioning(), instant)))
+        .filter(activeWorkflowInstances::contains)
```
fabriziodemaria: We should add a filter for comparing the backfill id as well

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -329,23 +329,30 @@ public void shouldUpdateBackfill() throws Exception {
...
     sinceVersion(Api.Version.V1);
 
-    assertThat(storage.backfill(BACKFILL_1.id()).get().concurrency(), equalTo(1));
+    WorkflowInstance wfi = WorkflowInstance.create(BACKFILL_1.workflowId(), "2017-01-01T01");
+    storage.storeBackfill(BACKFILL_1.builder().nextTrigger(Instant.parse("2017-01-01T02:00:00Z")).build());
+    /*storage.writeEvent(SequenceEvent.create(Event.triggerExecution(wfi, Trigger.backfill("backfill-1")), 1L, 1L));
+    storage.writeEvent(SequenceEvent.create(Event.dequeue(wfi),                                          2L, 2L));
+    storage.writeEvent(SequenceEvent.create(Event.submit(wfi, TestData.EXECUTION_DESCRIPTION),           3L, 3L));
+    storage.writeEvent(SequenceEvent.create(Event.submitted(wfi, "exec-1"),                              4L, 4L));
+    storage.writeEvent(SequenceEvent.create(Event.started(wfi),                                          5L, 5L));*/
```
fabriziodemaria: Unused comments

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -289,7 +291,7 @@ public void shouldFailOnMisalignedRange() throws Exception {
 
   @Test
   public void shouldFailOnAlreadyActiveWithinRange() throws Exception {
-    sinceVersion(Api.Version.V1);
+
```
fabriziodemaria: Let's not add additional empty lines in these tests

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -168,6 +186,43 @@ private BackfillsPayload getBackfills(RequestContext requestContext) {
     return Response.ok();
   }
 
+  private void haltActiveBackfillInstances(Backfill backfill, Client client) throws IOException {
+    final List<RunStateData> activeRunStateData = retrieveBackfillStatuses(backfill).stream()
+        .filter(this::activeState).collect(toList());
+    if (activeRunStateData.size() != 0) {
+      final boolean failed =
+          activeRunStateData.parallelStream().map(RunStateData::workflowInstance)
```
bergman: put the `.map` on a new line, makes it easier to see each operation

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -168,6 +186,43 @@ private BackfillsPayload getBackfills(RequestContext requestContext) {
     return Response.ok();
   }
 
+  private void haltActiveBackfillInstances(Backfill backfill, Client client) throws IOException {
+    final List<RunStateData> activeRunStateData = retrieveBackfillStatuses(backfill).stream()
+        .filter(this::activeState).collect(toList());
+    if (activeRunStateData.size() != 0) {
```
bergman: why not just run the map anyway? map on an empty list is a noop anyway

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -168,6 +186,43 @@ private BackfillsPayload getBackfills(RequestContext requestContext) {
     return Response.ok();
```
bergman: do the successful return inside the try instead, it's harder to read this way

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -168,6 +186,43 @@ private BackfillsPayload getBackfills(RequestContext requestContext) {
...
+  }
+
+  private boolean haltActiveBackfillInstance(WorkflowInstance workflowInstance, Client client) {
+    try {
+      final ByteString payload = serialize(Event.halt(workflowInstance));
+      final Request request = Request.forUri(schedulerApiUrl("events"), "POST")
+          .withPayload(payload);
+      return client.send(request)
+          .thenApply(response -> response.status().family().equals(SUCCESSFUL))
+          .toCompletableFuture().get();
```
bergman: why not keep it async all the way?

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -329,18 +341,99 @@ public void shouldUpdateBackfill() throws Exception {
...
 
     assertThat(storage.backfill(BACKFILL_1.id()).get().halted(), equalTo(true));
+    verify(serviceHelper.stubClient(), times(1)).send(any());
+  }
+
+  @Test
+  public void shouldReturnServerErrorIfFailedToSend() throws Exception {
+    sinceVersion(Api.Version.V1);
+
+    serviceHelper.stubClient().respond(Responses.sequence(ImmutableList.of(ResponseWithDelay
```
bergman: nice find on the sequenced responses thing!

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
honnix: will fix

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
honnix: will fix

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
honnix: A little bit complicated in this case. It will become a CompletableFuture of a CompletableFuture, and exception handling becomes hard to read.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -152,20 +173,67 @@ private BackfillsPayload getBackfills(RequestContext requestContext) {
...
+      final Request request = Request.forUri(schedulerApiUrl("events"), "POST")
+          .withPayload(payload);
+      return client.send(request)
+          .thenApply(response -> response.status().family().equals(SUCCESSFUL));
+    } catch (JsonProcessingException e) {
+      return CompletableFuture.completedFuture(false);
+    }
+  }
+
+  private boolean activeState(RunStateData runStateData) {
```
bergman: This function could be static, right? And maybe naming it `isActiveState` is more clear.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -152,20 +173,67 @@ private BackfillsPayload getBackfills(RequestContext requestContext) {
...
+  }
+
+  private CompletionStage<Response<ByteString>> haltBackfill(String id, RequestContext rc) {
     try {
       final Optional<Backfill> backfillOptional = storage.backfill(id);
       if (backfillOptional.isPresent()) {
-        storage.storeBackfill(backfillOptional.get().builder().halted(true).build());
+        final Backfill backfill = backfillOptional.get();
+        storage.storeBackfill(backfill.builder().halted(true).build());
+        return haltActiveBackfillInstances(backfill, rc.requestScopedClient());
```
bergman: Do we want the backfill entity to have halted=true before we know if halting the workflow instances worked? I haven't thought about it too much so see this as a question and not a complaint.

honnix: We can notify use that some active instances cannot be halted, while no new ones will be triggered, in this case.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
honnix: good point, will fix.

bergman:
👍

PR #66, opened 2017-02-28
[CLI] Print backfills on one line each

bergman:
It does go a bit wide if component and workflow names are long but it's much easier to get an overview.


honnix:
A general comment. Don't we want some colour here? It seems they will all be black/white-ish.

honnix:
👍

PR #67, opened 2017-02-28
[CLI] refactor error handling

bergman:
building on top of the refactor in #64

### DIFF
--- a/styx-cli/src/main/java/com/spotify/styx/cli/Main.java
+++ b/styx-cli/src/main/java/com/spotify/styx/cli/Main.java
@@ -21,16 +21,17 @@
 package com.spotify.styx.cli;
 
 import static com.spotify.styx.serialization.Json.OBJECT_MAPPER;
+import static com.spotify.styx.serialization.Json.deserialize;
 import static com.spotify.styx.serialization.Json.serialize;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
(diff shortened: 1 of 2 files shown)


styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -183,19 +185,30 @@ private void run() throws IOException, InterruptedException {
...
+          throw new ArgumentParserException("Unrecognized command: " + command, parser.parser);
       }
-      instance.getSignaller().signalShutdown();
+    } catch (ArgumentParserException e) {
+      parser.parser.handleError(e);
+      System.exit(EXIT_CODE_ARGUMENT_ERROR);
+    } catch (InterruptedException | ExecutionException e) {
+      e.printStackTrace();
+      System.exit(EXIT_CODE_UNKNOWN_ERROR);
+    } catch (Exception e) {
```
fabriziodemaria: We could better handle Json exception, but not a blocker for this PR

styx-common/src/main/java/com/spotify/styx/serialization/Json.java
```diff
@@ -65,6 +65,10 @@ public static ByteString serialize(Object value) throws JsonProcessingException
     return ByteString.of(OBJECT_MAPPER.writeValueAsBytes(value));
   }
 
+  public static <T> T deserialize(ByteString json, Class<T> clazz) throws IOException {
```
fabriziodemaria: Should we remove the two following methods `deserializeEvent` and `deserializeTrigger` to use this generic one only?

bergman: Yes, in another PR 😃

PR #69, opened 2017-03-02
List only active backfills by default

bergman:


styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -562,7 +571,7 @@ public Subparser parser(Subparsers subCommands) {
   }
 
   private enum BackfillCommand {
-    LIST("ls", "List backfills"),
+    LIST("ls", "List backfills, by default, only active backfills are printed. Use option -a (--show-all) to show all"),
```
honnix: A bit too long shown in the terminal. How about `List active backfills. Use option -a (--show-all) to show all`? Or we just rely user to do `styx backfill list -h` and just say `List active backfills`

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -548,8 +548,75 @@ void deleteResource(String id) throws IOException {
     return Optional.of(entityToBackfill(entity));
   }
 
-  List<Backfill> getBackfills() {
-    final EntityQuery query = Query.entityQueryBuilder().kind(KIND_BACKFILL).build();
+  List<Backfill> getBackfills(boolean showAll) {
```
honnix: These methods could likely be refactored by defining a private method taking in `queryBuilder` and doing the actual query; the other methods just call it.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -548,8 +548,75 @@ void deleteResource(String id) throws IOException {
...
+    final EntityQuery.Builder queryBuilder = Query.entityQueryBuilder().kind(KIND_BACKFILL);
+    if (!showAll) {
+      queryBuilder
```
honnix: could be extracted as a method

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
bergman: Agreed, taking your suggested text :)

bergman:
@honnix I made the changes you requested, it's much better now so thanks :)

honnix:
LGTM 👍

PR #70, opened 2017-03-03
[WIP] Add RateLimiter in Scheduler

fabriziodemaria:
This is one of two proposed solutions for implementing throttling of the rate of creation of Docker containers from the scheduler.
The other path to be investigated aims at adding the throttler in the DockerOutputHandler instead.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/Storage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/Storage.java
@@ -79,6 +79,11 @@ public interface Storage {
   boolean debugEnabled() throws IOException;
 
   /**
+   * Get the submission rate for Styx.
+   */
+  Optional<Double> submissionRate() throws IOException;
+
+  /**
    * Set the global enabled flag for Styx.
    *
    * @param enabled     The global enabled flag
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -34,6 +34,7 @@ import com.google.auto.value.AutoValue;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
+import com.google.common.util.concurrent.RateLimiter;
 import com.spotify.styx.model.Backfill;
 import com.spotify.styx.model.BackfillBuilder;
 import com.spotify.styx.model.Event;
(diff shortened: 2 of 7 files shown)


styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
```diff
@@ -324,6 +328,28 @@ public void shouldExecuteNewTriggers() throws Exception {
...
+    final RunState i0 = RunState.create(instance(WORKFLOW_ID1, "i0"), State.QUEUED, time);
+    init(i0);
+    scheduler.tick();
+
+    assertThat(stateManager.get(i0.workflowInstance()).state(), is(State.PREPARE));
+
+    final RunState i1 = RunState.create(instance(WORKFLOW_ID1, "i1"), State.QUEUED, time);
+    init(i1);
+    try {
+      await().atMost(Duration.ONE_SECOND).until(() -> scheduler.tick());
```
honnix: we might need ```fail``` here to make sure the exception is thrown out.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -88,25 +89,38 @@
...
     this.ttls = Objects.requireNonNull(ttls);
     this.stateManager = Objects.requireNonNull(stateManager);
     this.workflowCache = Objects.requireNonNull(workflowCache);
     this.storage = Objects.requireNonNull(storage);
     this.triggerListener = Objects.requireNonNull(triggerListener);
+    this.rateLimiter = Objects.requireNonNull(rateLimiter);
   }
 
   void tick() {
+    try {
```
honnix: a separated method maybe? ```tick()``` is getting longer. :)

styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
fabriziodemaria: Added

styx-common/src/main/java/com/spotify/styx/storage/Storage.java
```diff
@@ -79,6 +79,11 @@
   boolean debugEnabled() throws IOException;
 
   /**
+   * Get the submission rate for Styx.
```
ulzha: Not completely informative. It is "maximum rate" or something, not simply "rate". The documentation should describe units, too.

danielnorberg: "submission rate limit" ?

Issue #30, opened 2016-12-17
Weighted resource use, or plain concurrency limit


rouzwawi:
The current implementation of resource usage for a Workflow is through a plain resource reference in the [`DataEndpoint`](https://github.com/spotify/styx/blob/master/styx-common/src/main/java/com/spotify/styx/model/DataEndpoint.java#L54) definition. This has the problem that it only allows for a unit resource usage, per workflow instance.

In general, we need to align on what we mean with "resource" and "concurrency". We have to treat these two concepts separately:

- A general resource and resource usage mechanism (that we have today), need a better association mechanism from workflows to resources. There's however some amount of detail that has to be considered here
  - One workflow instance can use several resources (e.g. available memory or cpu quota in a GCP project), and these have different magnitudes and units (GB, count). The current model is backwards as it requires the resource limit to be set in a unit that is normalized to `1` unit used per workflow instance. Instead, a more natural mechanism would be for the workflow to define how many of each resource it will consume.
  - A workflow instance usually submits one or more processing jobs to various processing runtimes (Dataproc, Dataflow, Hadoop, etc), these might or might not happen concurrently in the workflow, depending on wiring. So a fixed resource use for the workflow instance is at best a "max resource use" definition.
- A concurrency limit on a workflow instance is a very simple limit that does not reflect any real resource usage. It leaves it completely up to the user to figure out how a workflow instance relates to some (unknown to Styx) resources, and from that derive a concurrency limit.

The second, more simplified concurrency limit can be reduced to use the more general resource/use mechanism. And we can do it as an internal detail that does not leak to the user.

I like that we are taking the first approach, but we need to be aware of the model we're dealing with and make it better reflected in the user touch points of Styx.

One immediate change that we need to do is to change the definition of the resource usage association in the workflow definition to be an object rather than a plain string:

```yaml
schedules:
  - id: example-workflow
    partitioning: hours
    resources:
      - id: nodes
        use: 32
```

The `use` field can default to 1 which is the current behaviour, but the schema having an object there will allow us to evolve the definition.

danielnorberg:
Seems like this might be an issue that we want to resolve soon.

An interesting question is whether it will be enough to trust the user-indicated resource usage values. As opposed to some other schedulers (e.g. borg) that are in control of the underlying job execution platform styx is currently not able to actually enforce the indicated resource usage limits. As preventing resource stock-outs in whole regions has become a pressing issue, maybe we need to think about how styx might gain this capability on e.g. gcp.

PR #71, opened 2017-03-03
[wip] Graph based scheduler

rouzwawi:
Opening this up for discussion. The work is not done.

The scheduler has logic that decides which workflow instances to to dequeue, taking resources into account. This PR replaces that logic with a graph flow problem (inspired by [firmament](http://firmament.io)). It is best described by an example:

graph3

The first layer after the source are all workflow instances eligible for dequeuing. Each of those are attached to a sequence of resource nodes with edges representing their limit (also reflected in the node name). There's one such path for each workflow, and they all end in the sink node.

If a resource is used by multiple workflows, the corresponding edge is aliased between the paths (dotted lines, which are not edges). Allocating flow along one of these will reduce capacity along all aliased edges.

Finally, all edges between the source and a workflow instance that have a non-zero flow will be dequeued.

Following paths from the source to the sink that have zero flow can give us a list of resources that were depleted.

## todo

- [ ] Add edge before sink for global resource limit
- [ ] Fix bug with backfills


honnix:
I like graph and optimisation!

```If a resource is used by multiple workflows, the corresponding edge is aliased between the paths (dotted lines, which are not edges). Allocating flow along one of these will reduce capacity along all aliased edges.```
Didn't quite get this. Is it an extension/variation of maximum flow problem?

How many different resource types we may have? This contributes to the complexity of the network. So far it seems we have pretty simple situation and maximum flow feels a little bit overshooting (I could be wrong by not knowing all history).

```N0_source``` seems to be a "virtual" node, the capacity can be set to infinity instead of 1, is that the case?

Another thing, not so important. Have we considered using something like https://github.com/jrtom/jung which feels more bullet-proof than home brew data structure and algorithm?

danielnorberg:
Just to clarify, would the purpose of this to be, like firmament offers, `find the policy-optimal assignment of work to cluster resources`?

I agree that the resource utilization problems we currently face would seem to benefit from an efficient bin-packing scheduler that can provide a more even data processing resource usage over time and eliminate hourly peaks etc.

However, one of the current vexing issues would seem to be that the actual resource usage of a data pipeline executed by styx is opaque to styx itself.

- Although it is possible for a user to annotate a workflow configuration with a list of consumed resources, it does not currently seem to be possible to indicate the magnitude of consumed resource. I.e., it does not seem possible to say that workflow `foo` consumes 100 cpu cores etc.

- Styx lacks facilities to enforce that workflows are annotated with some resource usage and e.g. deny/penalize workflows that claim to be "free".

- Styx is unable to use historical resource usage information to e.g. estimate resource usage of future workflow executions, etc.

- Etc...

Unless we address the above issues, would the graph scheduler actually produce high-quality scheduling descisions?

Would it make sense to start addressing these issues first?

rouzwawi:
The main purpose of this particular PR is to give us a more flexible foundation to build on. But as you've pointed out, it won't lead to any different scheduling as it stands right now. I've tried to scope this as implementing the same logic we have right now. Perhaps with the addition of doing multi-run solves for prioritizing natural trigger instances over backfills.

>Although it is possible for a user to annotate a workflow configuration with a list of consumed resources, it does not currently seem to be possible to indicate the magnitude of consumed resource. I.e., it does not seem possible to say that workflow foo consumes 100 cpu cores etc.

This has been brought up in issue #30

On the other issues; I agree that it would be good to start addressing those. But it should not be a stopper for this PR, right?

danielnorberg:
What is multi-run?

Yeah, these other issues might not be blockers for this PR but in practice it might mean that we have limited time to spend on the graph scheduler in the short term. And as it is seems like a non-insignificant increase in complexity, I would like to proceed with due care.

It might be beneficial to first focus on the pressing matter of preventing sharp resource usage peak induced stock-outs etc.

rouzwawi:
multi-run: e.g. configure the flow graph for priority 1 instances, solve, dequeue, then add p2 instances, solve on remaining capacity, dequeue, etc.

PR #72, opened 2017-03-03
read all active instances

honnix:
instead of reading active instances per component which will result
a lot of datastore interactions, we read everything one time.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
@@ -101,11 +101,11 @@ public final class ReplayEvents {
 
   public static Optional<RunState> getBackfillRunState(
       WorkflowInstance workflowInstance,
+      Map<WorkflowInstance, Long> activeWorkflowInstances,
       Storage storage,
       String backfillId) {
     final SettableTime time = new SettableTime();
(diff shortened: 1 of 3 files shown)


styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
```diff
@@ -119,16 +120,10 @@ private ReplayEvents() {
...
-          .readActiveWorkflowInstances(workflowInstance.workflowId().componentId());
-      if (instances.keySet().contains(workflowInstance)) {
-        lastConsumedEvent = instances.get(workflowInstance);
-      } else {
-        lastConsumedEvent = sequenceEvents.last().counter();
-      }
-    } catch (IOException e) {
-      throw Throwables.propagate(e);
+    if (activeWorkflowInstances.keySet().contains(workflowInstance)) {
+      lastConsumedEvent = activeWorkflowInstances.get(workflowInstance);
```
danielnorberg: No biggie but could merge the `contains()` and `get()` into a single `get()` or `getOrDefault()`.

honnix: good point! fixing...

danielnorberg:
LGTM 👍

PR #73, opened 2017-03-05
validate client version

honnix:
this will give server side a chance to reject request from certain
client version.

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/StyxApi.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/StyxApi.java
@@ -20,6 +20,7 @@
 
 package com.spotify.styx;
 
+import static com.spotify.styx.api.Middlewares.clientValidator;
 import static com.spotify.styx.util.Connections.createBigTableConnection;
 import static com.spotify.styx.util.Connections.createDatastore;
 import static java.util.Objects.requireNonNull;
@@ -31,14 +32,20 @@ import com.spotify.apollo.Environment;
 import com.spotify.apollo.route.Route;
 import com.spotify.styx.api.BackfillResource;
 import com.spotify.styx.api.CliResource;
+import com.spotify.styx.api.Middlewares;
 import com.spotify.styx.api.ResourceResource;
 import com.spotify.styx.api.StyxConfigResource;
 import com.spotify.styx.api.WorkflowResource;
 import com.spotify.styx.storage.AggregateStorage;
 import com.spotify.styx.storage.Storage;
+import com.spotify.styx.util.CachedSupplier;
 import com.spotify.styx.util.StorageFactory;
 import com.typesafe.config.Config;
 import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
 import org.apache.hadoop.hbase.client.Connection;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -100,13 +107,18 @@ public class StyxApi implements AppInit {
     final StyxConfigResource styxConfigResource = new StyxConfigResource(storage);
     final CliResource cliResource = new CliResource(schedulerServiceBaseUrl, storage);
 
+    final Supplier<Optional<List<String>>> clientBlacklistSupplier =
+        new CachedSupplier<>(storage::clientBlacklist, Instant::now);
+
     environment.routingEngine()
         .registerAutoRoute(Route.sync("GET", "/ping", rc -> "pong"))
         .registerRoutes(workflowResource.routes())
         .registerRoutes(resourceResource.routes())
-        .registerRoutes(backfillResource.routes())
+        .registerRoutes(backfillResource.routes()
+                            .map(r -> r.withMiddleware(clientValidator(clientBlacklistSupplier))))
         .registerRoutes(styxConfigResource.routes())
-        .registerRoutes(cliResource.routes());
+        .registerRoutes(cliResource.routes()
+                            .map(r -> r.withMiddleware(clientValidator(clientBlacklistSupplier))));
   }
 
   private static AggregateStorage storage(Environment environment) {
--- a/styx-api-service/src/main/java/com/spotify/styx/api/Middlewares.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/Middlewares.java
@@ -23,11 +23,17 @@ package com.spotify.styx.api;
 import static com.spotify.styx.serialization.Json.OBJECT_MAPPER;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.collect.ImmutableList;
 import com.spotify.apollo.Response;
 import com.spotify.apollo.Status;
 import com.spotify.apollo.route.AsyncHandler;
 import com.spotify.apollo.route.Middleware;
 import com.spotify.apollo.route.SyncHandler;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Supplier;
 import okio.ByteString;
 
 /**
@@ -49,7 +55,7 @@ public final class Middlewares {
       jsonAsync() {
     return innerHandler -> innerHandler.map(response -> {
       if (!response.payload().isPresent()) {
-        //noinspection unchecked
+        // noinspection unchecked
         return (Response<ByteString>) response;
       }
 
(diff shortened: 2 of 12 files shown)


styx-api-service/src/main/java/com/spotify/styx/api/CliResource.java
```diff
@@ -84,7 +84,7 @@ public CliResource(String schedulerServiceBaseUrl, Storage storage) {
             "GET", BASE + "/events/<cid>/<eid>/<iid>",
             rc -> eventsForWorkflowInstance(arg("cid", rc), arg("eid", rc), arg("iid", rc))))
 
-        .map(r -> r.withMiddleware(Middleware::syncToAsync))
+        .map(r -> r.withMiddleware(h -> Middleware.syncToAsync(h)))
```
danielnorberg: is this change necessary?

honnix: not really, didn't realise I changed it!

styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
```diff
@@ -214,4 +214,9 @@ public void storeResource(Resource resource) throws IOException {
   public Optional<Backfill> backfill(String id) throws IOException {
     return datastoreStorage.getBackfill(id);
   }
+
+  @Override
+  public List<String> clientBlacklist() {
+    return datastoreStorage.clientBlacklist();
```
danielnorberg: Is it likely that we'll need to update this list at runtime? If not, could it just be a "hardcoded" list either in code or the config file?

honnix: I think we will need to update this list at runtime. There was a case that we released a faulty CLI but the server side had no problem.

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -363,7 +365,9 @@ private String apiUrl(CharSequence... parts) {
   }
 
   private ByteString send(Request request) throws ExecutionException, InterruptedException {
-    final Response<ByteString> response = client.send(request).toCompletableFuture().get();
+    final Response<ByteString> response =
+        client.send(request.withHeader("Client-Version", STYX_CLI_VERSION)).toCompletableFuture()
```
danielnorberg: Should we include "Styx" in the header? E.g. `Styx-Client-Version: Styx CLI 1.2.3`. 

Alternatively, would it perhaps make sense to use the `User-Agent` header? E.g. `User-Agent: Styx CLI 1.2.3`.

styx-common/src/main/java/com/spotify/styx/util/CachedSupplier.java
```diff
@@ -18,10 +18,9 @@
  * -/-/-
  */
 
-package com.spotify.styx;
+package com.spotify.styx.util;
```
danielnorberg: Would it make sense to use the Guava memoizing supplier instead of rolling our own?

http://google.github.io/guava/releases/21.0/api/docs/com/google/common/base/Suppliers.html#memoizeWithExpiration-com.google.common.base.Supplier-long-java.util.concurrent.TimeUnit-

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
honnix: Would be better to use ```User-Agent``` indeed. I wasn't sure how Apollo will deal with this header, hopefully not rewriting it. I will do some experiment or read doc.

styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
danielnorberg: Ok, I guess we could just deploy a new version of the server then that blacklists that client, but I don't have a very strong opinion.

danielnorberg:
This PR seems to include PR https://github.com/spotify/styx/pull/72, might be good to rebase?

Overall LGTM apart from some comments 👍

styx-common/src/main/java/com/spotify/styx/util/CachedSupplier.java
honnix: Something from Guava sounds more reliable. What do you think, @rouzwawi ? Did we have some problem then decided to have our own cached solution?

honnix:
You are right, I must have messed up with branches.

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
honnix: It looks at least Apollo doesn't touch ```User-Agent```, so use that instead of reinventing the wheel.

styx-common/src/main/java/com/spotify/styx/util/CachedSupplier.java
rouzwawi: This was just a product of my ignorance :)

I think it does exactly the same as the guava momoized supplier. Feel free to remove it.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -596,9 +599,19 @@ private Entity backfillToEntity(Backfill backfill) {
...
         .filter(e -> e.contains(PROPERTY_CONFIG_CONCURRENCY))
         .map(e -> e.getLong(PROPERTY_CONFIG_CONCURRENCY));
   }
+
+  List<String> clientBlacklist() {
+    final String blacklist = readConfigString(PROPERTY_CONFIG_CLIENT_BLACKLIST, "");
+    if (blacklist.isEmpty()) {
+      return ImmutableList.of();
+    } else {
+      return Arrays.stream(blacklist.split(";")).map(
```
rouzwawi: Datastore supports lists of values natively. That's probably better than manual splitting on `;`.

styx-api-service/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -37,6 +44,32 @@
  */
 public final class Middlewares {
 
+  private static class ClientValidator implements
+                               Middleware<AsyncHandler<? extends Response<?>>,
+                                   AsyncHandler<? extends Response<ByteString>>> {
+    private final Supplier<List<String>> clientBlacklist;
+
+    ClientValidator(Storage storage) {
+      clientBlacklist = new CachedSupplier<>(storage::clientBlacklist, Instant::now);
```
rouzwawi: I would rather pass in a `Supplier<List<String>>` as a dependency instead of constructing it from `Storage` here.

This should also use the injected `Time` instance instead of using `Instant::now`. But this should be solved by just moving the construction up to the main service class.

styx-api-service/src/main/java/com/spotify/styx/StyxApi.java
```diff
@@ -100,13 +102,16 @@ public void create(Environment environment) {
     final StyxConfigResource styxConfigResource = new StyxConfigResource(storage);
     final CliResource cliResource = new CliResource(schedulerServiceBaseUrl, storage);
 
+    final Middleware clientValidator = Middlewares.clientValidator(storage);
```
rouzwawi: Pass in `new CachedSupplier<>(storage::clientBlacklist, time)` instead. See my other comment.

rouzwawi: I guess there's one difference in that `CachedSupplier` supports injecting time while the Guava decorator uses wall clock time.

honnix: Let's keep it then? Feels like still worth it.

styx-api-service/src/main/java/com/spotify/styx/api/Middlewares.java
honnix: Good point. I like this better. Didn't really like the class anyway. We can get rid of it and use closure instead.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
honnix: Will fix.

styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
honnix: If you don't have strong opinion, I would prefer runtime configuration. ;) hehehe

styx-common/src/main/java/com/spotify/styx/util/CachedSupplier.java
honnix: We keep it as it offers a bit more than guava.

honnix:
@danielnorberg have time to take a look again?

honnix:
@rouzwawi can I get a thumb up?

styx-api-service/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -67,4 +73,21 @@ private Middlewares() {
       }
     });
   }
+
+  public static Middleware<AsyncHandler<? extends Response<?>>,
+      AsyncHandler<? extends Response<ByteString>>> clientValidator(
+      Supplier<Optional<List<String>>> supplier) {
+    return innerHandler -> requestContext -> {
```
danielnorberg: Add a metric for the number of rejected requests?

danielnorberg:
👍

styx-api-service/src/main/java/com/spotify/styx/api/Middlewares.java
honnix: We don't have any metric emitted from styx-api-service. This feels like a separated PR. I will add todo for now.

PR #74, opened 2017-03-06
upgrade bigtable-hbase-1.1

honnix:
this version includes netty 4.1.6.Final which fixed a memory leak. Similar issue has been reported here. https://github.com/grpc/grpc-java/issues/2358

Due to maven upper bound rule, a few transit dependencies have been bumped.


danielnorberg:
👍

PR #75, opened 2017-03-06
Move rate limiting to docker runner

danielnorberg:


styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
```diff
@@ -104,7 +104,7 @@ public boolean debugEnabled() throws IOException {
   }
 
   @Override
-  public Optional<Double> submissionRate() throws IOException {
+  public Optional<Double> submissionRateLimit() throws IOException {
```
ulzha: :star:

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -120,12 +121,14 @@
   public static final String STYX_MODE_DEVELOPMENT = "development";
 
   public static final int SCHEDULER_TICK_INTERVAL_SECONDS = 2;
+  public static final int SUBMISSION_RATE_LIMITER_UPDATE_TICK_INTERVAL_SECONDS = 5;
```
ulzha: We now have a number of inconsistent ad-hoc "get config value from storage" implementations. Cf. https://github.com/spotify/styx/pull/75/files#diff-a473be0c753046d2e7f56a00c3ed5b62L127

I wonder if it would be valuable to un-blaze that trail, and do "CONFIG_UPDATE_INTERVAL_SECONDS" or something, with consistent (such that doesn't drown in warning.log noise) error signaling etc.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -422,6 +449,15 @@ private static void startScheduler(Scheduler scheduler, ScheduledExecutorService
         TimeUnit.SECONDS);
   }
 
+  private static void startSubmissionRateLimiterUpdate(Storage storage, RateLimiter rateLimiter,
+      ScheduledExecutorService exec) {
+    exec.scheduleAtFixedRate(
+        guard(() -> updateSubmissionRateLimiter(storage, rateLimiter)),
+        SUBMISSION_RATE_LIMITER_UPDATE_TICK_INTERVAL_SECONDS,
```
ulzha: Is there any point in adding initialDelay?

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: I think it could definitely make sense to refactor how runtime configuration is handled. Not sure atm how that should look though. Perhaps a separate PR?

honnix: It also makes sense to do a bit caching, feels too many interactions with datastore and most of the time the value doesn't change. Could be a separated PR.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: Perhaps not. Was done that way just to be consistent with `startTriggerManager()` and `startScheduler()`. Might not make sense there either.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: @honnix yeah

honnix: Does it make sense to use this as you pointed out for another PR?

http://google.github.io/guava/releases/21.0/api/docs/com/google/common/base/Suppliers.html#memoizeWithExpiration-com.google.common.base.Supplier-long-java.util.concurrent.TimeUnit-

ulzha: It could look like the current approach generalized. Something like `startRuntimeConfigUpdate(Storage configStorage, RateLimiter rateLimiter, Map<String, Resource> resources, SomeOtherUpdatableThatWeCareAbout foo, ..., ScheduledExecutorService exec)` would be fine IMHO.

I just want to ensure that we leave a nice message to the next editor of this code, that conveys that we want to go in consistent direction, not in ad-hoc directions. I don't require refactor in this PR, just rename the unnecessarily specific names to point in that more general direction (makes names more digestible too).

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
ulzha: Delay in scheduling is arguably not a big deal, but I -1 astonishingly starting a service with default configuration when the user has configured something else.

Specifying 0 would not be perfect anyway in the current setup I guess (racy) but would get +1, perfect being the enemy of good...

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -94,7 +94,7 @@
   public static final String PROPERTY_ALL_TRIGGERED = "allTriggered";
   public static final String PROPERTY_HALTED = "halted";
   public static final String PROPERTY_DEBUG_ENABLED = "debug";
-  public static final String PROPERTY_SUBMISSION_RATE = "submissionRate";
+  public static final String PROPERTY_SUBMISSION_RATE = "submissionRateLimit";
```
ulzha: `PROPERTY_SUBMISSION_RATE_LIMIT`

danielnorberg: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: Agreed 👍

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: @ulzha Do you mean that you'd rather have `startSubmissionRateLimiterUpdate()` be named something like `startConfigUpdate()` ?

danielnorberg: done!

ulzha:
:+1:

PR #76, opened 2017-03-07
mark a success natural trigger

honnix:
To get statistics of natural triggers and possibably generate alerts
if we have been missing the value for some time.


danielnorberg:
👍

PR #78, opened 2017-03-07
fix trigger execution event loss

danielnorberg:
Write trigger execution event before updating next trigger.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -264,16 +264,22 @@ public class Scheduler {
         return;
       }
 
-      final int remainingCapacity = backfill.concurrency() - backfillStates.getOrDefault(backfill.id(), 0L).intValue();
+      final Workflow workflow = workflowOpt.get();
+
+      final int remainingCapacity =
+          backfill.concurrency() - backfillStates.getOrDefault(backfill.id(), 0L).intValue();
       if (remainingCapacity <= 0) {
         return;
       }
 
-      final Workflow workflow = workflowOpt.get();
-
       final List<Instant> partitionsRemaining =
           ParameterUtil.rangeOfInstants(backfill.nextTrigger(), backfill.end(),
                                         workflow.schedule().partitioning());
+
+      if (partitionsRemaining.isEmpty()) {
+        storeBackfill(backfill.builder().allTriggered(true).build());
+        return;
+      }
 
       final int partitionsToTrigger = Math.min(remainingCapacity, partitionsRemaining.size());
 
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -62,7 +62,9 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 import org.slf4j.Logger;
@@ -251,19 +253,19 @@ public class Scheduler {
             counting()));
 
     backfills.forEach(backfill -> {
-      final BackfillBuilder builder = backfill.builder();
 
       final Optional<Workflow> workflowOpt = workflowCache.workflow(backfill.workflowId());
 
       if (!workflowOpt.isPresent()) {
         LOG.warn("workflow not found for backfill, skipping rest of triggers: {}", backfill);
+        final BackfillBuilder builder = backfill.builder();
         builder.halted(true);
         storeBackfill(builder.build());
         return;
       }
 
-      final int needed = backfill.concurrency() - backfillStates.getOrDefault(backfill.id(), 0L).intValue();
-      if (needed <= 0) {
+      final int remainingCapacity = backfill.concurrency() - backfillStates.getOrDefault(backfill.id(), 0L).intValue();
+      if (remainingCapacity <= 0) {
         return;
       }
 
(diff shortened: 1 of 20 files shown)


danielnorberg:
@rouzwawi @honnix @ulzha PTAL

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -121,17 +122,18 @@ public void restore(RunState runState, long count) {
...
   @Override
-  public void receive(Event event) throws IsClosed {
+  public CompletionStage<Void> receive(Event event) throws IsClosed {
     ensureRunning();
 
     final InstanceState state = states.get(event.workflowInstance());
     if (state == null) {
       LOG.warn("Received event for unknown workflow instance: {}", event);
-      return;
+      return null;
```
honnix: This will eventually propagate to `TriggerManager.tick()` where we do `processed.toCompletableFuture`, or? `CompletableFuture.completedFuture(null)` instead?

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -101,7 +103,6 @@ public void initialize(RunState runState) throws IsClosed {
     final long counter;
     try {
       counter = storage.getLatestStoredCounter(workflowInstance).orElse(NO_EVENTS_PROCESSED);
-      storeActivation(workflowInstance, counter);
```
honnix: So we store activation later in `process` after storing event, is that correct? If yes, we seem missing one `counter` value. Does that matter?

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
```diff
@@ -217,26 +217,27 @@ public void shouldRemoveStateIfTerminal() throws Exception {
...
-    assertThat(storage.getLatestStoredCounter(INSTANCE), hasValue(2L));
-    assertThat(storage.getCounterFromActiveStates(INSTANCE), hasValue(2L));
-
-    stateManager.receive(Event.success(INSTANCE));
-
-    assertTrue(stateManager.awaitIdle(1000));
-    assertThat(storage.activeStatesMap.values(), is(empty()));
-  }
+  // TODO: rewrite/reformulate this test
+//  @Test
```
honnix: Maybe `@Ignore("...")` instead of commenting out?

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -97,10 +99,11 @@ void tick() {
 
       if (enabled.contains(workflow.id())) {
         try {
-          triggerListener.event(
+          CompletionStage<Void> processed = triggerListener.event(
               workflow,
               Trigger.natural(),
               decrementInstant(next, partitioning));
+          processed.toCompletableFuture().get(1, TimeUnit.MINUTES);
```
honnix: What might cause timeout, a big queue to process? Is it possible the event eventually gets processed and job gets to run, but we might trigger it one more time?

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
rouzwawi: Why was it disabled in the first place?

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
rouzwawi: Seems like this should actually return a failed future?

danielnorberg: indeed, good catch

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
danielnorberg: Yeah, perhaps we should wait forever?

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
danielnorberg: I want to rewrite this test as part of this PR. The reason it was disabled is that this test now fails due to the removal of the synchronous `storeActivation` call.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
danielnorberg: I'm not sure. @rouzwawi ?

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
honnix: Waiting forever is not necessarily bad. If it takes really long time to complete, we probably have a super long queue, and firing new events could make it worse. However this does change current behaviour. I would like to get some words from @rouzwawi

danielnorberg: Waiting forever should be equivalent to executing the event handler and resulting state transitions etc synchronously. I believe @rouzwawi agreed that this is what we want to do but let's verify =)

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
danielnorberg: @rouzwawi ping

rouzwawi: This would just have stored whatever was the last counter value (or -1) in datastore, which then was incremented further down in the actual state.

Since there's no read-increment-write sequence in `process()`, but rather a `currentCount = counter++`, this will behave the same (that is, it will write events starting at 0,1,2,... or prevTrig+1, prevTrig+2...).

danielnorberg:
As there has been some discussion in other channels to the effect that we might actually want all of the `StateManager.receive()` calls to be synchronous (blocking), it might make sense to make the `StateManager.receive()` actually be synchronous. But perhaps that can be part of future refactoring in a separate PR.

honnix:
This PR is getting huge.

danielnorberg:
@honnix yup, we've got to ship this before it has time to get any larger

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -251,49 +253,53 @@ private void triggerBackfills(Collection<InstanceState> activeStates) {
...
+      final int remainingCapacity =
+          backfill.concurrency() - backfillStates.getOrDefault(backfill.id(), 0L).intValue();
+
+      Instant partition = backfill.nextTrigger();
+
+      for (int i = 0; i < remainingCapacity && partition.isBefore(backfill.end()); i++) {
+        try {
+          triggerListener
+              .event(workflow, Trigger.backfill(backfill.id()), partition)
+              .toCompletableFuture().get();
```
honnix: Do we want to have similar comment like https://github.com/spotify/styx/pull/78/files#diff-8681f2f92e0c50ca9a99b3fc6ee3e4a2R105

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -251,49 +253,53 @@ private void triggerBackfills(Collection<InstanceState> activeStates) {
...
+          Thread.currentThread().interrupt();
+          throw new RuntimeException(e);
+        } catch (ExecutionException e) {
+          LOG.error("failed to trigger backfill for state [{}]: {}",
+              partition, backfill);
+          throw new RuntimeException(e);
+        }
+
+        partition = incrementInstant(partition, backfill.partitioning());
+        storeBackfill(backfill.builder()
```
honnix: So we decided to store for each partition, in case of losing something?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -251,49 +253,53 @@ private void triggerBackfills(Collection<InstanceState> activeStates) {
             counting()));
 
     backfills.forEach(backfill -> {
-      final BackfillBuilder builder = backfill.builder();
```
honnix: empty line

honnix: a general comment. this method is getting too long to follow. we will need a refactoring later. but now it is ok.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: Yeah, this covers the case where e.g. two partitions get backfill triggered but then the third partition fails. Then we avoid re-triggering the two successful ones again. (Not guaranteed exactly-once but still)

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: Yeah, we might want to refactor this further.

PR #79, opened 2017-03-08
log polled event

honnix:


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -302,28 +325,11 @@ public void eventReceived(Action action, Pod pod) {
...
-        return pod.getStatus().toString();
-      }
-    }
-
     private void reconnect() {
       LOG.warn("Re-establishing watching from {}", lastResourceVersion);
 
       try {
-        client.pods()
+        watch = client.pods()
```
honnix: it was not necessary a big problem, but definitely not correct.

danielnorberg: Close the previous watcher before replacing it.

honnix: cool

danielnorberg: Or is the watcher already closed when `reconnect()` is called?

honnix: it should have been, I will take a look whether `close()` hurts if we do it again. Maybe we can skip.

honnix: `watcher.close()` is not very useful. It sends `close` frame to server and print stack trace if `IOException` which it will definitely get. We can skip it.

danielnorberg:
👍

PR #84, opened 2017-03-13
use the correct termination log file

ulzha:
The default Kubernetes termination message path: https://kubernetes.io/docs/tasks/debug-application-cluster/determine-reason-pod-failure/#setting-the-termination-log-file


honnix:
👍

Issue #20, opened 2016-12-06
Retry delay calculation should not count missing dependency runs


rouzwawi:
The current exponential backoff calculation that is applied when an execution fails uses the `tries` counter for the delay calculation. This means that a job that has been missing dependencies for N executions fails on execution N+1, it will get a high retry delay even though it's the first actual execution that did something. See [TerminationHandler](https://github.com/spotify/styx/blob/master/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java#L98).

Fix the retry calculation to not consider missing dependency runs in the delay calculation.

rouzwawi:
@i-maravic

PR #85, opened 2017-03-13
count number of failure for backoff

honnix:
instead of using tries which is wrong, using failures to do backoff

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/state/RunState.java
+++ b/styx-common/src/main/java/com/spotify/styx/state/RunState.java
@@ -38,6 +38,7 @@ import com.spotify.styx.model.Event;
 import com.spotify.styx.model.EventVisitor;
 import com.spotify.styx.model.ExecutionDescription;
 import com.spotify.styx.model.WorkflowInstance;
+import com.spotify.styx.state.Message.MessageLevel;
 import com.spotify.styx.util.Time;
 import com.spotify.styx.util.TriggerUtil;
 import java.time.Instant;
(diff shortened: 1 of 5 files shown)


styx-common/src/main/java/com/spotify/styx/state/RunState.java
```diff
@@ -284,6 +286,15 @@ public RunState terminate(WorkflowInstance workflowInstance, int exitCode) {
       }
     }
 
+    int failures(StateData data, int exitCode) {
+      switch (exitCode) {
+        case SUCCESS_EXIT_CODE:
+        case MISSING_DEPS_EXIT_CODE:
+          return 0;
+        default:                     return data.failures() + 1;
```
danielnorberg: new line?

danielnorberg:
👍

honnix:
@fabriziodemaria @bergman Please review this as well. Not sure I'm catching the right logic.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
```diff
@@ -81,7 +81,7 @@ private void checkRetry(RunState state) {
       if (state.data().lastExit().map(c -> c == MISSING_DEPS_EXIT_CODE).orElse(false)) {
         delayMillis = Duration.ofMinutes(MISSING_DEPS_RETRY_DELAY_MINUTES).toMillis();
       } else {
-        delayMillis = retryUtil.calculateDelay(state.data().tries()).toMillis();
```
fabriziodemaria: `calculateDelay` method's argument name in `RetryUtil` should be changed to `failures` just to avoid confusion

styx-common/src/main/java/com/spotify/styx/state/RunState.java
```diff
@@ -263,11 +263,13 @@ public RunState terminate(WorkflowInstance workflowInstance, Optional<Integer> e
       switch (state()) {
         case RUNNING:
           final double cost = exitCost(exitCode);
+          final int failures = failures(data(), exitCode);
```
fabriziodemaria: This could be called something like `consecutiveFailures`

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
honnix: `consecutiveFailures` will be the name to use

fabriziodemaria:
👍

honnix:
This fixes #20

fabriziodemaria:
close #20

PR #86, opened 2017-03-13
indicate missing exit code by -1 instead of 127

danielnorberg:
Exit code 127 means "command not found" and can be confusing to see when the actual issue is e.g. a missing termination log message.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesPodEventTranslator.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesPodEventTranslator.java
@@ -46,6 +46,8 @@ import java.util.function.Predicate;
 
 public final class KubernetesPodEventTranslator {
 
+  private static final int NO_EXIT_CODE = -1;
+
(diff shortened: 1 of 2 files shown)


honnix:
👍

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesPodEventTranslator.java
```diff
@@ -82,13 +82,16 @@ private static int getExitCode(Pod pod, ContainerStatus status) {
...
         LOG.warn("Unexpected termination log message for container {}", status.getContainerID(), e);
-
-        // make sure to signal an error to be on the safe side, as opposed to the purported exit code
-        return 127;
+        return NO_EXIT_CODE;
       }
     }
 
-    return terminated.getExitCode();
+    if (terminated.getExitCode() == null) {
```
honnix: Started thinking about whether we need this branch. We are not checking existence of `status.getState().getTerminated()` or so. Maybe we should trust kube.

PR #88, opened 2017-03-13
Make exit code Optional instead of claiming 127 when exit code is missing

danielnorberg:
This is an alternative to https://github.com/spotify/styx/tree/dano/indicate-missing-exit-code


honnix:
👍

PR #90, opened 2017-03-14
detailed logging

honnix:


honnix:
@danielnorberg PTAL, we can take in this for next release as well

styx-scheduler-service/src/main/java/com/spotify/styx/StateInitializingTrigger.java
```diff
@@ -69,7 +70,9 @@ public void event(Workflow workflow, Trigger trigger, Instant instant) {
       stateManager.receive(
           Event.triggerExecution(workflowInstance, trigger));
     } catch (StateManager.IsClosed isClosed) {
-      LOG.warn("State receiver is closed", isClosed);
+      LOG.warn(format(
```
danielnorberg: Use logger message formatting instead of String format: 
```
LOG.warn("State receiver is closed when processing workflow {} for trigger {} at {}", workflow, trigger, instant, isClosed);
```

danielnorberg:
👍

PR #91, opened 2017-03-14
enhance exit code paranoia

danielnorberg:
Consider a non-zero container exit code with a zero termination log exit code to be a failure.


danielnorberg:
@honnix PTAL

honnix:
👍

PR #92, opened 2017-03-16
log workflow instance would be more useful

honnix:


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesPodEventTranslator.java
```diff
@@ -68,6 +68,8 @@ public TerminationLogMessage(
   }
 
   private static Optional<Integer> getExitCode(Pod pod, ContainerStatus status, Stats stats) {
+    final String workflowInstance = pod.getMetadata().getAnnotations()
+        .getOrDefault(KubernetesDockerRunner.STYX_WORKFLOW_INSTANCE_ANNOTATION, "N/A");
```
danielnorberg: Could also pass in the workflow id from the caller?

honnix: `pod` is used for other purpose in this method however. do we really want to add another param as signature?

honnix:
@danielnorberg how about this?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesPodEventTranslator.java
```diff
@@ -175,7 +180,10 @@ public TerminationLogMessage(
         exited = true;
         exitCode = pod.getStatus().getContainerStatuses().stream()
             .filter(IS_STYX_CONTAINER)
-            .map(cs -> getExitCode(pod, cs, stats))
+            .map(cs -> getExitCodeIfValid(
+                workflowInstance.toKey(),
+                pod.getMetadata().getAnnotations().get(DOCKER_TERMINATION_LOGGING_ANNOTATION),
```
danielnorberg: Could this stay inside `getExitCodeIfValid`?

honnix: So we keep `Pod` as the parameter? Sure. No strong opinion.

danielnorberg: ¯\\_(ツ)_/¯

Keeps the PR diff smaller

danielnorberg:
👍

PR #95, opened 2017-03-18
[WIP] v2 API

honnix:


styx-schedule-source/src/main/java/com/spotify/styx/schedule/model/ScheduleDefinition.java
```diff
@@ -24,27 +24,28 @@
...
   public static ScheduleDefinition create(
-      @JsonProperty("data_endpoints") @Nullable List<DataEndpoint> dataEndpoints) {
-    if (dataEndpoints == null) {
-      dataEndpoints = Collections.emptyList();
+      @JsonProperty("workflows") @Nullable List<Schedule> schedules) {
+    if (schedules == null) {
+      schedules = Collections.emptyList();
     }
-    return new AutoValue_ScheduleDefinition(dataEndpoints);
+    return new AutoValue_ScheduleDefinition(schedules);
```
ulzha: We require backwards compatibility, so this is what I stubbed locally:
```
  @JsonCreator
  public static ScheduleDefinition create(
      @JsonProperty("data_endpoints") @Nullable List<DataEndpoint> dataEndpoints,
      @JsonProperty("workflows") @Nullable List<DataEndpoint> workflows) {
    if (dataEndpoints == null) {
      dataEndpoints = Collections.emptyList();
    }
    if (workflows == null) {
      workflows = Collections.emptyList();
    }
    workflows.addAll(dataEndpoints);
    return new AutoValue_ScheduleDefinition(workflows);
  }
```
s/workflows/schedules/g

honnix: It would be better to have two models separated and do conversion, so we can easily get rid of the old one later. Like what the other deprecated models. I will give put some conversion methods to help.

ulzha: Ah, I see; I'll try to invoke the deprecated.ScheduleDefinition.

honnix: would be something like `new_model_instance = deprecated.ScheduleDefinition.create(<old_model_instance>)`, for OSS Styx, there is only new model be used.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -87,6 +86,7 @@ public BackfillResource(String schedulerServiceBaseUrl, Storage storage) {
     this.storage = Objects.requireNonNull(storage);
   }
 
+  // FIXME
```
bergman: remove?

styx-api-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -0,0 +1,80 @@
...
+
+    return Api.prefixRoutes(schedulerProxies, V2);
+  }
+
+  private CompletionStage<Response<ByteString>> proxyToScheduler(String path, RequestContext rc) {
+    return rc.requestScopedClient()
+        .send(rc.request().withUri(schedulerServiceBaseUrl + SCHEDULER_BASE_PATH + path));
+  }
+
+  private static String arg(RequestContext rc) {
```
bergman: inline

styx-api-service/src/main/java/com/spotify/styx/api/StyxConfigResource.java
```diff
@@ -61,7 +62,7 @@ public StyxConfigResource(Storage storage) {
             rc -> patchStyxConfig(rc.request()))
     );
 
-    return Api.prefixRoutes(v0, V0, V1);
+    return Api.prefixRoutes(v0, V0, V1, V2);
```
bergman: could we rename the `v0` variable to something like `v0Routes`

bergman: or just `routes`

PR #97, opened 2017-03-20
Fix backfill filtering

bergman:
Once again bitten by datastore `.filter()` not adding to a list but overwriting with each call.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -31,11 +31,14 @@ import com.google.cloud.datastore.KeyFactory;
 import com.google.cloud.datastore.PathElement;
 import com.google.cloud.datastore.Query;
 import com.google.cloud.datastore.QueryResults;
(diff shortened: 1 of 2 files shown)


honnix:
👍

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -551,13 +554,24 @@ void deleteResource(String id) throws IOException {
...
+    if (andedFilters.size() > 1) {
       queryBuilder
-          .filter(PropertyFilter.eq(PROPERTY_ALL_TRIGGERED, false))
-          .filter(PropertyFilter.eq(PROPERTY_HALTED, false));
+          .filter(
+              CompositeFilter.and(
+                  andedFilters.get(0),
+                  andedFilters.subList(1, andedFilters.size()).toArray(new Filter[] {})));
+    } else if (andedFilters.size() == 1) {
+      queryBuilder.filter(andedFilters.get(0));
```
honnix: How about using `CompositeFilter` anyway even there is only one filter?

PR #99, opened 2017-03-21
v2 API and new model

honnix:
New model wll be used to do scheduling on workflow instead of data
endpoints.

v0 and v1 APIs are deprecated now.

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
@@ -2,7 +2,7 @@
  * -\-\-
  * Spotify Styx API Service
  * --
- * Copyright (C) 2016 Spotify AB
+ * Copyright (C) 2017 Spotify AB
  * --
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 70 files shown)


bergman:
To be clear, this is a squashed version of #95

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -119,12 +118,12 @@ public BackfillResource(String schedulerServiceBaseUrl, Storage storage) {
...
     return cat(
-        Api.prefixRoutes(entityRoutes, V1),
-        Api.prefixRoutes(routes, V1)
+        Api.prefixRoutes(entityRoutes, V2),
+        Api.prefixRoutes(routes, V2)
     );
   }
 
-  private BackfillsPayload getBackfills(RequestContext rc) {
+  public BackfillsPayload getBackfills(RequestContext rc) {
```
ulzha: Why did all the private->public take place? Doesn't look like a consequence of endpoint->workflow change...

honnix: The deprecated BackfillResource has no logic but calling the new version, so that's why it is public now.

styx-api-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -0,0 +1,76 @@
...
+  private final String schedulerServiceBaseUrl;
+
+  public SchedulerResource(String schedulerServiceBaseUrl) {
+    this.schedulerServiceBaseUrl = Objects.requireNonNull(schedulerServiceBaseUrl);
+  }
+
+  public Stream<? extends Route<? extends AsyncHandler<? extends Response<ByteString>>>> routes() {
+    final List<Route<AsyncHandler<Response<ByteString>>>> schedulerProxies = Arrays.asList(
+        Route.async(
+            "GET", BASE + "/<endpoint:path>",
```
ulzha: So many mentions of "endpoint" without deprecatesauce...

honnix: they are not the "data endpoint", rather URL thing.

bergman:
👍

PR #100, opened 2017-03-21
partitioning changes to schedule

honnix:
@rouzwawi 

Question about open source schedule definition. Should we go and break backward compatibility?

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
@@ -250,7 +250,7 @@ public final class BackfillResource {
       if (!workflowOpt.isPresent()) {
         return Response.forStatus(Status.NOT_FOUND.withReasonPhrase("workflow not found"));
       }
-      partitioning = workflowOpt.get().schedule().partitioning();
+      partitioning = workflowOpt.get().schedule().schedule();
     } catch (Exception e) {
       throw Throwables.propagate(e);
     }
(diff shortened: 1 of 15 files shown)


styx-common/src/main/java/com/spotify/styx/model/Schedule.java
```diff
@@ -39,7 +39,7 @@
   public abstract String id();
 
   @JsonProperty
-  public abstract Partitioning partitioning();
+  public abstract Partitioning scheduleSpec();
```
rouzwawi: Hmm, I still feel like just calling it `schedule` makes most sense. Try reading the name in the context of [the docs here](https://github.com/spotify/styx/blob/2a05a2d7479a470cc057d2d738a6a8a581be7660/README.md#workflow-schedule-configuration) (from the cron PR). It would result in a configuration that looks like:

```yaml
schedules:
- id: foo
  schedule: @hourly
...
```

compared to

```yaml
schedules:
- id: foo
  schedule_spec: @hourly
...
```

honnix: I think so too

honnix: let's make user interface clean at this moment

styx-local-files-source/src/main/java/com/spotify/styx/YamlScheduleDefinition.java
```diff
@@ -34,6 +34,7 @@
 abstract class YamlScheduleDefinition {
 
   @JsonProperty
+  // TODO should we just go break backward compatibility?
```
rouzwawi: What does this comment refer to?

honnix:
That Schedule class which has schedule instead of partitioning. That will break backward compatibility. But I think we can do that to move forward.

rouzwawi:
Ok, cool. 👍

PR #108, opened 2017-03-24
clean up data endpoint

honnix:


doc/api.apib
```diff
@@ -46,21 +46,21 @@ aglio -i doc/api.apib -o api.html --theme-template triple --theme-variables stre
...
 
 + Response 200 (application/json)
 
         [{
             "workflow_instance": {
-              "workflow_id": {
+              "workflow": {
                 "component_id": "styx-canary",
-                "endpoint_id": "LuigiCanary"
+                "workflow": "LuigiCanary"
```
danielnorberg: Looks weird to me that the "workflow" object contains a "workflow" field. Should one of these be `workflow_id`?

honnix: I tried both, and was not sure which one is better. Now you said it, let's go for `workflow_id`!

styx-common/src/main/java/com/spotify/styx/model/WorkflowInstance.java
```diff
@@ -39,7 +39,7 @@
 
   /**
    * This property contains the thing needed to turn a general purpose Workflow into something
-   * that satisfies a specific schedule partition. Might be for example
+   * that satisfies a specific schedule. Might be for example
```
danielnorberg: ...satisfies a specific workflow instance?

danielnorberg:
👍

PR #109, opened 2017-03-24
api: trigger proxy: propagate status code

danielnorberg:


danielnorberg:
@honnix PTAL

styx-api-service/src/main/java/com/spotify/styx/api/deprecated/CliResource.java
```diff
@@ -114,21 +115,18 @@ public CliResource(StatusResource statusResource, String schedulerServiceBaseUrl
     } catch (JsonProcessingException e) {
       throw Throwables.propagate(e);
```
honnix: I'm thinking we should return 400 error code here instead of 500.

styx-api-service/src/main/java/com/spotify/styx/api/deprecated/CliResource.java
```diff
@@ -114,21 +115,18 @@ public CliResource(StatusResource statusResource, String schedulerServiceBaseUrl
...
+    final Request request = rc.request()
+        .withUri(schedulerServiceBaseUrl + SCHEDULER_BASE_PATH + TRIGGER_PATH)
+        .withPayload(payload);
+    return rc.requestScopedClient().send(request).thenApply(response ->
+        response.withPayload(response.payload().map(p -> {
+          try {
+            return WorkflowInstance.create(
+                Json.deserialize(p, com.spotify.styx.model.WorkflowInstance.class));
+          } catch (IOException e) {
+            throw new RuntimeException(e);
```
honnix: ` throw Throwables.propagate(e);` seems to be used everywhere and feels more reliable.

styx-api-service/src/main/java/com/spotify/styx/api/deprecated/CliResource.java
danielnorberg: 👍

styx-api-service/src/main/java/com/spotify/styx/api/deprecated/CliResource.java
danielnorberg: https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate

honnix:
👍

PR #113, opened 2017-03-27
k8s: keep reference to watcher

bergman:
@honnix i found this old branch, do you think it's still needed?


bergman:
Doesn't it solve the problem of not having the reference to the new watcher saved somewhere? The `this` is needed to point to the instance of `KubernetesDockerRunner`.

honnix:
That problem it does solve. But still we get watcher disconnected and retries occasionally. However it does recover after a while.

👍

danielnorberg:
👍

bergman:
Closing this since honnix already fixed it earlier.

PR #114, opened 2017-03-27
[PAUSED] modularize cli

bergman:


styx-cli/src/main/java/com/spotify/styx/cli/BackfillCli.java
```diff
@@ -0,0 +1,226 @@
...
+
+      if (alias != null && !alias.isEmpty()) {
+        subparser.aliases(alias);
+      }
+
+      return subparser;
+    }
+  }
+
+  static void runCommand(Main main, BackfillCommand backfillCommand)
```
danielnorberg: Can we eliminate the circular dependency between `Main` and `{Backfill, Resource}Cli`?

bergman: I was bothered by that too, maybe i can take another look at it.

honnix:
Do we still want to have this PR? CLI has become very thin after splitting client.

bergman:
I'll try to do another round on this, not sure if it still makes sense.

honnix:
This PR is now having quite bad conflicts. Shall we close and try again if you plan to?

PR #115, opened 2017-03-27
Update API blueprints

fabriziodemaria:


doc/api.apib
```diff
@@ -145,7 +148,9 @@ aglio -i doc/api.apib -o api.html --theme-template triple --theme-variables stre
 
         {
             "enabled": "true",
-            "docker_image": "busybox:latest"
+            "docker_image": "luigi-canary-dummy:dummy",
+            "commit_sha": "f043333085fa87738ac24f04d64fb58ecc845111",
+            "next_natural_trigger": null
```
fabriziodemaria: This should be fixed to return the actual value

doc/api.apib
```diff
@@ -174,11 +183,342 @@ aglio -i doc/api.apib -o api.html --theme-template triple --theme-variables stre
...
+            "commit_sha": "f043333085fa87738ac24f04d64fb58ecc845111",
+        }
+
++ Response 200 (application/json)
+
+        {
+            "enabled": null ,
+            "docker_image": "luigi-canary-dummy:dummy",
+            "commit_sha": "f043333085fa87738ac24f04d64fb58ecc845111",
+            "next_natural_trigger": null
```
fabriziodemaria: `next_natural_trigger` and `enabled` should not be part of the response, if we want to keep the current behaviour of PATCH state at Component level that doesn't allow to modify such parameters via the API

doc/api.apib
```diff
@@ -174,11 +183,342 @@ aglio -i doc/api.apib -o api.html --theme-template triple --theme-variables stre
 + Request (application/json)
 
         {
-            "docker_image": "busybox:latest"
+            "docker_image": "luigi-canary-dummy:dummy",
+            "commit_sha": "f043333085fa87738ac24f04d64fb58ecc845111",
+        }
```
fabriziodemaria: Is there a reason not to allow `enabled` PATCH at component level? Also, `next_natural_trigger` is silently ignored according to the current implementation. We can either support it or explicitly return an error to the user

doc/api.apib
```diff
@@ -20,7 +20,7 @@ aglio -i doc/api.apib -o api.html --theme-template triple --theme-variables stre
 
 # Group Workflows
 
-## Workflow [/v1/workflows/{component}/{workflow_id}]
```
ulzha: Unsure if removing v1 altogether is correct, given that it's not sunset yet. Should it be kept in a separate file?

ulzha: (Just playing a devil's advocate. I'd like outdated stuff to disappear immediately.)

doc/api.apib
ulzha: Unsure if the component level actions were ever required by anyone? I think if usages are extremely rare then I'd prefer to not maintain codebase for setting state at component level at all.

doc/api.apib
```diff
@@ -174,11 +183,342 @@ aglio -i doc/api.apib -o api.html --theme-template triple --theme-variables stre
...
+        {
+            "enabled": null ,
+            "docker_image": "luigi-canary-dummy:dummy",
+            "commit_sha": "f043333085fa87738ac24f04d64fb58ecc845111",
+            "next_natural_trigger": null
+        }
+
+# Group Resources
+
+## Resources [/v2/resources]
```
ulzha: Lacks documentation somewhere. It is non-obvious how the ids can be invented, and what concurrency means here. It's also too abstractly documented (no example) [in Confluence](https://confluence.spotify.net/display/DEX/Styx#Styx-yaml).

doc/api.apib
```diff
@@ -174,11 +183,342 @@ aglio -i doc/api.apib -o api.html --theme-template triple --theme-variables stre
...
+                    "id": "backfill-1489054446085-52684",
+                    "next_trigger": "2017-01-01T15:00:00Z",
+                    "schedule": "hours",
+                    "start": "2017-01-01T00:00:00Z",
+                    "workflow_id": {
+                        "component_id": "styx-canary",
+                        "id": "LuigiCanary"
+                    }
+                },
+                "statuses": null
```
ulzha: Not documented what the statuses will look like.

doc/api.apib
honnix: #116

doc/api.apib
honnix: #117

doc/api.apib
honnix: I think it's probably OK for now to have v2 only, and an issue to track this. #118

doc/api.apib
honnix: it is good enough to copy https://github.com/spotify/styx/pull/115/files/d2d0d6f8edaac0001ae91eff69f8dd2759685881#diff-cfe769770442367be801c65d5428a63cR380 here

doc/api.apib
honnix: A PR addressing #116 has been created. I think we should put the desired response here, that means `next_natural_trigger` will have a value instead of `null`.

doc/api.apib
honnix: I'm fine with current state that we don't allow patching `enabled` and `next_natural_trigger` on component level. The API blueprint here looks good.

doc/api.apib
honnix: A PR addressing #117 has been created. So here we can remove `enabled` and `next_natural_trigger` from response.

honnix:
Overall LGTM.

honnix:
👍

PR #119, opened 2017-03-28
Update state graph and related docs

fabriziodemaria:


honnix:
About those two hanging events: `timeout` and `halt`, does it mean we can transit from any state to `FAILED` or `ERROR` upon receiving them?

honnix:
LGTM 👍

PR #121, opened 2017-03-28
Set `STYX_TRIGGER_ID`

kanterov:
Need this feature a lot!

I'm not sure this is an optimal approach, but seems to work. Of course, I didn't test it anyhow except unit tests, because I have no idea how to do it.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
@@ -33,12 +33,14 @@ import com.spotify.styx.state.RunState;
 import com.spotify.styx.state.StateManager;
 import com.spotify.styx.storage.Storage;
 import com.spotify.styx.util.ResourceNotFoundException;
+import com.spotify.styx.util.TriggerUtil;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ExecutorService;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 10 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
```diff
@@ -141,7 +141,8 @@ private String dockerRunnerStart(RunState state) throws IOException {
         dockerImage,
         ImmutableList.copyOf(command),
         executionDescription.dockerTerminationLogging(),
-        executionDescription.secret());
+        executionDescription.secret(),
+        state.data().triggerId());
```
honnix: `triggerId` is probably deprecated as noted by comment. `TriggerUtil.triggerId(trigger)` feels better. @rouzwawi ?

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
```diff
@@ -338,7 +343,7 @@ public void runsDockerImageWithArgsTemplate() throws Exception {
     WorkflowInstance workflowInstance = dockerRuns.get(0)._1;
     RunSpec runSpec = dockerRuns.get(0)._2;
     assertThat(workflowInstance.workflowId(), is(HOURLY_WORKFLOW.id()));
-    assertThat(runSpec, is(RunSpec.simple("busybox", "--hour", "2016-03-14T15")));
+    assertThat(runSpec, is(naturalRunSpec("busybox", ImmutableList.of("--hour", "2016-03-14T15"), "natural-trigger")));
```
honnix: `TriggerUtil.NATURAL_TRIGGER_ID`

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
rouzwawi: Yes

rouzwawi:
LGTM overall. Just change to use `StadeData.trigger()` with `TeiggerUtil`. I believe that could also make the trigger id mandatory in `RunSpec`. At least that's how it should be since we should not have any executions without a trigger.

kanterov:
@rouzwawi @honnix thanks for review, done.

@rouzwawi `triggerId` is still optional because `trigger` in `runState` is optional.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
```diff
@@ -39,6 +39,8 @@
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ExecutorService;
+
+import com.spotify.styx.util.TriggerUtil;
```
honnix: organize import? it shall be before `java.*` somewhere.

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
```diff
@@ -51,6 +52,8 @@
 import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.Optional;
+
+import com.spotify.styx.util.TriggerUtil;
```
honnix: same here.

honnix:
Minor issue. Otherwise LGTM. 👍

PR #122, opened 2017-03-29
return next_natural_triggerls

### LINKED ISSUE #116: missing next_natural_trigger value in response
`curl http://<host>/api/v2/workflows/<cid>/<wfid>/state`

`next_natural_trigger` in response is always `null`

honnix:
fix #116

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -224,15 +224,7 @@ class DatastoreStorage {
     final Key workflowKey = workflowKey(workflowId);
     return getOpt(datastore, workflowKey)
         .filter(e -> e.contains(PROPERTY_WORKFLOW_JSON))
-        .map(e -> {
-          try {
-            return OBJECT_MAPPER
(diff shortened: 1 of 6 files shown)


fabriziodemaria:
We should update tests in `WorkflowResourceTest` as well

rouzwawi:
This should return both the trigger and offset triggers.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -421,6 +421,7 @@ public WorkflowState workflowState(WorkflowId workflowId) throws IOException {
     WorkflowState.Builder builder = WorkflowState.builder().enabled(enabled(workflowId));
     getDockerImage(workflowId).ifPresent(builder::dockerImage);
     getCommitSha(workflowId).ifPresent(builder::commitSha);
+    getNextNaturalTrigger(workflowId).ifPresent(builder::nextNaturalTrigger);
```
rouzwawi: Can we refactor this method to not fetch the workflow once for each property?

honnix:
@rouzwawi PTAL?

rouzwawi:
LGTM!

PR #123, opened 2017-03-29
do not return Optional.empty() as null

### LINKED ISSUE #117: enabled and next_natural_trigger values being null
When patching at component level, `enabled` and `next_natural_trigger` in response should not be `null`.

honnix:
fix #117


honnix:
This is a more general problem maybe. For many cases we return `null` in json, should we not include it at all? Semantically they are very similar, but visually feels strange.

rouzwawi:
I think it's more informative to return `null` than omitting the field.

Also, the `WorkflowState` type has a dual role. It's used both for parsing PATCH requests (where everything is optional) and for returning the GET payload.

honnix:
I don't have strong opinions as long as it is consistent throughout the system.

honnix:
close #117

PR #125, opened 2017-03-29
make trigger name as container env variable

honnix:
@kanterov @rouzwawi 

Does it make sense to have trigger name as well?

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -40,6 +40,7 @@ import com.spotify.styx.state.Message;
 import com.spotify.styx.state.RunState;
 import com.spotify.styx.state.StateManager;
 import com.spotify.styx.state.Trigger;
+import com.spotify.styx.util.TriggerUtil;
 import io.fabric8.kubernetes.api.model.EnvVar;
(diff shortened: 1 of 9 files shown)


kanterov:
Makes a lot of sense for me, do I understand correctly that it's one of `adhoc`, `backfill`, `unknown`, `natural-trigger`?

honnix:
Exactly!

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -158,7 +160,11 @@ static Pod createPod(WorkflowInstance workflowInstance, RunSpec runSpec) {
         .build());
     env.add(new EnvVarBuilder()
         .withName(TRIGGER_ID)
-        .withValue(runSpec.triggerId().orElse(null))
+        .withValue(runSpec.trigger().map(TriggerUtil::triggerId).orElse(null))
+        .build());
+    env.add(new EnvVarBuilder()
+        .withName(TRIGGER_NAME)
```
rouzwawi: I know that it's called `TriggerUtil.name`, but I think using the term `name` is misleading. It's actually the trigger  type, e.g. `natural`, `ad-hoc`, `backfill`. `name` sounds like something that the user might have set.

I would use `STYX_TRIGGER_TYPE` on this user facing part and refactor the utilities to also say type instead of name.

rouzwawi:
👍

PR #127, opened 2017-03-31
being able to print plain string message as cli output

honnix:


styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -415,6 +415,9 @@ private void triggerWorkflowInstance()
     final WorkflowInstance workflowInstance = getWorkflowInstance(namespace);
     final ByteString payload = serialize(workflowInstance);
     send(Request.forUri(apiUrl("scheduler", "trigger"), "POST").withPayload(payload));
+    cliOutput.printMessage(CliOutput.CliMessage.create(
```
danielnorberg: Would it make sense to include the workflow id and parameter in the message?

Perhaps the json output should be more structured?

honnix: Thought about this actually. How to make it generic enough for other cases. How about:

```
{
   "type":"info/warning/error",
   "message":"...",
   "details":"..."
}
```

In this case, `details` may contain workflow information.

danielnorberg: How about outputting the WorkflowInstance returned by the POST?

danielnorberg: I.e.

```
{
  "workflow": {
    "component_id": "...", 
    "workflow_id"
  },
  "parameter": "..."
}
```

danielnorberg:
👍

PR #128, opened 2017-03-31
more meaningful units than Hz

honnix:


styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -60,43 +59,43 @@
...
 
   private static final MetricId TERMINATION_LOG_INVALID = BASE
       .tagged("what", "termination-log-invalid")
-      .tagged("unit", UNIT_FREQUENCY);
+      .tagged("unit", "count");
 
   private static final MetricId EXIT_CODE_MISMATCH = BASE
       .tagged("what", "exit-code-mismatch")
-      .tagged("unit", UNIT_FREQUENCY);
+      .tagged("unit", "count");
```
danielnorberg: As this is used for a meter, `count` would be emitted as `count/s` which doesn't seem to make much sense. How about e.g. `mismatch` instead? We could also omit the unit and fall back to the default of `n` which is emitted as `n/s`.

Similar for the rest of the units.

https://github.com/spotify/semantic-metrics#meter

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -60,43 +59,43 @@
...
 
   private static final MetricId TERMINATION_LOG_MISSING = BASE
       .tagged("what", "termination-log-missing")
-      .tagged("unit", UNIT_FREQUENCY);
+      .tagged("unit", "count");
 
   private static final MetricId TERMINATION_LOG_INVALID = BASE
       .tagged("what", "termination-log-invalid")
-      .tagged("unit", UNIT_FREQUENCY);
+      .tagged("unit", "count");
```
danielnorberg: Omit unit and use default?

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -60,43 +59,43 @@
...
 
   private static final MetricId NATURAL_TRIGGER_RATE = BASE
       .tagged("what", "natural-trigger-rate")
-      .tagged("unit", UNIT_FREQUENCY);
+      .tagged("unit", "triggering");
 
   private static final MetricId TERMINATION_LOG_MISSING = BASE
       .tagged("what", "termination-log-missing")
-      .tagged("unit", UNIT_FREQUENCY);
+      .tagged("unit", "count");
```
danielnorberg: Omit unit and use default?

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -60,43 +59,43 @@
...
 
   private static final MetricId PULL_IMAGE_ERROR_RATE = BASE
       .tagged("what", "pull-image-error-rate")
-      .tagged("unit", UNIT_FREQUENCY);
+      .tagged("unit", "error");
 
   private static final MetricId NATURAL_TRIGGER_RATE = BASE
       .tagged("what", "natural-trigger-rate")
-      .tagged("unit", UNIT_FREQUENCY);
+      .tagged("unit", "triggering");
```
danielnorberg: `s/triggering/trigger` ?

`triggering/s` vs `trigger/s`

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -60,43 +59,43 @@
...
 
   private static final MetricId EXIT_CODE_MISMATCH = BASE
       .tagged("what", "exit-code-mismatch")
-      .tagged("unit", UNIT_FREQUENCY);
+      .tagged("unit", "count");
 
   private static final MetricId SUBMISSION_RATE_LIMIT = BASE
       .tagged("what", "submission-rate-limit")
-      .tagged("unit", UNIT_FREQUENCY);
+      .tagged("unit", "submission");
```
danielnorberg: `submission` -> `submission/s` ?

https://github.com/spotify/semantic-metrics#gauge

danielnorberg:
There's some trickiness wrt units of meters and gauges etc to consider here.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
honnix: Using default looks better.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
honnix: Yes, sure. Didn't realize it's not a meter.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
honnix: Will fix.

danielnorberg:
👍  nice!

PR #130, opened 2017-04-03
styx cli: workflow show

danielnorberg:
Add a `workflow show <component> <workflow>` command.

```
$ styx workflow show dano-test-pipeline dano-test
Component: dano-test-pipeline
 Workflow: dano-test
  Enabled: false
 Schedule: DAYS
   Commit: 79b49cdd161d3f17b43f005be22ead4095ca35c8
    Image: registry.spotify.net/spotify/dano-test-pipeline:1490015757-d811bbe@sha256:c666cbfb44ac1debc3bb709a0268d19e3cd4e45a4d204dcbb4c3455b70c7dff7
     Args: [bash, -c, echo hello digest world {}]
```

The output of the command is a `Workflow` with some fields from `WorkflowState` added / merged. Notably the `WorkflowState` takes precedence for the `Image` field value.

PR #132, opened 2017-04-05
restore 404 handling, do not print stacktrace

danielnorberg:
Factored out status code checking into a separate reusable method called `requireSuccess`. 

This could perhaps do with some more refactoring. It seems a little odd that the `send()` method interprets a 404 as an error.

### DIFF
--- a/styx-cli/src/main/java/com/spotify/styx/cli/Main.java
+++ b/styx-cli/src/main/java/com/spotify/styx/cli/Main.java
@@ -31,6 +31,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.collect.ImmutableList;
 import com.spotify.apollo.Client;
 import com.spotify.apollo.Request;
+import com.spotify.apollo.Response;
 import com.spotify.apollo.core.Service;
 import com.spotify.apollo.core.Services;
 import com.spotify.apollo.environment.ApolloEnvironmentModule;
@@ -49,7 +50,6 @@ import com.spotify.styx.model.WorkflowId;
 import com.spotify.styx.model.WorkflowInstance;
 import com.spotify.styx.model.WorkflowState;
 import com.spotify.styx.util.EventUtil;
-import com.spotify.styx.util.FutureUtil;
 import com.spotify.styx.util.ParameterUtil;
 import java.io.IOException;
 import java.net.URLEncoder;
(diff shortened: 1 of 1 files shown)


danielnorberg:
@honnix PTAL

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -379,16 +379,21 @@ private void workflowShow() throws ExecutionException, InterruptedException, IOE
...
     CompletableFuture.allOf(workflowResponse, stateResponse).get();
 
-    final Workflow workflow = deserialize(workflowResponse.get(), Workflow.class);
-    final WorkflowState workflowState = deserialize(stateResponse.get(), WorkflowState.class);
+    final ByteString workflowPayload = requireSuccess(workflowResponse.get())
+        .payload().orElse(ByteString.EMPTY);
+    final ByteString statePayload = requireSuccess(stateResponse.get())
+        .payload().orElse(ByteString.EMPTY);
+
+    final Workflow workflow = deserialize(workflowPayload, Workflow.class);
```
honnix: Do we get `null` if `workflowPayload` is `ByteString.EMPTY`?

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -484,22 +489,26 @@ private String apiUrl(CharSequence... parts) {
   }
 
   private ByteString send(Request request) throws ExecutionException, InterruptedException {
-    return sendAsync(request).toCompletableFuture().get();
+    // TODO: Move requireSuccess use into callers of send()
```
honnix: A following PR to do for `TODO` I guess?

danielnorberg: The `send` method already threw `RuntimeException` on 404 so this just restores that behavior. I think that might not belong in `send` though hence the TODO.

https://github.com/spotify/styx/pull/130/files#diff-0a3c00857227c1cef7cc1e40e3d912acL454

honnix: Yep, that was my point. I think that might not belong in `send`. Just asking your plan when to move it out.

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
honnix: Although we shouldn't get it anyway...

danielnorberg: I guess the `AutoValue_Workflow` will throw NPE. That's true for other methods like `resourceShow` as well. The cli doesn't really seem to validate anything it gets from the styx api.

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
honnix: cool.

honnix:
👍

PR #133, opened 2017-04-10
fix k8s master connection leaks: bump k8s client

danielnorberg:
Also bump & add other dependencies as needed.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -82,7 +82,6 @@ import com.spotify.styx.workflow.WorkflowInitializer;
 import com.typesafe.config.Config;
 import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
-import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.NamespacedKubernetesClient;
 import java.io.Closeable;
 import java.io.IOException;
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -83,6 +83,7 @@ import com.typesafe.config.Config;
 import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.NamespacedKubernetesClient;
 import java.io.Closeable;
 import java.io.IOException;
 import java.security.GeneralSecurityException;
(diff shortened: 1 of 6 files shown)


danielnorberg:
@honnix PTAL

honnix:
Looks good on my phone. ;) Will read more later. 

Any related issue worth mentioning here? How did you find the leak?

honnix:
👍

danielnorberg:
Found the leaks in the warning logs of our production deployment of the styx scheduler.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -590,12 +591,12 @@ private static DockerRunner createDockerRunner(
       LOG.info("Creating LocalDockerRunner");
       return closer.register(DockerRunner.local(scheduler, stateManager));
     } else {
-      final KubernetesClient kubernetes = closer.register(getKubernetesClient(config, id));
+      final NamespacedKubernetesClient kubernetes = closer.register(getKubernetesClient(config, id));
```
honnix: Does it have to be `NamespacedKubernetesClient`?

danielnorberg: Yes

danielnorberg: The `inNamespace` method is only available on the `NamspacedKubernetesClient`. 

https://github.com/spotify/styx/blob/master/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java#L104

honnix: OK. Cool.

honnix:
LGTM. For `2.x` upgrade. 👍

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
bergman: Shouldn't the import for the non-namespaced client be removed then?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -190,8 +192,12 @@ static Pod createPod(WorkflowInstance workflowInstance, RunSpec runSpec) {
           .withSecretName(secret.name())
           .endSecret()
           .endVolume();
-      container =
-          container.addToVolumeMounts(new VolumeMount(secret.mountPath(), secret.name(), true));
+      VolumeMount volumeMount = new VolumeMountBuilder()
```
bergman: Name this variable something sensible rather than just the type of it.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java

danielnorberg:
@bergman Haven't tested against an actual k8s cluster yet. Perhaps we should have some e2e test with k8s (e.g. minikube) as part of the circle-ci build.

honnix:
If we merge and release, it should be tested against (at least installed in) the staging env.

danielnorberg:
@bergman I have now tested against a real GKE (and datastore/bigtable) and it seems to work, I'm able to trigger workflows and successfully run pods etc.

danielnorberg:
@honnix @bergman can I merge this?

honnix:
Merge. I think So.

honnix:
👍 to be clear

PR #135, opened 2017-04-18
Add exit code metrics

fabriziodemaria:


styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -106,7 +112,8 @@
   private final ConcurrentMap<String, Meter> storageOperationMeters;
   private final ConcurrentMap<String, Histogram> dockerOperationHistograms;
   private final ConcurrentMap<String, Meter> dockerOperationMeters;
-  private final ConcurrentHashMap<WorkflowId, Gauge> activeStatesPerWorkflowGauges;
+  private final ConcurrentMap<WorkflowId, Gauge> activeStatesPerWorkflowGauges;
+  private final ConcurrentMap<Tuple2<WorkflowId, Integer>, Meter> exitCodePerWorkflowMeters;
```
honnix: As we talked, I actually prefer the nested map over tuple.

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/MonitoringHandlerTest.java
```diff
@@ -71,6 +73,21 @@ public void shouldUpdateOnDefaultCase() throws Exception {
   }
 
   @Test
+  public void shouldMarkExitCode() throws Exception {
```
honnix: Another test case to cover the other branch? If there is no exit code.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -53,6 +55,10 @@
       .tagged("what", "workflow-count")
       .tagged("unit", "workflow");
 
+  private static final MetricId EXIT_CODE_RATE = BASE
+      .tagged("what", "exit-code-rate")
+      .tagged("unit", "code");
```
honnix: `code` as unit feels strange, we can leave it as default which will be `n`.

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/MonitoringHandlerTest.java
fabriziodemaria: Added

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
fabriziodemaria: We talked about it IRL and I think we agreed on keeping the tuple keys? @honnix

honnix: yup

honnix:
👍

PR #136, opened 2017-04-18
RFC: Use default retry mechanisms for Google Cloud APIs

bergman:
The Datastore operations should be good but the Bigtable operations are very unclear to me how to configure.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
@@ -25,6 +25,7 @@ import static com.spotify.styx.serialization.Json.serialize;
 
 import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
+import com.google.common.collect.Iterators;
 import com.google.common.collect.Sets;
 import com.spotify.styx.model.Event;
 import com.spotify.styx.model.SequenceEvent;
@@ -37,7 +38,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
-import java.util.TreeSet;
 import java.util.stream.Collectors;
 import okio.ByteString;
 import org.apache.hadoop.hbase.TableName;
@@ -71,7 +71,7 @@ public class BigtableStorage {
       final Scan scan = new Scan()
           .setRowPrefixFilter(Bytes.toBytes(workflowInstance.toKey() + '#'));
 
-      final SortedSet<SequenceEvent> set = newSortedEventSet();
+      final SortedSet<SequenceEvent> set = Sets.newTreeSet(SequenceEvent.COUNTER_COMPARATOR);
 
       for (Result result : eventsTable.getScanner(scan)) {
         set.add(parseEventResult(result));
@@ -138,8 +138,7 @@ public class BigtableStorage {
   Optional<Long> getLatestStoredCounter(WorkflowInstance workflowInstance)
       throws IOException {
     final Set<SequenceEvent> storedEvents = readEvents(workflowInstance);
-    final Optional<SequenceEvent> lastStoredEvent = storedEvents.stream().reduce((a, b) -> b);
-    return lastStoredEvent.map(SequenceEvent::counter);
+    return Optional.ofNullable(Iterators.getLast(storedEvents.iterator(), null)).map(SequenceEvent::counter);
   }
 
   WorkflowInstanceExecutionData executionData(WorkflowInstance workflowInstance) throws IOException {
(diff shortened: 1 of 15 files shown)


styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
```diff
@@ -177,31 +159,6 @@ private SequenceEvent parseEventResult(Result r) throws IOException {
...
 
-  private void storeWithRetries(RunnableWithException<IOException> storingOperation) throws IOException {
-    int storeRetries = 0;
-    boolean succeeded = false;
-
-    while (storeRetries < MAX_BIGTABLE_RETRIES && !succeeded) {
-      try {
-        storingOperation.run();
-        succeeded = true;
-      } catch (ResourceNotFoundException e) {
```
bergman: This doesn't throw this anymore so callers have to see for themselves if `e.getCause()` is a `ResourceNotFoundException`. Kind of clunky.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -450,36 +434,6 @@ public WorkflowState workflowState(WorkflowId workflowId) throws IOException {
...
-    int storeRetries = 0;
-
-    while (storeRetries < MAX_RETRIES) {
-      try {
-        return storingOperation.apply();
-      } catch (ResourceNotFoundException e) {
-        throw e;
-      } catch (DatastoreException | IOException e) {
-        if (e.getCause() instanceof ResourceNotFoundException) {
-          throw (ResourceNotFoundException) e.getCause();
```
bergman: As with the Bigtable case, this check would have to live in the caller. Unless we stop using `runInTransaction` but I guess there is a reason we use that?

honnix: I think we have more than one interactions for those cases, so wrapping inside a transaction makes sense.

styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
```diff
@@ -93,18 +81,16 @@
...
-        final String keyString = String.format("%s#%08d", workflowInstanceKey, sequenceEvent.counter());
-        final byte[] key = Bytes.toBytes(keyString);
-        final Put put = new Put(key, sequenceEvent.timestamp());
-
-        final byte[] eventBytes = serialize(sequenceEvent.event()).toByteArray();
-        put.addColumn(EVENT_CF, EVENT_QUALIFIER, eventBytes);
-        eventsTable.put(put);
-      }
-    });
+    final String workflowInstanceKey = sequenceEvent.event().workflowInstance().toKey();
```
honnix: Does it mean we won't retry for Bigtable?

styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
```diff
@@ -153,11 +139,7 @@ void writeEvent(SequenceEvent sequenceEvent) throws IOException {
       throws IOException {
     final Set<SequenceEvent> storedEvents = readEvents(workflowInstance);
     final Optional<SequenceEvent> lastStoredEvent = storedEvents.stream().reduce((a, b) -> b);
```
honnix: This way of getting the last element feels quite brutal. ;)

Maybe something like `return Optional.ofNullable(Iterators.getLast(storedEvents.iterator(), null)).map(SequenceEvent::counter);`?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
honnix: Hope `ResourceNotFoundException` will never be the cause of `DatastoreException` which is unfortunately a `RuntimeException`.

styx-common/src/main/java/com/spotify/styx/util/Connections.java
```diff
@@ -54,6 +56,7 @@ public static Datastore createDatastore(Config config) {
     return DatastoreOptions.builder()
         .namespace(namespace)
         .projectId(projectId)
+        .retryParams(RetryParams.defaultInstance())
```
honnix: This seems to be the default behaviour which means by default it retries. Check `ServiceOptions` L217. We have been doing something on top of the default retry.

bergman: You're right!

styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
bergman: 👍

honnix:
I think we agreed to park this PR for now and upgrade datastore client lib first and see whether it solves retry problem. Am I right?

honnix:
Shall we close this? Application layer retry seems still appreciated.

PR #137, opened 2017-04-20
eagerly restore container execution state

danielnorberg:
Before starting the scheduler, poll for execution container state in order to record the completion of any containers that succeeded while styx was offline.

This avoids styx unnecessarily failing successfully completed executions due to having artifically "timed out" when starting up after downtime.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,6 +86,11 @@
         <version>19.0</version>
       </dependency>
       <dependency>
+        <groupId>com.github.rholder</groupId>
+        <artifactId>guava-retrying</artifactId>
+        <version>2.0.0</version>
+      </dependency>
+      <dependency>
         <groupId>com.google.auto.value</groupId>
         <artifactId>auto-value</artifactId>
         <version>1.3</version>
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -26,6 +26,11 @@ import static com.spotify.styx.state.RunState.State.RUNNING;
 import static java.util.stream.Collectors.toSet;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.github.rholder.retry.RetryException;
+import com.github.rholder.retry.Retryer;
+import com.github.rholder.retry.RetryerBuilder;
+import com.github.rholder.retry.StopStrategies;
+import com.github.rholder.retry.WaitStrategies;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Lists;
@@ -62,6 +67,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -215,6 +221,21 @@ class KubernetesDockerRunner implements DockerRunner {
     }
   }
 
+  @Override
+  public void restore() {
+    final Retryer<Object> retryer = RetryerBuilder.newBuilder()
+        .retryIfException()
+        .withWaitStrategy(WaitStrategies.exponentialWait())
+        .withStopStrategy(StopStrategies.stopAfterAttempt(3))
+        .build();
+
+    try {
+      retryer.call(Executors.callable(this::tryPollPods));
+    } catch (ExecutionException | RetryException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
   public void init() {
     EXECUTOR.scheduleWithFixedDelay(
         this::pollPods,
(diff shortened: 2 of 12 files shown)


danielnorberg:
Would be good to get in some tests of this new behavior in this PR but I'm opening it now to validate the general approach.

danielnorberg:
@fabriziodemaria PTAL

pom.xml
```diff
@@ -86,6 +86,11 @@
         <version>19.0</version>
       </dependency>
       <dependency>
+        <groupId>com.github.rholder</groupId>
+        <artifactId>guava-retrying</artifactId>
```
honnix: Good to know this lib. We can replace own simple retry code (storage related) with this.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -210,6 +216,21 @@ public void close() throws IOException {
...
+    final Retryer<Object> retryer = RetryerBuilder.newBuilder()
+        .retryIfException()
+        .withWaitStrategy(WaitStrategies.exponentialWait())
+        .withStopStrategy(StopStrategies.stopAfterAttempt(3))
+        .build();
+
+    try {
+      retryer.call(Executors.callable(this::tryPollPods));
+    } catch (ExecutionException | RetryException e) {
+      throw new RuntimeException(e);
```
honnix: This is stop Styx starting process. Is it intended?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -246,20 +267,24 @@ void examineRunningWFISandAssociatedPods(PodList podList) {
...
-        logEvent(Watcher.Action.MODIFIED, pod, resourceVersion, true);
-        inspectPod(Watcher.Action.MODIFIED, pod);
-      }
+      tryPollPods();
     } catch (Throwable t) {
       LOG.warn("Error while polling pods", t);
     }
   }
 
+  private synchronized void tryPollPods() {
```
honnix: Does it have to be `synchronized`? `DockerRunner` is created in a very lazy way.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: Yes, if styx is unable to contact k8s at startup I believe it should crash and be restarted.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: Yes, it needs to either be `synchronized` or `restore` needs to enqueue a call to `tryPollPods` on the `EXECUTOR` and introduce a subtle (and easily violated) requirement that the executor is single threaded. I don't believe `inspectPod` and `tryPollPods` are thread safe.

danielnorberg: As for the lazy creation of `DockerRunner`, this PR changes that behavior to eagerly instantiate the docker runner in `StyxScheduler.restoreState`: https://github.com/spotify/styx/pull/137/files#diff-a3fabd4f5e1e00cf48f40dfdfa3356f1R405

The lazy docker runner instantiation mechanism itself is a bit confusing and should perhaps be changed. Perhaps that could be a separate PR though.

honnix: Yeah, you are right. Disregard my comment. It feels like some design thoughts behind `RoutingDockerRunner`, not sure what it exactly is. There could be some reason making it lazy.

honnix:
Overall LGTM.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
fabriziodemaria: The connection towards k8s should already be setup correctly at this point, and `restore` could be treated as a best effort procedure instead of crashing. I don't have a strong opinion on this, both approaches are fine.

fabriziodemaria:
@danielnorberg as you mentioned in a comment above, it would be good to have a test around this, but overall looks good.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: The connection might well be setup/configured correctly but connections and requests can and do fail at any time. I would rather crash and retry instead of a best effort `restore` that might lose those completed executions. Also, we do not do best effort reads from bigtable and datastore when restoring state. It seems odd to treat k8s differently.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: I guess the reason for the laziness has to do with the ability to reconfigure the target GKE cluster at runtime.

honnix: Ah OK, that `globalDockerRunnerId` thing. It doesn't have to be lazy.

honnix:
+1

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
bergman: Why is the retry logic needed at all? It's part of a startup routine, I think it's ok to fail hard here.

Is it expected to fail intermittently?

bergman: Hmm, how would completed executions be lost if we crash hard on first error?

danielnorberg: Crashing hard and restarting on first error implies replaying all bigtable events etc again, which is relatively expensive. Thus I chose to have retries here to cheaply tolerate temporary netsplits.

PR #138, opened 2017-04-20
Add styx-client module

fabriziodemaria:
The CLI would use the `styx-client` to perform the requests, also part of this PR.

**TODO**
- The CLI version is part of the requests' header; (removed)

_NOTE_
I am not sure about the way the client handles errors and exceptions at this point. The `throwable` we add to a failing `CompletionStage` is different in different cases but very obscure to users. 
I think we should extract at least two kind of checked exceptions: `ClientErrorException` and `ServerErrorException`. A more generic `IOException` can also be added to the signatures for unexpected arguments/payloads. In this way errors can be better documented and easier to handle.
However I am not sure what are the best practices for this case.

### DIFF
--- a/styx-cli/pom.xml
+++ b/styx-cli/pom.xml
@@ -16,6 +16,11 @@
       <artifactId>styx-common</artifactId>
       <version>1.0.8-SNAPSHOT</version>
     </dependency>
+    <dependency>
+      <groupId>com.spotify</groupId>
+      <artifactId>styx-client</artifactId>
+      <version>1.0.8-SNAPSHOT</version>
+    </dependency>
 
     <dependency>
       <groupId>com.spotify</groupId>
@@ -69,6 +74,7 @@
       <artifactId>jackson-annotations</artifactId>
       <version>${jackson.version}</version>
     </dependency>
+
   </dependencies>
 
   <build>
--- a/styx-cli/pom.xml
+++ b/styx-cli/pom.xml
@@ -42,7 +42,6 @@
     <dependency>
       <groupId>net.sourceforge.argparse4j</groupId>
       <artifactId>argparse4j</artifactId>
-      <version>0.7.0</version>
     </dependency>
 
     <dependency>
@@ -72,7 +71,6 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-annotations</artifactId>
-      <version>${jackson.version}</version>
     </dependency>
 
   </dependencies>
--- a/styx-cli/src/main/java/com/spotify/styx/cli/Main.java
+++ b/styx-cli/src/main/java/com/spotify/styx/cli/Main.java
@@ -49,6 +49,7 @@ import com.spotify.styx.model.Workflow;
 import com.spotify.styx.model.WorkflowId;
 import com.spotify.styx.model.WorkflowInstance;
 import com.spotify.styx.model.WorkflowState;
+import com.spotify.styx.model.data.EventInfo;
 import com.spotify.styx.util.EventUtil;
 import com.spotify.styx.util.ParameterUtil;
 import java.io.IOException;
@@ -429,7 +430,7 @@ public final class Main {
 
     final ObjectNode json = (ObjectNode) jsonNode;
     final ArrayNode events = json.withArray("events");
-    final ImmutableList.Builder<CliOutput.EventInfo> eventInfos = ImmutableList.builder();
+    final ImmutableList.Builder<EventInfo> eventInfos = ImmutableList.builder();
     for (JsonNode eventWithTimestamp : events) {
       final long ts = eventWithTimestamp.get("timestamp").asLong();
       final JsonNode event = eventWithTimestamp.get("event");
@@ -439,14 +440,14 @@ public final class Main {
       try {
         Event typedEvent = OBJECT_MAPPER.convertValue(event, Event.class);
         eventName = EventUtil.name(typedEvent);
-        eventInfo = CliUtil.info(typedEvent);
+        eventInfo = EventUtil.info(typedEvent);
       } catch (IllegalArgumentException e) {
         // fall back to just inspecting the json
         eventName = event.get("@type").asText();
         eventInfo = "";
       }
 
-      eventInfos.add(CliOutput.EventInfo.create(ts, eventName, eventInfo));
+      eventInfos.add(EventInfo.create(ts, eventName, eventInfo));
     }
 
     cliOutput.printEvents(eventInfos.build());
--- a/styx-cli/src/main/java/com/spotify/styx/cli/Main.java
+++ b/styx-cli/src/main/java/com/spotify/styx/cli/Main.java
@@ -25,20 +25,29 @@ import com.spotify.apollo.core.Service;
 import com.spotify.apollo.core.Services;
 import com.spotify.apollo.environment.ApolloEnvironmentModule;
 import com.spotify.apollo.http.client.HttpClientModule;
+import com.spotify.styx.api.BackfillPayload;
+import com.spotify.styx.api.BackfillsPayload;
+import com.spotify.styx.api.ResourcesPayload;
+import com.spotify.styx.api.RunStateDataPayload;
+import com.spotify.styx.client.ApiErrorException;
 import com.spotify.styx.client.StyxApolloClient;
-import com.spotify.styx.client.util.ApiErrorException;
+import com.spotify.styx.model.Backfill;
+import com.spotify.styx.model.Resource;
 import com.spotify.styx.model.Workflow;
 import com.spotify.styx.model.WorkflowState;
+import com.spotify.styx.model.data.EventInfo;
 import com.spotify.styx.util.ParameterUtil;
 import java.io.IOException;
 import java.time.format.DateTimeParseException;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.function.BiConsumer;
+import java.util.concurrent.ExecutionException;
 import javaslang.Tuple;
+import javaslang.Tuple2;
 import net.sourceforge.argparse4j.ArgumentParsers;
 import net.sourceforge.argparse4j.impl.Arguments;
 import net.sourceforge.argparse4j.inf.Argument;
@@ -73,18 +82,6 @@ public final class Main {
   private final Service cliService;
   private final CliOutput cliOutput;
   private StyxApolloClient styxClient;
-  private final BiConsumer<Object, Throwable> throwableChecker = (response, throwable) -> {
-    if (throwable != null) {
-      if (throwable instanceof ApiErrorException) {
-        System.err.println("API error: " + throwable.getMessage());
-      } else if (throwable.getCause() == null) {
-        System.err.println(throwable.toString());
-      } else {
-        System.err.println(throwable.getCause().getMessage());
-      }
-      System.exit(EXIT_CODE_CLIENT_ERROR);
-    }
-  };
 
   private Main(
       StyxCliParser parser,
@@ -143,7 +140,6 @@ public final class Main {
     final Command command = namespace.get(COMMAND_DEST);
 
     try (Service.Instance instance = cliService.start()) {
-      final Service.Signaller signaller = instance.getSignaller();
       final Client client = ApolloEnvironmentModule.environment(instance).environment().client();
       styxClient = new StyxApolloClient(client, apiHost);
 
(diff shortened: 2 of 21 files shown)


styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -250,265 +232,158 @@ private void run() {
...
-      try {
-        Event typedEvent = OBJECT_MAPPER.convertValue(event, Event.class);
-        eventName = EventUtil.name(typedEvent);
-        eventInfo = CliUtil.info(typedEvent);
-      } catch (IllegalArgumentException e) {
-        // fall back to just inspecting the json
-        eventName = event.get("@type").asText();
-        eventInfo = "";
-      }
+    workflowFuture.thenCombine(workflowStateFuture, Tuple::of)
```
fabriziodemaria: I am looking for a way to achieve the same without using `Tuple`...

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -90,9 +60,9 @@
   private static final String PARAMETER_DEST = "parameter";
 
   private static final int EXIT_CODE_SUCCESS = 0;
-  private static final int EXIT_CODE_UNKNOWN_ERROR = 1;
-  private static final int EXIT_CODE_ARGUMENT_ERROR = 2;
-  private static final int EXIT_CODE_API_ERROR = 3;
+  private static final int EXIT_CODE_ARGUMENT_ERROR = 1;
```
bergman: I think 1 for unknown is nicer since whenever i see something other than 1 I assume that there was a specific error.

styx-cli/pom.xml
```diff
@@ -69,6 +69,11 @@
       <artifactId>jackson-annotations</artifactId>
       <version>${jackson.version}</version>
     </dependency>
+      <dependency>
+          <groupId>com.spotify</groupId>
+          <artifactId>styx-client</artifactId>
+          <version>1.0.6-SNAPSHOT</version>
```
honnix: We should probably starting using `<version>${project.version}</version>` for inter-module dependency. Or even move to `dependencyManagement`. Maybe a separated PR. However things like `<version>${jackson.version}</version>` in this pom.xml should be gone.

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
honnix: Agree.

styx-cli/pom.xml
TinaRanic: yeah, the versions for some of the dependencies, like argparse4j and jackson are already defined in the parent pom under deps management, so no need to specify them here.

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -250,265 +232,158 @@ private void run() {
...
-    final BackfillInput backfillInput = BackfillInput.create(
-        Instant.parse(start), Instant.parse(end), component, workflow, concurrency);
-
-    final ByteString payload = serialize(backfillInput);
-    final ByteString response = send(
-        Request.forUri(apiUrl("backfills"), "POST").withPayload(payload));
-    final Backfill backfill = deserialize(response, Backfill.class);
-    cliOutput.printBackfill(backfill);
+    styxClient.backfillCreate(component, workflow, start, end, concurrency)
+        .whenComplete((backfill, throwable) -> cliOutput.printBackfill(backfill));
```
honnix: Shouldn't we check throwable as well here? Since in the exception handler we don't really use response, `CompletionStage:exceptionally` would be better?

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -250,265 +232,158 @@ private void run() {
...
-      editedBackfillBuilder.concurrency(concurrency);
-    }
-    final ByteString putPayload = serialize(editedBackfillBuilder.build());
-    final ByteString putResponse = send(
-        Request.forUri(apiUrl("backfills", id), "PUT").withPayload(putPayload));
-
-    final Backfill newBackfill = deserialize(putResponse, Backfill.class);
-    cliOutput.printBackfill(newBackfill);
+    styxClient.backfillEditConcurrency(id, concurrency)
+        .whenComplete((backfill, throwable) -> cliOutput.printBackfill(backfill));
```
honnix: Same here, exception handling.

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -250,265 +232,158 @@ private void run() {
...
   }
 
-  private void backfillHalt() throws ExecutionException, InterruptedException {
+  private void backfillHalt() {
     final String id = namespace.getString(parser.backfillHaltId.getDest());
-    send(Request.forUri(apiUrl("backfills", id), "DELETE"));
+
+    styxClient.backfillHalt(id).whenComplete((none, throwable) -> cliOutput.printMessage(
```
honnix: exception handling

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
honnix: `Tuple` is not bad in this case.

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
```diff
@@ -0,0 +1,319 @@
...
+                          final String apiHost,
+                          final String userAgentIdentifier) {
+    this.client = client;
+    this.apiHost = apiHost;
+    this.userAgentIdentifier = userAgentIdentifier;
+  }
+
+  @Override
+  public CompletionStage<RunStateDataPayload> activeStates(Optional<String> componentId) {
+    String url = apiUrl("status", "activeStates");
```
honnix: This kind of URL handling feels quite primitive, we should probably switch to https://square.github.io/okhttp/3.x/okhttp/okhttp3/HttpUrl.html. Have a `TODO` maybe? Not to get scope creep.

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
```diff
@@ -0,0 +1,319 @@
...
+            final JsonNode event = eventWithTimestamp.get("event");
+
+            String eventName;
+            String eventInfo;
+            try {
+              Event typedEvent = OBJECT_MAPPER.convertValue(event, Event.class);
+              eventName = EventUtil.name(typedEvent);
+              eventInfo = EventUtil.info(typedEvent);
+            } catch (IllegalArgumentException e) {
+              // fall back to just inspecting the json
```
bergman: Should the client be this relaxed with the types? The CLI was made that way because people don't upgrade right away but I'm not sure that should apply to all clients.

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
fabriziodemaria: I was trying to follow an "order by severity" numbering, but reverted back to the original solution.

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
fabriziodemaria: The problem with `exceptionally` is that it accepts a `Function` that must return a value, so it gets a bit messy with `return null` and explicit casts to make it work. Probably worse than the unused `response` in the `BiConsumer` of `whenComplete`.

honnix:
Overall looks very good!

One thing from the users' point of view. How would a user instantiate `StyxApolloClient`? `StyxApolloClient client = new StyxApolloClient()` feels a bit odd.

How about something like this? Maybe not make much sense.

```
interface StyxClient extends StyxWorkflowClient, StyxBackfillClient, StyxResourceClient,
                             StyxSchedulerClient, StyxStatusClient {
}

class StyxApolloClient implements StyxClient {
}

StyxClient styxClient = StyxClient.builder()
    .client(client)
    .apiHost(apiHost)
    .userAgentIdentifier(userAgentIdentifier)
    .build();

StyxBackfillClient styxClient = StyxClient.builder()
    .client(client)
    .apiHost(apiHost)
    .userAgentIdentifier(userAgentIdentifier)
    .buildBackfillClient(); // just a cast here
```

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
```diff
@@ -0,0 +1,319 @@
...
+  private final Client client;
+  private final String apiHost;
+  private final String userAgentIdentifier;
+
+  public StyxApolloClient(final Client client,
+                          final String apiHost,
+                          final String userAgentIdentifier) {
+    this.client = client;
+    this.apiHost = apiHost;
+    this.userAgentIdentifier = userAgentIdentifier;
```
honnix: Hmm, this should be part of the client now, because what we want is to block the client, not cli.

honnix: `this.userAgentIdentifier = StyxApolloClient.class.getPackage().getImplementationVersion();`

honnix:
If for sure we can differentiate client and server error, it makes sense to have specific exceptions. Then users can catch and see the problem.

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
```diff
@@ -0,0 +1,317 @@
...
+    final String url = apiUrl("backfills", backfillId);
+    return executeRequest(Request.forUri(url), BackfillPayload.class);
+  }
+
+  @Override
+  public CompletionStage<BackfillsPayload> backfillList(Optional<String> componentId,
+                                                        Optional<String> workflowId,
+                                                        boolean showAll,
+                                                        boolean status) {
+    List<String> queries = new ArrayList<>();
```
honnix: final

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -253,262 +235,161 @@ private void run() {
...
-        .payload().orElse(ByteString.EMPTY);
-    final ByteString statePayload = requireSuccess(stateResponse.get())
-        .payload().orElse(ByteString.EMPTY);
+  private void workflowShow() throws ExecutionException, InterruptedException {
+    final String component = namespace.getString(parser.workflowShowComponentId.getDest());
+    final String workflow = namespace.getString(parser.workflowShowWorkflowId.getDest());
 
-    final Workflow workflow = deserialize(workflowPayload, Workflow.class);
-    final WorkflowState workflowState = deserialize(statePayload, WorkflowState.class);
+    CompletableFuture<Workflow> workflowFuture =
```
honnix: final

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
```diff
@@ -0,0 +1,317 @@
...
+          return response;
+        default:
+          final String message = response.status().code() + " " + response.status().reasonPhrase();
+          throw new ApiErrorException(message, response.status().code());
+      }
+    });
+  }
+
+  private String apiUrl(String... parts) {
+    List<String> encodedPartsList = Arrays.stream(parts).map(part -> {
```
honnix: final

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
```diff
@@ -0,0 +1,317 @@
...
+        return URLEncoder.encode(part, UTF_8);
+      } catch (UnsupportedEncodingException e) {
+        throw new RuntimeException(e);
+      }
+    }).collect(Collectors.toList());
+    return "http://" + apiHost + STYX_API_ENDPOINT + "/" + String.join("/", encodedPartsList);
+  }
+
+  private String addQueryToApiUrl(String url, List<String> queries) {
+    List<String> encodedQueryParts = queries.stream().map(query -> {
```
honnix: final

honnix:
LGTM apart from a few `final`s. Didn't check all actual logic. I assume you have tested against styx service. Would be nicer if we can add test cases later in another PR. Regarding that URL handling comment, I would still prefer adding a TODO and later using the offering from okhttp.

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
fabriziodemaria: I think it is fine to be relaxed in the client for this API as well, since `EventInfo` is mainly "informational" thus it feels ok to make a best effort in constructing it. This detail about forward compatibility can be mentioned in the Javadocs though, I will add it.

bergman:
👍

PR #139, opened 2017-04-21
Tag 'audit' for some logs

honnix:
@ulzha


ulzha:
Inconsistent capitalization (I mean lowercase "workflow") with other log messages (which start with uppercase), otherwise LGTM :)

PR #141, opened 2017-04-21
middleware to fire audit log for mutable operations

honnix:


honnix:
It seems there is no obvious way to get client IP at this level though.

fabriziodemaria:
LGTM

PR #143, opened 2017-04-25
s/SchedulerResource/SchedulerResourceProxy

danielnorberg:
Rename the `SchedulerResource` class in the api module to `SchedulerProxyResource` in order to resolve the conflict with `SchedulerResource` in the scheduler module. The full class names, including package names, were exactly the same, causing e.g. the standalone service to fail as only one of them would be loaded.


danielnorberg:
@honnix PTAL

honnix:
+1

PR #144, opened 2017-04-25
readme: fix schedules example

danielnorberg:
`@` cannot start a token in yaml. I.e. this example would have to be `schedule: '@hourly'`, which is an eyesore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ To define a schedule, simply write a  yaml file to `/etc/styx` (given the above 
 ```yaml
 schedules:
   - id: my-workflow
-    schedule: @hourly
+    schedule: hourly
     docker_image: my-workflow:0.1
     docker_args: ['./run.sh', '{}']
 ```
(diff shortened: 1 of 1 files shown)


danielnorberg:
@honnix PTAL

README.md
```diff
@@ -108,7 +108,7 @@ To define a schedule, simply write a  yaml file to `/etc/styx` (given the above
 ```yaml
 schedules:
   - id: my-workflow
-    schedule: @hourly
+    schedule: hourly
```
honnix: Or '@hourly' and it intended to be?

honnix: It feels rouz wanted to deprecate hourly, daily, etc.

danielnorberg: @rouzwawi wdyt?

bergman:
I guess the examples should have had either a comment about needing quotes around values that start with `@` or removal of the `@`?

PR #145, opened 2017-04-25
use netty-tcnative-boringssl-static uberjar

danielnorberg:
The uberjar contains native code for windows, linux and osx. Eliminates risk of build/run-time platform mismatch.


honnix:
Nice! +1

PR #146, opened 2017-04-26
[wip] support service account key creation

honnix:
start from here?

### DIFF
--- a/styx-api-service/src/test/java/com/spotify/styx/api/deprecated/WorkflowResourceTest.java
+++ b/styx-api-service/src/test/java/com/spotify/styx/api/deprecated/WorkflowResourceTest.java
@@ -89,7 +89,8 @@ public class WorkflowResourceTest extends VersionedApiTest {
 
   private static final WorkflowConfiguration WORKFLOW_CONFIGURATION =
       WorkflowConfiguration
-          .create("bar", Schedule.DAYS, empty(), empty(), empty(), empty(), empty(), emptyList());
+          .create("bar", Schedule.DAYS, empty(), empty(), empty(), empty(), empty(), empty(),
+                  emptyList());
 
   private static final Workflow WORKFLOW =
       Workflow.create("foo", URI.create("/hejhej"), WORKFLOW_CONFIGURATION);
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -35,6 +35,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.hash.Hashing;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.spotify.styx.model.Event;
 import com.spotify.styx.model.EventVisitor;
@@ -65,7 +66,7 @@ import io.fabric8.kubernetes.client.Watch;
 import io.fabric8.kubernetes.client.Watcher;
 import java.io.IOException;
 import java.net.ProtocolException;
-import java.util.Base64;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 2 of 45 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -126,37 +136,120 @@
...
+        throw new InvalidExecutionException(
+            "Referenced secret '" + specSecret.name() + "' was not found");
+      } else {
+        LOG.info("[AUDIT] Workflow {} refers to secret {}",
+                 workflowInstance.workflowId(), specSecret.name());
+      }
+    }
+  }
+
+  private static String buildSecretName(String serviceAccount) {
```
danielnorberg: include sanitized prefix substrings of the component, workflow and service account names in the secret name to ease debugging?

e.g. `styx-wf-sa-keys.foo-comp.bar-workfl.baz-acct-d34db33fc4f3d00d` ?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -126,37 +136,120 @@
...
+      spec = spec.addNewVolume()
+          .withName(STYX_WORKFLOW_SA_SECRET_NAME)
+          .withNewSecret()
+          .withSecretName(buildSecretName(runSpec.serviceAccount().get()))
+          .endSecret()
+          .endVolume();
+      container = container
+          .addToVolumeMounts(new VolumeMount(STYX_MANAGED_WORKFLOW_SA_SECRET_MOUNT_PATH,
+                                             STYX_WORKFLOW_SA_SECRET_NAME, true))
+          // TODO: do we need set this env as default value?
```

This way we can migrate to network (metadata server) based SA credentials in the future without the user having to care.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -126,37 +136,120 @@
...
-      });
-
       final Pod pod = client.pods().create(createPod(workflowInstance, runSpec));
       return pod.getMetadata().getName();
     } catch (KubernetesClientException kce) {
       throw new IOException("Failed to create Kubernetes pod", kce);
     }
   }
 
+  private void ensureSecret(WorkflowInstance workflowInstance, RunSpec runSpec) {
```
danielnorberg: Perhaps SA key creation could be factored into a separate class?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
honnix: Do we want store replicated secrets? Now we have one `service_account` mapped to one `kube secret`. By prefixing component and workflow, we will end up having many secrets having keys pointing to the same SA.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
honnix: 👍 You are on top of that already.

styx-api-service/src/test/java/com/spotify/styx/api/deprecated/WorkflowResourceTest.java
```diff
@@ -89,7 +89,8 @@ public WorkflowResourceTest(Api.Version version) {
 
   private static final WorkflowConfiguration WORKFLOW_CONFIGURATION =
       WorkflowConfiguration
-          .create("bar", Schedule.DAYS, empty(), empty(), empty(), empty(), empty(), emptyList());
+          .create("bar", Schedule.DAYS, empty(), empty(), empty(), empty(), empty(), empty(),
```
bergman: I vote for builders.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -109,54 +123,156 @@
...
+                                    .withName(secretName)
+                                    .withAnnotations(annotations)
+                                    .endMetadata()
+                                    .withData(keys)
+                                    .build());
+        LOG.info("[AUDIT] Secret {} created for service account {}", secretName, serviceAccount);
+      }
+    }
+
+    if (runSpec.secret().isPresent()) {
```
honnix: @bergman @fabriziodemaria @danielnorberg let's continue here. do we want them to co-eixst.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
honnix: There is however one issue. This env will overwrite what's put in `Dockerfile` by user when building image.

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesDockerRunnerTest.java
```diff
@@ -140,6 +173,89 @@ public void tearDown() throws Exception {
     kdr.close();
   }
 
+  // TODO
```
danielnorberg: WIP! Wasn't immediately obvious to me how to fix the mocks in these tests so I commented them out.

danielnorberg: Temporarily, will fix later unless you do it first.

bergman:
👍

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -111,106 +125,139 @@
...
+    if (runSpec.serviceAccount().isPresent()) {
+      spec.addNewVolume()
+          .withName(STYX_WORKFLOW_SA_SECRET_NAME)
+          .withNewSecret()
+          .withSecretName(buildSecretName(runSpec.serviceAccount().get()))
+          .endSecret()
+          .endVolume();
+      container
+          .addToVolumeMounts(new VolumeMountBuilder()
+              .withMountPath(STYX_WORKFLOW_SA_SECRET_MOUNT_PATH)
```
fabriziodemaria: What happens in the unlikely case that the mounting path is the same of the one for the old-fashioned secret and both are mounted? If that not allowed, should we throw an exception with a custom message as we do, for example, in case of failure in creating the keys?

styx-api-service/src/test/java/com/spotify/styx/api/deprecated/WorkflowResourceTest.java
honnix: Let's keeping voting for it. :D

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
honnix: I wouldn't see that happening in real world though. Let's keep it in mind, and do it in another PR if really need. This PR is big enough. ;)

danielnorberg:
🎉

PR #147, opened 2017-04-27
jacoco coverage + coveralls

danielnorberg:


honnix:
Nice! One comment, we will need to exclude generated source.

danielnorberg:
@honnix now excluding source and the jacoco coverage report is also visible in circleci: https://1334-74431644-gh.circle-artifacts.com/0/tmp/circle-artifacts.6HcPSjY/jacoco-aggregate/index.html

honnix:
+1 Super!

PR #150, opened 2017-04-28
add key creation + allow both SA key & secret

danielnorberg:
wip

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -160,14 +160,11 @@ class KubernetesDockerRunner implements DockerRunner {
       if (secret == null) {
         final String serviceAccount = runSpec.serviceAccount().get();
 
-        // TODO: need gcp project of the service account
-        final String project = "TODO";
-
         final ServiceAccountKey jsonKey;
         final ServiceAccountKey p12Key;
         try {
-          jsonKey = serviceAccountKeyManager.createJsonKey(project, serviceAccount);
-          p12Key = serviceAccountKeyManager.createP12Key(project, serviceAccount);
+          jsonKey = serviceAccountKeyManager.createJsonKey(serviceAccount);
+          p12Key = serviceAccountKeyManager.createP12Key(serviceAccount);
         } catch (IOException e) {
           logger.error("Failed to create service account keys", e);
           throw new RuntimeException(e);
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -31,12 +31,14 @@ import com.github.rholder.retry.Retryer;
 import com.github.rholder.retry.RetryerBuilder;
 import com.github.rholder.retry.StopStrategies;
 import com.github.rholder.retry.WaitStrategies;
+import com.google.api.services.iam.v1.model.ServiceAccountKey;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
 import com.google.common.hash.Hashing;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.spotify.styx.ServiceAccountKeyManager;
 import com.spotify.styx.model.Event;
 import com.spotify.styx.model.EventVisitor;
 import com.spotify.styx.model.ExecutionDescription;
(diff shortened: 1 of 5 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -145,23 +152,38 @@ public String start(WorkflowInstance workflowInstance, RunSpec runSpec) throws I
...
       final String secretName = buildSecretName(runSpec.serviceAccount().get());
       final Secret secret = client.secrets().withName(secretName).get();
       if (secret == null) {
         final String serviceAccount = runSpec.serviceAccount().get();
 
-        // TODO: create keys calling GCP API, base64 already
-        final String jsonKey = "";
-        final String p12Key = "";
+        // TODO: need gcp project of the service account
+        final String project = "TODO";
```
honnix: I think according to IAM API description, project is optional, if `-` is given, project will be figured out from service account name.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -119,24 +123,27 @@
...
-    this(client, stateManager, stats, DEFAULT_POLL_PODS_INTERVAL_SECONDS);
+  KubernetesDockerRunner(KubernetesClient client, StateManager stateManager, Stats stats,
+      ServiceAccountKeyManager serviceAccountKeyManager) {
+    this(client, stateManager, stats, serviceAccountKeyManager, DEFAULT_POLL_PODS_INTERVAL_SECONDS);
   }
 
   @Override
   public String start(WorkflowInstance workflowInstance, RunSpec runSpec) throws IOException {
-    ensureSecret(workflowInstance, runSpec);
+    ensureSecrets(workflowInstance, runSpec);
```
honnix: Why plural? There is only one secret either case.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -145,23 +152,38 @@ public String start(WorkflowInstance workflowInstance, RunSpec runSpec) throws I
...
+        final ServiceAccountKey p12Key;
+        try {
+          jsonKey = serviceAccountKeyManager.createJsonKey(project, serviceAccount);
+          p12Key = serviceAccountKeyManager.createP12Key(project, serviceAccount);
+        } catch (IOException e) {
+          logger.error("Failed to create service account keys", e);
+          throw new RuntimeException(e);
+        }
+
+        final Map<String, String> keys = ImmutableMap.of(
```
honnix: Yeah, that was the helper class I was trying to recall. Thanks!

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -171,7 +193,9 @@ private void ensureSecret(WorkflowInstance workflowInstance, RunSpec runSpec) {
                                     .build());
         LOG.info("[AUDIT] Secret {} created for service account {}", secretName, serviceAccount);
       }
-    } else if (runSpec.secret().isPresent()) {
+    }
+
+    if (runSpec.secret().isPresent()) {
```
honnix: I think if we don't want to support them coexist, it's better here to be clear that we are not expecting both `service_account` and `secret`.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: I think it's required. It's part of the POST request url.

`projects/{PROJECT_ID}/serviceAccounts/{SERVICE_ACCOUNT_EMAIL}`

honnix: ```The resource name of the service account in the following format: projects/{PROJECT_ID}/serviceAccounts/{SERVICE_ACCOUNT_EMAIL}. Using - as a wildcard for the project will infer the project from the account. The account value can be the email address or the uniqueId of the service account.```

danielnorberg: Wow ok totally missed that =)

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: I was thinking that we'd let them coexist. We want to gradually migrate and the old-style secret still makes sense for oss?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
honnix: I tried that in your poc project and it worked.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
honnix: That can still be achieved by having some of the workflows using `service_account` while others using `secret`, in the same component. It feels weird having both. If the user doesn't feel like to migrate now because they have other keys stored in kube, they don't have to migrate and everything should still work.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: I was thinking that we'd let both `secret` and `service_account` coexist.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: Old-style secrets still make sense for OSS I think. Smaller organizations might want to use k8s secrets.

For us I think having both allows users to migrate their workflows. Not allowing both makes it harder because you have to make sure to change both the container and the workflow config at the same time.

danielnorberg: And I think it's confusing behavior that adding `service_account` to a workflow then effectively disables the existing `secret` field.

honnix: @fabriziodemaria @bergman Get everybody here since the decision affects internal code as well. Having to change both workflow config and container at the same time is a good argument.

PR #151, opened 2017-04-28
small tweaks

bergman:


bergman:
supposed to go on top of #146

PR #153, opened 2017-04-29
let's be paranoid

honnix:
Check mount path specified by `secret` and stop execution if it
collides with reserved path.

As `secret` seems more problematic, we check it prior to servicr account
to be a bit lazy.


honnix:
covered by #160

Issue #118, opened 2017-03-28
add API blueprint for deprecated API versions


honnix:
For v0 and v1, we will need another blueprint to keep track those.

PR #154, opened 2017-04-29
api blueprint for deprecated version

honnix:
addressing #118

### DIFF
--- a/doc/api-deprecated.apib
+++ b/doc/api-deprecated.apib
@@ -1,7 +1,8 @@
 # Styx API specification
 
 This specification is for deprecated API versions. Users are recommended to use
-[latest version] if backward compatibility is not an issue.
+[latest version] if backward compatibility is not an issue. In this document,
+`endpoint` is interchangeable with `workflow` when describing a workflow.
 
 This document is written using [api-blueprint]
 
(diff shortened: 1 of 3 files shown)


honnix:
@fabriziodemaria PTAL

doc/api.apib
```diff
@@ -31,40 +22,43 @@ aglio -i doc/api.apib -o api.html --theme-template triple --theme-variables stre
 
 Resources related to the Workflows registered in Styx.
 
-## Workflow [/v2/workflows/{component}/{workflow_id}]
+## Workflow [/v2/workflows/{component_id}/{workflow_id}]
 
 + Parameters
-    + component: `styx-canary` (string) - Workflow Component
-    + workflow_id: `LuigiCanary` (string) - Workflow ID
```
fabriziodemaria: Should we change the `luigi-canary-dummy:dummy` entries as well? Perhaps remove all the Luigi parts :)

doc/api-deprecated.apib
```diff
@@ -0,0 +1,558 @@
...
+
+        {
+            "parameter": "2017-01-01T01",
+            "workflow_id": {
+                "component_id": "styx-canary",
+                "endpoint_id": "StyxCanary"
+            }
+        }
+
+# Group Global Config
```
fabriziodemaria: Actually should we rename this `Group Styx Config` to reflect the class name as in all the other cases? Same in the file for APIs version 2. I am not sure how I came up with this "Global" name before..

fabriziodemaria:
Minor things commented, nothing important! Feel free to merge as is 👍

doc/api.apib
honnix: Will do

honnix:
incorporate #201 to avoid a rather big merge.

PR #158, opened 2017-05-01
bump gcp deps

danielnorberg:
Latest datastore + bigtable libs + required deps

### DIFF
--- a/styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
+++ b/styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
@@ -87,7 +87,7 @@ public class BackfillResourceTest extends VersionedApiTest {
 
   private AggregateStorage storage = new AggregateStorage(
       bigtable,
-      localDatastore.options().service(),
+      localDatastore.getOptions().getService(),
       Duration.ZERO);
 
   private static final Backfill BACKFILL_1 = Backfill.newBuilder()
@@ -140,7 +140,7 @@ public class BackfillResourceTest extends VersionedApiTest {
   @AfterClass
   public static void tearDownClass() throws Exception {
     if (localDatastore != null) {
-      localDatastore.stop();
+      localDatastore.stop(org.threeten.bp.Duration.ofSeconds(30));
     }
   }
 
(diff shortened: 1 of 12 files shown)


danielnorberg:
@bergman PTAL

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -140,7 +141,11 @@ public static void setUpClass() throws Exception {
   @AfterClass
   public static void tearDownClass() throws Exception {
     if (localDatastore != null) {
-      localDatastore.stop();
+      try {
+        localDatastore.stop(org.threeten.bp.Duration.ofSeconds(30));
```
bergman: This looks like a weird place to find joda-time.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -22,10 +22,10 @@
 
 import static com.spotify.styx.serialization.Json.OBJECT_MAPPER;
 
+import com.google.cloud.Timestamp;
```
bergman: is Timestamp the new DateTime? Are they compatible?

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
danielnorberg: it's not joda-time but a library with a few backports of the new java 8 time api classes. I guess the datastore library uses it to not require java 8.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: Afaict yes http://googlecloudplatform.github.io/google-cloud-java/0.17.1/apidocs/com/google/cloud/datastore/BaseEntity.html#getTimestamp-java.lang.String-

danielnorberg:
Ran this against real datastore & bigtable, was able to successfully trigger and run a workflow etc.

PR #160, opened 2017-05-01
validate & recreate SA keys if necessary

danielnorberg:
Also makes SA key secret checking and creation single threaded to avoid concurrent executions of workflows using the same service account from stepping on each others feet.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
@@ -20,6 +20,7 @@
 
 package com.spotify.styx;
 
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.services.iam.v1.Iam;
 import com.google.api.services.iam.v1.model.CreateServiceAccountKeyRequest;
 import com.google.api.services.iam.v1.model.ServiceAccountKey;
@@ -41,6 +42,33 @@ public class ServiceAccountKeyManager {
   public ServiceAccountKey createP12Key(String serviceAccount) throws IOException {
     return createKey(serviceAccount, new CreateServiceAccountKeyRequest()
         .setPrivateKeyType("TYPE_PKCS12_FILE"));
+  }
+
+  public boolean serviceAccountExists(String serviceAccount) throws IOException {
+    try {
+      iam.projects().serviceAccounts().get("projects/-/serviceAccounts/" + serviceAccount)
+          .execute();
+      return true;
+    } catch (GoogleJsonResponseException e) {
+      if (e.getStatusCode() == 404) {
+        return false;
+      }
+      throw e;
+    }
+  }
+
+  public boolean keyExists(String keyName) throws IOException {
+    try {
+      iam.projects().serviceAccounts().keys()
+          .get(keyName)
+          .execute();
+      return true;
+    } catch (GoogleJsonResponseException e) {
+      if (e.getStatusCode() == 404) {
+        return false;
+      }
+      throw e;
+    }
   }
 
   private ServiceAccountKey createKey(String serviceAccount,
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -23,6 +23,7 @@ package com.spotify.styx.docker;
 import static com.spotify.styx.docker.KubernetesPodEventTranslator.translate;
 import static com.spotify.styx.serialization.Json.OBJECT_MAPPER;
 import static com.spotify.styx.state.RunState.State.RUNNING;
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.stream.Collectors.toSet;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
@@ -67,7 +68,6 @@ import io.fabric8.kubernetes.client.Watch;
 import io.fabric8.kubernetes.client.Watcher;
 import java.io.IOException;
 import java.net.ProtocolException;
-import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
@@ -80,15 +80,11 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * A {@link DockerRunner} implementation that submits container executions to a Kubernetes cluster.
  */
 class KubernetesDockerRunner implements DockerRunner {
-
-  private static final Logger logger = LoggerFactory.getLogger(KubernetesDockerRunner.class);
 
   private static final String NAMESPACE = "default";
   static final String STYX_RUN = "styx-run";
@@ -130,6 +126,8 @@ class KubernetesDockerRunner implements DockerRunner {
 
   private Watch watch;
 
+  private final Object secretMutationLock = new Object() {};
+
   KubernetesDockerRunner(NamespacedKubernetesClient client, StateManager stateManager, Stats stats,
       ServiceAccountKeyManager serviceAccountKeyManager, int pollPodsIntervalSeconds) {
     this.stateManager = Objects.requireNonNull(stateManager);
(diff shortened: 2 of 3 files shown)


danielnorberg:
@honnix PTAL

styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
```diff
@@ -43,6 +44,20 @@ public ServiceAccountKey createP12Key(String serviceAccount) throws IOException
         .setPrivateKeyType("TYPE_PKCS12_FILE"));
   }
 
+  public boolean keyExists(String serviceAccount, String keyName) throws IOException {
+    try {
+      ServiceAccountKey key = iam.projects().serviceAccounts().keys()
```
honnix: no need of this variable.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -208,7 +168,120 @@ private void ensureSecrets(WorkflowInstance workflowInstance, RunSpec runSpec) {
...
+    final String jsonKeyJson = data.get(STYX_WORKFLOW_SA_JSON_KEY);
+    final String p12KeyJson = data.get(STYX_WORKFLOW_SA_P12_KEY);
+    if (jsonKeyJson == null || p12KeyJson == null) {
+      return false;
+    }
+
+    final ServiceAccountKey jsonKey;
+    final ServiceAccountKey p12Key;
+    try {
+      jsonKey = Json.OBJECT_MAPPER.readValue(jsonKeyJson, ServiceAccountKey.class);
```
honnix: It's only private key gets stored in kube secret, not the complete JSON. https://github.com/spotify/styx/pull/160/files#diff-d96c43ca98b2c4d50df1a802530ab077R266

Get key name from annotation instead?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -208,7 +168,120 @@ private void ensureSecrets(WorkflowInstance workflowInstance, RunSpec runSpec) {
...
+          workflowInstance.workflowId(), e);
+      return false;
+    }
+
+    final boolean keysExist;
+    try {
+      keysExist = serviceAccountKeyManager.keyExists(serviceAccount, jsonKey.getName())
+          && serviceAccountKeyManager.keyExists(serviceAccount, p12Key.getName());
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
honnix: Let's stick to `InvalidExecutionException`.

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesDockerRunnerTest.java
```diff
@@ -25,9 +25,11 @@
 import static com.spotify.styx.docker.KubernetesPodEventTranslatorTest.running;
 import static com.spotify.styx.docker.KubernetesPodEventTranslatorTest.terminated;
 import static com.spotify.styx.docker.KubernetesPodEventTranslatorTest.waiting;
+import static java.nio.charset.StandardCharsets.UTF_8;
```
honnix: Unused import

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -155,49 +152,12 @@ public String start(WorkflowInstance workflowInstance, RunSpec runSpec) throws I
...
-    }
+    ensureServiceAccountKeySecret(workflowInstance, runSpec);
+    ensureCustomSecret(workflowInstance, runSpec);
+  }
 
-    if (runSpec.secret().isPresent()) {
-      final WorkflowConfiguration.Secret specSecret = runSpec.secret().get();
+  private void ensureCustomSecret(WorkflowInstance workflowInstance, RunSpec runSpec) {
+    runSpec.secret().ifPresent(specSecret -> {
       final Secret secret = client.secrets().withName(specSecret.name()).get();
```
honnix: Feel like to incorporate this check https://github.com/spotify/styx/pull/153/files#diff-d96c43ca98b2c4d50df1a802530ab077R160 ? Or I can open another PR after this one has been merged. Having #153 opened doesn't make much sense due to the massive change brought by this PR.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -208,7 +168,120 @@ private void ensureSecrets(WorkflowInstance workflowInstance, RunSpec runSpec) {
         LOG.info("[AUDIT] Workflow {} refers to secret {}",
                  workflowInstance.workflowId(), specSecret.name());
       }
+    });
+  }
+
+  private void ensureServiceAccountKeySecret(WorkflowInstance workflowInstance, RunSpec runSpec) {
```
honnix: So as we discussed, this method is supposed to be `synchronized` as the first simplest approach.

danielnorberg:
@honnix PTAL again

honnix:
@bergman PTAL as well

styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
```diff
@@ -43,6 +44,33 @@ public ServiceAccountKey createP12Key(String serviceAccount) throws IOException
...
+  public boolean serviceAccountExists(String serviceAccount) throws IOException {
+    try {
+      iam.projects().serviceAccounts().get("projects/-/serviceAccounts/" + serviceAccount)
+          .execute();
+      return true;
+    } catch (GoogleJsonResponseException e) {
+      if (e.getStatusCode() == 404) {
+        return false;
+      }
+      throw e;
```
ulzha: In most of the code we use `throw Throwables.propagate(e)`...

honnix: There was a discussion that `Throwables.propagate` has been deprecated by Google. Also in this case we don't need a `RuntimeException` I think.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -215,12 +185,101 @@ private void ensureSecrets(WorkflowInstance workflowInstance, RunSpec runSpec) {
...
+    }
+
+    final String serviceAccount = runSpec.serviceAccount().get();
+
+    // Check that the service account exists
+    final boolean serviceAccountExists = serviceAccountKeyManager.serviceAccountExists(serviceAccount);
+    if (!serviceAccountExists) {
+      LOG.error("[AUDIT] Workflow {} refers to non-existent service account {}", workflowInstance.workflowId(),
+          serviceAccount);
+      throw new InvalidExecutionException("Referenced service account '" + serviceAccount + "' was not found");
```
ulzha: Inconsistent quoting (here `'`s surround service account and not elsewhere) :] I do see that this was unchanged in this PR though.

honnix: Good catch. I don't think there is a need to quote. It's a full string without any space anyway.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -215,12 +186,101 @@ private void ensureSecrets(WorkflowInstance workflowInstance, RunSpec runSpec) {
...
+
+      // Check if we have a valid service account key secret already
+      final Secret existingSecret = client.secrets().withName(secretName).get();
+      if (existingSecret != null) {
+
+        if (serviceAccountKeysExist(existingSecret)) {
+          return;
+        }
+
+        LOG.warn("[AUDIT] Service account keys have been deleted for {}, recreating",
```
ulzha: This is a lie the first time round, they just haven't been created, right?

ulzha: Also it seems noisy to `warn` about it - isn't `info` more appropriate, as key rotation is legitimate business?

(Suggested convention: `error` for internal errors, "this line should never happen" type of things; `warn` for external errors that will stop workflows from running if they persist; `info` or lower for everything else.)

honnix: If the keys are not created the secret shouldn't be in kube. We check secret first.

honnix: Yeah. That I agree.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -215,12 +186,101 @@ private void ensureSecrets(WorkflowInstance workflowInstance, RunSpec runSpec) {
...
+
+      final Map<String, String> keys = ImmutableMap.of(
+          STYX_WORKFLOW_SA_JSON_KEY, jsonKey.getPrivateKeyData(),
+          STYX_WORKFLOW_SA_P12_KEY, p12Key.getPrivateKeyData()
+      );
+
+      final Map<String, String> annotations = ImmutableMap.of(
+          STYX_WORKFLOW_SA_JSON_KEY_NAME_ANNOTATION, jsonKey.getName(),
+          STYX_WORKFLOW_SA_P12_KEY_NAME_ANNOTATION, p12Key.getName(),
+          STYX_WORKFLOW_SA_SECRET_ANNOTATION, serviceAccount
```
ulzha: Isn't it confusing to call it "secret annotation" if it stores serviceAccount? I expect "SA_ID_ANNOTATION" or "SA_ANNOTATION".

honnix: Yeah. Confusing indeed. Will fix.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -215,12 +186,101 @@ private void ensureSecrets(WorkflowInstance workflowInstance, RunSpec runSpec) {
...
+          .withNewMetadata()
+          .withName(secretName)
+          .withAnnotations(annotations)
+          .endMetadata()
+          .withData(keys)
+          .build();
+
+      client.secrets().create(newSecret);
+
+      LOG.info("[AUDIT] Secret {} created for {} referred to by workflow {}",
```
ulzha: "For" might read ambiguously, as in "for [use by this service account]", while it's actually "storing [this service account's private key!!!11]".

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesDockerRunnerTest.java
```diff
@@ -192,6 +204,24 @@ public void shouldThrowIfSecretNotExist() throws IOException, StateManager.IsClo
...
+  public void shouldThrowIfMountToReservedPath() throws IOException, StateManager.IsClosed {
+    when(secrets.withName(any(String.class))).thenReturn(namedResource);
+    when(namedResource.get()).thenReturn(null);
+    when(k8sClient.secrets()).thenReturn(secrets);
+    Pod createdPod = KubernetesDockerRunner.createPod(WORKFLOW_INSTANCE, RUN_SPEC_WITH_SECRET_AND_SA);
+    when(pods.create(any(Pod.class))).thenReturn(createdPod);
+
+    stateManager.receive(Event.terminate(WORKFLOW_INSTANCE, Optional.of(0)));
+    stateManager.receive(Event.success(WORKFLOW_INSTANCE));
+    kdr.close();
```
ulzha: Why here too? It seems it's done in `tearDown`?

honnix: In general cases we create in setup and for a few cases the default is not good enough so we close it and create a new one. Not very good indeed. Shall we address this in separated PR?

ulzha:
LGTM apart from those naming and wording nitpickings.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
honnix: How about "Secret [secret1] created to store keys of [sa] referred by ..."?

PR #161, opened 2017-05-01
disallow managed SA key secret access

danielnorberg:


danielnorberg:
@honnix PTAL

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -198,6 +198,12 @@ private void ensureSecrets(WorkflowInstance workflowInstance, RunSpec runSpec) {
 
     if (runSpec.secret().isPresent()) {
       final WorkflowConfiguration.Secret specSecret = runSpec.secret().get();
+      if (specSecret.name().startsWith(STYX_WORKFLOW_SA_SECRET_NAME)) {
+        LOG.info("[AUDIT] Workflow {} refers to secret {} in the managed service account key namespace, "
```
honnix: Try some other word than `namespace`? It makes me think of kube namespace.

honnix:
A minor comment. Otherwise 👍

PR #162, opened 2017-05-01
Circleci 2.0

danielnorberg:
CircleCI 2.0 is beta but it seems to provide ~ 3x faster builds (!) 🎉


honnix:
there is another config file in `gh-pages`, anything to change there?

danielnorberg:
I don't think the config in the `gh-pages` needs to be changed for the master branch to use circleci 2.0.

honnix:
:+1: doesn't hurt to upgrade even it is beta. The gain is good enough.

honnix:
@danielnorberg I think we need to tweak the threshold a bit in `coveralls.io`. I don't seem to have permission since you set up styx there. It's pretty annoying that build marked as failure due to 0.01% drop of coverage.

danielnorberg:
@honnix Yeah, also thinking that we might want to look at some other coverage service instead. There's quite a few ones out there that look like they might be more useful than coveralls.

- https://codecov.io/
- https://codeclimate.com/
- https://www.codacy.com/

Some of them also do code analysis and style checking etc, which seems potentially useful.

PR #163, opened 2017-05-02
fabric8: use more builders and less fluent api

bergman:


honnix:
LGTM. Is it covered by the test case added a few days ago?

bergman:
These changes are mostly covered but some detailed review could still be useful.

honnix:
Verified by printing `pod` and doing diff. Hard way. 👍

Issue #164, opened 2017-05-02
No longer possible to point cli to a specific port


danielnorberg:
```
STYX_CLI_HOST=localhost:8080 styx t schedules.yaml test-workflow 2017-04-04
Exception in thread "main" java.lang.IllegalArgumentException: unexpected host: localhost:8080
	at com.squareup.okhttp.HttpUrl$Builder.host(HttpUrl.java:705)
	at com.spotify.styx.client.StyxApolloClient.getUrlBuilder(StyxApolloClient.java:354)
	at com.spotify.styx.client.StyxApolloClient.triggerWorkflowInstance(StyxApolloClient.java:164)
	at com.spotify.styx.cli.Main.triggerWorkflowInstance(Main.java:375)
	at com.spotify.styx.cli.Main.run(Main.java:157)
	at com.spotify.styx.cli.Main.main(Main.java:137)
```

@fabriziodemaria I think this might be caused by this change: https://github.com/spotify/styx/pull/138/files#diff-5b13ae6a3917caa3674163023f79466aR354

How about using e.g. guava `HostAndPort.fromString()` to parse the `STYX_CLI_HOST` env var?

PR #165, opened 2017-05-02
Fix cli host port

danielnorberg:
https://github.com/spotify/styx/issues/164


danielnorberg:
@fabriziodemaria PTAL

honnix:
:+1:

PR #170, opened 2017-05-09
Garbage collect service account keys and secrets

danielnorberg:
Adds service account key and secret garbage collection to ensure that unused service account keys and secrets are eventually removed and not stored forever.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -122,6 +122,7 @@ public class StyxScheduler implements AppInit {
 
   public static final int SCHEDULER_TICK_INTERVAL_SECONDS = 2;
   public static final int TRIGGER_MANAGER_TICK_INTERVAL_SECONDS = 1;
+  public static final int CLEANER_TICK_INTERVAL_SECONDS = 30;
   public static final int RUNTIME_CONFIG_UPDATE_INTERVAL_SECONDS = 5;
   public static final Duration DEFAULT_RETRY_BASE_DELAY = Duration.ofMinutes(3);
   public static final int DEFAULT_RETRY_MAX_EXPONENT = 4;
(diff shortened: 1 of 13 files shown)


danielnorberg:
@TinaRanic @ulzha PTAL

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -157,6 +162,38 @@ public String start(WorkflowInstance workflowInstance, RunSpec runSpec) throws I
     }
   }
 
+  @Override
```
TinaRanic: I'm curious if peoples behaviour is to remove or disable workflows when not using them. Because if the latter is the case, we wouldn't have real benefit..

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -122,6 +122,7 @@
 
   public static final int SCHEDULER_TICK_INTERVAL_SECONDS = 2;
   public static final int TRIGGER_MANAGER_TICK_INTERVAL_SECONDS = 1;
+  public static final int CLEANER_TICK_INTERVAL_SECONDS = 30;
```
TinaRanic: Is there a need to run this frequent?

danielnorberg: Good question. How often do we want to run?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: Good point.

Perhaps we should just always remove all SA keys & secrets (perhaps excluding those used by currently booting pods). Then we wouldn't keep keys for "idle" / disabled workflows and just create them as needed. We'd also achieve key "rotation" for free...

🤔

PR #171, opened 2017-05-09
Aggressively garbage collect service account keys and secrets

danielnorberg:
An alternative to https://github.com/spotify/styx/pull/170. This PR makes styx more aggressively remove keys and secrets that are not in immediate use by any pods. It is assumed that it is fine for workflows to recreate keys and secrets as needed.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
@@ -172,6 +172,11 @@ public class AggregateStorage implements Storage {
   }
 
   @Override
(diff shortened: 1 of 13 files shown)


honnix:
I'm slightly concerned about the performance impact. Cleanup now feels pretty heavy. Pulling all pods from kube is anywhere from being fast.

danielnorberg:
@honnix I'm not too worried about pulling all pods from k8s at our scale.

```
$ kubectl get pods > /dev/null  1.08s user 0.10s system 19% cpu 6.014 total
```

It takes a few seconds but that's fine. We might want to run the cleanup a bit more seldom than every 30s though.

PR #172, opened 2017-05-09
Rotate and garbage collect service account keys and secrets

danielnorberg:
An alternative to https://github.com/spotify/styx/pull/170 and https://github.com/spotify/styx/pull/171 that adds key rotation using epochs.

* The secret epoch is included in the SA secret name.
* SA secrets and keys in past epochs are removed once all pods using them are gone.
* New keys and secrets are created as needed in the current secret epoch when workflows execute.
* The secret epoch is smeared per-SA and increases once every day.


honnix:
Didn't get how bullet 2 works. The secret will be deleted if nobody uses it even in current epoch. I guess I read it wrong.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -127,38 +135,110 @@
...
+      final PodList pods = client.pods().list();
+      final Set<String> activeSecrets = pods.getItems().stream()
+          .flatMap(pod -> pod.getSpec().getVolumes().stream())
+          .map(volume -> volume.getSecret().getSecretName())
+          .collect(Collectors.toSet());
+
+      // Enumerate all service account secrets that are not in active use
+      final SecretList secrets = client.secrets().list();
+      final List<Secret> inactiveServiceAccountSecrets = secrets.getItems().stream()
+          .filter(secret -> secret.getMetadata().getName().startsWith(STYX_WORKFLOW_SA_SECRET_NAME))
```
honnix: Filter by epoch?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -127,38 +135,110 @@
...
   }
 
   @Override
   public String start(WorkflowInstance workflowInstance, RunSpec runSpec) throws IOException {
-    ensureSecrets(workflowInstance, runSpec);
+    final String secretEpoch = secretEpochProvider.get();
+    synchronized (secretMutationLock) {
+      ensureSecrets(workflowInstance, runSpec, secretEpoch);
+      try {
+        final Pod pod = client.pods().create(createPod(workflowInstance, runSpec, secretEpoch));
```
honnix: If we filter by epoch when deleting, do we still need to sync here? We must be very lucky to hit that millisecond. If that did happen, letting it fail isn't too bad.  As payback, we will have better throughput.

styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
```diff
@@ -172,6 +172,11 @@ public void updateNextNaturalTriggerOld(WorkflowId workflowId, Instant instant)
   }
 
   @Override
+  public Map<WorkflowId, Workflow> workflows() throws IOException {
```
honnix: We trigger workflow instance based on whether it has `nextNaturalTrigger` or not. So if it never had, we probably never created SA secret for it anyway. https://github.com/spotify/styx/blob/829da8b159c16fbd5d47982024a44b0616af5cfe/styx-scheduler-service/src/main/java/com/spotify/styx/workflow/WorkflowInitializer.java#L70-L70

I think the case is something wrong in the workflow configuration that we never managed to parse it correctly. Did I missing something?

styx-common/src/main/java/com/spotify/styx/util/GcpUtil.java
```diff
@@ -0,0 +1,42 @@
...
+
+package com.spotify.styx.util;
+
+import com.google.api.client.googleapis.json.GoogleJsonError;
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
+import java.util.Optional;
+
+public class GcpUtil {
+
+  public static final String PERMISSION_DENIED = "PERMISSION_DENIED";
```
honnix: This constant doesn't feel necessary. At least use it in test case? ;)

It's interesting that `403` itself is not enough.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -106,6 +111,7 @@
   private static final int DEFAULT_POLL_PODS_INTERVAL_SECONDS = 60;
   static final String STYX_WORKFLOW_SA_ENV_VARIABLE = "GOOGLE_APPLICATION_CREDENTIALS";
   private static final String STYX_WORKFLOW_SA_ID_ANNOTATION = "styx-wf-sa";
+  private static final String STYX_WORKFLOW_SA_SECRET_EPOCH_ANNOTATION = "styx-wf-sa-epoch";
```
honnix: Be consistent with the actual string? Either `STYX_WORKFLOW_SA_EPOCH_ANNOTATION` or `styx-wf-sa-secret-epoch`.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -127,38 +135,110 @@
...
+        final Pod pod = client.pods().create(createPod(workflowInstance, runSpec, secretEpoch));
+        return pod.getMetadata().getName();
+      } catch (KubernetesClientException kce) {
+        throw new IOException("Failed to create Kubernetes pod", kce);
+      }
+    }
+  }
+
+  @Override
+  public void cleanup() throws IOException {
```
honnix: How about splitting this method into three methods with comprehensive names instead of commenting?
1. Get all active secrets
2. Get all inactive secrets
3. Delete

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -127,38 +135,110 @@
...
+        LOG.warn("[AUDIT] Permission denied when trying to delete unused service account key {}");
+      } else {
+        throw e;
+      }
     }
   }
 
-  private void ensureSecrets(WorkflowInstance workflowInstance, RunSpec runSpec) throws IOException {
-    ensureServiceAccountKeySecret(workflowInstance, runSpec);
+  private String serviceAccountEmail(Secret secret) {
```
honnix: It looks we never call it `...Email` in other places, so maybe just `serviceAccount` or `getServiceAccount`? Or maybe just inline if splitting `cleanup`.

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesDockerRunnerTest.java
```diff
@@ -270,6 +281,116 @@ public void shouldRunIfSecretExists() throws IOException, StateManager.IsClosed
...
+    verify(serviceAccountKeyManager).deleteKey(keyName(SERVICE_ACCOUNT, "p12-key-3"));
+    verify(secrets).delete(secret3);
+  }
+
+  private static Secret fakeServiceAccountKeySecret(String serviceAccount, String epoch, String jsonKeyId, String p12KeyId) {
+    final String jsonKeyName = keyName(serviceAccount, jsonKeyId);
+    final String p12KeyName = keyName(serviceAccount, p12KeyId);
+
+    final ObjectMeta metadata = new ObjectMeta();
+    metadata.setName("styx-wf-sa-keys-" + epoch + "-" + Hashing.sha256().hashString(serviceAccount, UTF_8));
```
honnix: Is it intended that we duplicating `buildSecretName` here? Also those hardcoded strings.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -122,6 +122,7 @@
 
   public static final int SCHEDULER_TICK_INTERVAL_SECONDS = 2;
   public static final int TRIGGER_MANAGER_TICK_INTERVAL_SECONDS = 1;
+  public static final int CLEANER_TICK_INTERVAL_SECONDS = 30;
```
honnix: As we discussed, 30s is too much for polling. I think we could be very happy already to have it run every half an hour or even more. xD

honnix:
A few nitpickings. Major comment is about sync and filter by epoch. I didn't check the other two PRs as  they are included in this one anyway.

honnix:
btw, coveralls.io not working well.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: Right, I guess that will make the key rotation less aggressive and cause less key churn, which seems like a good idea given that the IAM write request quota is not very high out of the box.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: Yeah, perhaps we can eliminate synchronization here as well as in `cleanup()` if we filter by epoch. Especially if we also use the `creationTimestamp` to implement some grace period and not delete keys / secrets that were recently created. 

We'll still want to ensure that concurrently starting workflows do not try to concurrently create keys. I'll restore the previous synchronization.

styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
danielnorberg: Previous incarnations of this PR passed in a list of _all_ workflows to `DockerRunner.cleanup()` in order to not remove SA keys & secrets used by any workflow configuration. But that is no longer needed by this PR so could be reverted.

Might still want to implement it in a separate PR as at least the MetricsStats code seems to rely on `WorkflowCache.workflows()` returning _all_ workflows, not just the enabled and naturally triggering ones.

styx-common/src/main/java/com/spotify/styx/util/GcpUtil.java
danielnorberg: I think 403 could just mean that we failed to properly set up the credentials use by the styx service, whereas I interpret 403 with `PERMISSION_DENIED` as "your credentials are good but you do not have permissions to modify keys for this SA".

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: ok

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesDockerRunnerTest.java
danielnorberg: Yeah, one purpose of the tests is to lock down behavior of code in order to prevent backwards incompatible behavior changes. E.g., if the tests would just invoke `buildSecretName` directly, then a change of the `styx-wf-sa-keys` constant in `KubernetesDockerRunner` would pass the tests a little bit too easily. This way I feel there's a greater chance that we realize that changing that constant would cause styx to leave secrets orphaned in k8s that we'd need to handle etc.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -126,44 +117,54 @@
...
+  @Override
+  public void cleanup() throws IOException {
+    serviceAccountSecretManager.cleanup();
+  }
+
+  private KubernetesSecretSpec ensureSecrets(WorkflowInstance workflowInstance, RunSpec runSpec)
+      throws IOException {
+    return KubernetesSecretSpec.builder()
+        .customSecret(ensureCustomSecret(workflowInstance, runSpec))
+        .serviceAccountSecret(runSpec.serviceAccount().isPresent()
```
honnix: why not `runSpec.serviceAccount().map(...)`?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -329,10 +240,9 @@ static Pod createPod(WorkflowInstance workflowInstance, RunSpec runSpec) {
       // TODO: do we need set this env as default value?
```
honnix: Let's get rid of the TODO since it has been decided.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -348,7 +258,8 @@ static Pod createPod(WorkflowInstance workflowInstance, RunSpec runSpec) {
           .withReadOnly(true)
```
honnix: Align with saVolume? `secretVolume.getName()`, purely being picky.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesGCPServiceAccountSecretManager.java
```diff
@@ -0,0 +1,269 @@
...
+  private final Clock clock;
+
+  private final Object secretMutationLock = new Object() {};
+
+  KubernetesGCPServiceAccountSecretManager(
+      NamespacedKubernetesClient client,
+      ServiceAccountKeyManager serviceAccountKeyManager,
+      EpochProvider epochProvider,
+      Clock clock) {
+    this.client = Objects.requireNonNull(client).inNamespace("default");
```
honnix: Not a big deal, but don't want to have default namespace in two places. Maybe having a setter in this class and letting `KubernetesDockerRunner` decide. Not sure.

danielnorberg: Maybe it should be set by StyxScheduler passing the k8s client to the runner and secret manager?

honnix: Not sure of that. `StyxScheduler` shouldn't know too much about kube itself.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: Can't throw IOException in map lambda.. =/

honnix: Ah, right. I missed that.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: Maybe `ensureServiceAccountKeySecret` should just propagate IOException as a runtime exception?

honnix: Feels odd being different than the rest methods. I don't have good idea. Catch and throw `InvalidExecutionException` or `RuntimeException`? Nah.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -661,7 +661,8 @@ private static NamespacedKubernetesClient getKubernetesClient(Config config, Str
           .withClientKeyData(cluster.getMasterAuth().getClientKey())
           .build();
 
-      return new DefaultKubernetesClient(kubeConfig);
+      return new DefaultKubernetesClient(kubeConfig)
```
honnix: OK. Not bad actually!

honnix:
👍

PR #175, opened 2017-05-10
fix workflows count metric

danielnorberg:


honnix:
This is nice. ^

Test case please... ;)

danielnorberg:
@honnix added a test =)

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -377,7 +377,7 @@ void tickTriggerManager() {
 
   private void warmUpCache(WorkflowCache cache, Storage storage) {
     try {
-      storage.workflowsWithNextNaturalTrigger().keySet().forEach(cache::store);
+      storage.workflows().values().forEach(cache::store);
```
honnix: not sure we really want to cache everything including those not initialised.

honnix:
You didn't change that metric (all workflows)?

honnix:
Sorry that's done via the cache. Now I understand why changing that part. 

👍

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
honnix: disregard my comment.

PR #178, opened 2017-05-11
extend SA key lifetime to 48 hours

danielnorberg:
Even though styx will not delete keys used by running pods, containers might have started e.g. asynchronous dataflow jobs before exiting with exit code 0.

Deleting the keys used by these jobs will cause the jobs to fail.

A minimum key lifetime of 48 hours guarantees that such jobs starting just before the end of the 24h rotation epoch have at least 24h to finish before the key it uses is removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ The [Service Account] email address belonging to a project in [Google Cloud Plat
 
 If the workflow intends to use keys of a [Service Account],
 Styx will create both JSON and p12 keys for the specified `service_account`, rotate keys on daily basis,
-and garbage collect unused keys.
+and garbage collect unused keys older than 48h.
 
 Styx stores the created keys in [Kubernetes Secrets] and mounts them under `/etc/styx-wf-sa-keys/`
 in the container.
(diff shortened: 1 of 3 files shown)


danielnorberg:
@ulzha @honnix @TinaRanic PTAL

honnix:
Do we have the case you described?

danielnorberg:
@honnix I think it might make sense to allow this use case. It also covers the case where a workflow pod would fail to start because it got assigned a wf sa secret that was concurrently deleted by `cleanup()`.

honnix:
```If the workflow intends to use keys of a Service Account, Styx will create both JSON and p12 keys for the specified service_account, rotate keys on daily basis, and garbage collect unused keys.```

This in readme will need to be changed, right? Not daily basis, but older than 48 hours.

**Comment by danielnorberg** [comment]
Keys are still rotated every 24h. But keys would survive at least 48h.
            
```
       0h   24h   48h   72h   96h  ...
key 1  |-----------|
key 2        |-----------|
key 3              |-----------|
...
````         

Maybe say something like `... and garbage collect unused keys older than 48h.` ?

honnix:
That would work.

honnix:
+1

PR #179, opened 2017-05-11
concurrent & cached sa key & secret creation

danielnorberg:


danielnorberg:
@honnix PTAL

honnix:
👍 

Fixed some minor things.

PR #180, opened 2017-05-11
add cleanup method do routing docker runner

danielnorberg:
Also remove default method, force all DockerRunner implementations to implement `cleanup()`.


honnix:
:+1:

PR #184, opened 2017-05-11
make publisher closable

honnix:
@danielnorberg PTAL

internal change coming soon.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/publisher/Publisher.java
+++ b/styx-common/src/main/java/com/spotify/styx/publisher/Publisher.java
@@ -22,12 +22,13 @@ package com.spotify.styx.publisher;
 
 import com.spotify.styx.model.ExecutionDescription;
 import com.spotify.styx.model.WorkflowInstance;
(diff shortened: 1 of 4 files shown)


danielnorberg:
👍

styx-common/src/main/java/com/spotify/styx/publisher/Publisher.java
```diff
@@ -22,12 +22,13 @@
...
 import com.spotify.styx.model.ExecutionDescription;
 import com.spotify.styx.model.WorkflowInstance;
+import java.io.Closeable;
 import java.io.IOException;
 
 /**
  * An interface for publishing events.
  */
-public interface Publisher {
+public interface Publisher extends Closeable {
```
danielnorberg: Use AutoCloseable instead?

honnix: For open source part, `Closeable` feels less surprising. `AutoCloseable` only gives a bit benefit to internal stuff.

PR #185, opened 2017-05-12
pod garbage collection

danielnorberg:
When kube submissions time out (e.g. due to styx restart), the pod created by the submission will never be cleaned up.

This PR implements pod garbage collection by removing unwanted pods when polling.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -287,25 +287,13 @@ class KubernetesDockerRunner implements DockerRunner {
   }
 
   @Override
-  public void cleanup(String executionId) {
-    final Pod pod = client.pods().withName(executionId).get();
-    if (pod != null) {
-      cleanup(pod);
+  public void cleanup(WorkflowInstance workflowInstance, String executionId) {
+    if (!debug.get()) {
+      LOG.info("Cleaning up {} pod: {}", workflowInstance.toKey(), executionId);
+      client.pods().withName(executionId).delete();
+    } else {
+      LOG.info("Keeping {} pod: {}", workflowInstance.toKey(), executionId);
     }
-  }
-
-  private void cleanup(Pod pod) {
-    final String name = pod.getMetadata().getName();
-    readPodWorkflowInstance(pod).ifPresent(
-        workflowInstance -> {
-          if (!debug.get()) {
-            LOG.info("Cleaning up {} pod: {}", workflowInstance.toKey(), name);
-            client.pods().delete(pod);
-          } else {
-            LOG.info("Keeping {} pod: {}", workflowInstance.toKey(), name);
-          }
-        }
-    );
   }
 
   @Override
@@ -394,7 +382,7 @@ class KubernetesDockerRunner implements DockerRunner {
       if (runState.isPresent()) {
         emitPodEvents(Watcher.Action.MODIFIED, pod, runState.get());
       } else {
-        cleanup(pod);
+        cleanup(workflowInstance.get(), pod.getMetadata().getName());
       }
     }
   }
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -368,7 +368,7 @@ class KubernetesDockerRunner implements DockerRunner {
     for (Pod pod : list.getItems()) {
       // Emit events
       logEvent(Watcher.Action.MODIFIED, pod, resourceVersion, true);
-      final Optional<WorkflowInstance> workflowInstance = lookupPodWorkflowInstance(pod);
+      final Optional<WorkflowInstance> workflowInstance = readPodWorkflowInstance(pod);
       if (!workflowInstance.isPresent()) {
         continue;
       }
@@ -382,7 +382,7 @@ class KubernetesDockerRunner implements DockerRunner {
     }
   }
 
-  private Optional<WorkflowInstance> lookupPodWorkflowInstance(Pod pod) {
+  private Optional<WorkflowInstance> readPodWorkflowInstance(Pod pod) {
     final Map<String, String> annotations = pod.getMetadata().getAnnotations();
     final String podName = pod.getMetadata().getName();
     if (!annotations.containsKey(KubernetesDockerRunner.STYX_WORKFLOW_INSTANCE_ANNOTATION)) {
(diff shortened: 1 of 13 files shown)


danielnorberg:
@bergman @fabriziodemaria @honnix @TinaRanic @ulzha PTAL

danielnorberg:
Changed the approach of this PR to integrated with the existing polling instead in order to not delete pods before emitting events for them.

Needs tests.

honnix:
I like this approach much better.

danielnorberg:
Rebased and added tests

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -365,17 +365,24 @@ private synchronized void tryPollPods() {
...
+
+      if (isUnwantedPod && isStyxPod) {
+        LOG.info("Deleting unwanted pod {}", pod.getMetadata().getName());
+        client.pods().delete(pod);
+      }
     }
   }
 
-  private void inspectPod(Watcher.Action action, Pod pod) {
+  private boolean inspectPod(Watcher.Action action, Pod pod) {
```
honnix: Would it feel more natural the other way around: if it is wanted, return true, otherwise false. :)

danielnorberg: Yeah, I went a bit back and forth about this and the reason I ended up having true mean the pod is unwanted is that I don't want a default value (false in the case of boolean) to indicate that a potentially dangerous action should be taken (pod deletion). It's slightly easier for a future code change to make `inspectPod` return false by mistake than it is to return true by mistake.

Still, I'm not too happy with this code, it could do with some refactoring to make things more explicit.

honnix: Maybe splitting it into two methods? The first one simply does those `if` checks and the second one does translation if the first one passes.

Something like:

```
if (isUnwantedStyxPod(...)) {
        LOG.info("Deleting unwanted pod {}", pod.getMetadata().getName());
        client.pods().delete(pod);
} else {
        doTranslation(...);
}
```

honnix: A little bit overhead though...

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -384,23 +392,27 @@ private void inspectPod(Watcher.Action action, Pod pod) {
...
-      return;
+      return Optional.empty();
     }
 
-    final List<Event> events = translate(workflowInstance, runState, action, pod, stats);
+    return Optional.of(runState);
+  }
+
+  private void inspectPod(Watcher.Action action, Pod pod, RunState runState) {
+    final List<Event> events = translate(runState.workflowInstance(), runState, action, pod, stats);
```
honnix: Didn't realise `translate` is so weird. It could get `workflowInstance` by itself.

danielnorberg: I guess it would be nice to avoid duplicating this logic though.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -365,42 +366,63 @@ private synchronized void tryPollPods() {
...
       logEvent(Watcher.Action.MODIFIED, pod, resourceVersion, true);
-      inspectPod(Watcher.Action.MODIFIED, pod);
+      final Optional<WorkflowInstance> workflowInstance = readPodWorkflowInstance(pod);
+      if (!workflowInstance.isPresent()) {
+        continue;
+      }
+      final Optional<RunState> runState = lookupPodRunState(pod, workflowInstance.get());
+      if (runState.isPresent()) {
+        emitPodEvents(Action.MODIFIED, pod, runState.get());
+      } else      {
```
honnix: format

honnix:
👍 with a minor comment.

honnix:
Merge and go sleep. =D

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -366,41 +366,61 @@ private synchronized void tryPollPods() {
...
+      final Optional<WorkflowInstance> workflowInstance = readPodWorkflowInstance(pod);
+      if (!workflowInstance.isPresent()) {
+        continue;
+      }
+      final Optional<RunState> runState = lookupPodRunState(pod, workflowInstance.get());
+      if (runState.isPresent()) {
+        emitPodEvents(Watcher.Action.MODIFIED, pod, runState.get());
+      } else {
+        LOG.info("Deleting unwanted pod {}", pod.getMetadata().getName());
+        client.pods().delete(pod);
```
danielnorberg: @honnix maybe we should respect the `debug` flag?

danielnorberg: And not delete pods if `debug == true` I mean

honnix: Yeah, but they are unwanted anyway. Maybe leave it like this until we really need?

honnix: How bad does it look passing in the debug flag?

danielnorberg: Not too bad I think. As a bonus we now cache the debug flag and do not inject the entire `Storage` interface into the `DockerRunnerHandler`.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -285,7 +288,24 @@ private static EnvVar envVar(String name, String value) {
 
   @Override
   public void cleanup(String executionId) {
-    client.pods().withName(executionId).delete();
+    final Pod pod = client.pods().withName(executionId).get();
+    if (pod != null) {
+      cleanup(pod);
```
honnix: This feels a bit overhead. The pod survived readPodWorkflowInstance check already when processing event.

danielnorberg: This is the method called by `DockerRunnerHandler` and it doesn't currently pass in the workflow instance. But we can change that if we want I guess.

honnix: Yeah, that's when state machine transits which means the pod is a valid one and executionId is the pod name. This however handles some corner cases for example pod gone for no reason or being manually deleted after generating event and before transiting state. It's OK.

honnix: Ah, that's for logging.

honnix:
:+1:

PR #186, opened 2017-05-12
remove lingering SA key when recreating

danielnorberg:
If only one key has been removed, ensure that the remaining key is not
left behind when new keys are created.


danielnorberg:
@honnix @ulzha PTAL

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesGCPServiceAccountSecretManager.java
```diff
@@ -118,30 +118,36 @@ String ensureServiceAccountKeySecret(String workflowId,
   private String getOrCreateServiceAccountKeySecret(
```
honnix: A refactoring for this method? It doesn't fit into one page for my browser any more. ;)

honnix:
A minor comment, otherwise LGTM.

honnix:
👍

PR #187, opened 2017-05-12
remove secrets used by terminated pods

danielnorberg:
Orphaned styx pods (e.g. due to https://github.com/spotify/styx/pull/185) currently causes SA secrets and keys from being removed when rotated.

Mitigate that by excluding terminated pods when enumerating actively used secrets.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -294,6 +294,12 @@
         <version>2.0.0</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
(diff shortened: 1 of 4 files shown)


danielnorberg:
Not sure how codecov concludes this results in less coverage...

danielnorberg:
@honnix @ulzha PTAL

honnix:
👍

PR #188, opened 2017-05-12
handle old resource exception

honnix:
Due to new version of fabric8, the old way catching up resource version
doesn't work now.

@rouzwawi PTAL


rouzwawi:
👍

PR #190, opened 2017-05-17
factor backfill triggering out side of Scheduler

honnix:
It seems better to let it live on its own.

@ulzha @danielnorberg  PTAL


danielnorberg:
👍

PR #191, opened 2017-05-18
allow decoration of workflow resources

danielnorberg:
The `WorkflowResourceDecorator` interface can be implemented to perform arbitrary modification of effective resource use by workflow instances at scheduler dequeue time.

This is effective regardless of trigger type.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowResourceDecorator.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowResourceDecorator.java
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ * 
  *      http://www.apache.org/licenses/LICENSE-2.0
- *
+ * 
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowResourceDecorator.java
```diff
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ *
```
honnix: hahahaha

styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowResourceDecorator.java
```diff
@@ -0,0 +1,35 @@
...
+ * -/-/-
+ */
+
+package com.spotify.styx;
+
+import com.spotify.styx.model.WorkflowConfiguration;
+import com.spotify.styx.model.WorkflowInstance;
+import java.util.Set;
+
+public interface WorkflowResourceDecorator {
```
honnix: Do you think we should do some refactoring for the package? Some of the classes can be moved to `workflow`, e.g. this one, WorkflowCache, InMemWorkflowCache.

Maybe a separated PR.

honnix:
minor comments, otherwise LGTM.

styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowResourceDecorator.java
danielnorberg: *tableflip*

styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowResourceDecorator.java
danielnorberg: Or maybe a `scheduler` package?

danielnorberg: But yes, could be another PR.

honnix:
👍

PR #192, opened 2017-05-18
push down cleanup in order not to delete pod when there is no execution id

honnix:
@danielnorberg a quick fix by pushing down cleanup action

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -378,12 +378,9 @@ class KubernetesDockerRunner implements DockerRunner {
       if (!workflowInstance.isPresent()) {
         continue;
       }
-      final Optional<RunState> runState = lookupPodRunState(pod, workflowInstance.get());
-      if (runState.isPresent()) {
-        emitPodEvents(Watcher.Action.MODIFIED, pod, runState.get());
-      } else {
-        cleanup(workflowInstance.get(), pod.getMetadata().getName());
-      }
+      lookupPodRunState(pod, workflowInstance.get(),
+                        () -> cleanup(workflowInstance.get(), pod.getMetadata().getName()))
+          .ifPresent(runState1 -> emitPodEvents(Watcher.Action.MODIFIED, pod, runState1));
     }
   }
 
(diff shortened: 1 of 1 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -402,11 +399,17 @@ private synchronized void tryPollPods() {
...
+  }
+
+  private Optional<RunState> lookupPodRunState(Pod pod, WorkflowInstance workflowInstance,
+                                               Runnable cleanupAction) {
     final String podName = pod.getMetadata().getName();
 
     final RunState runState = stateManager.get(workflowInstance);
     if (runState == null) {
       LOG.warn("Pod event for unknown or inactive workflow instance {}", workflowInstance);
+      cleanupAction.run();
```
danielnorberg: call cleanup directly instead?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -420,6 +423,7 @@ private synchronized void tryPollPods() {
     if (!podName.equals(executionId)) {
       LOG.warn("Pod event not matching current exec id, current:{} != pod:{}",
           executionId, podName);
+      cleanupAction.run();
```
danielnorberg: same as above

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -378,12 +378,9 @@ private synchronized void tryPollPods() {
       if (!workflowInstance.isPresent()) {
         continue;
       }
-      final Optional<RunState> runState = lookupPodRunState(pod, workflowInstance.get());
-      if (runState.isPresent()) {
-        emitPodEvents(Watcher.Action.MODIFIED, pod, runState.get());
-      } else {
-        cleanup(workflowInstance.get(), pod.getMetadata().getName());
-      }
+      lookupPodRunState(pod, workflowInstance.get(), () -> cleanup(workflowInstance.get(),
```
danielnorberg: this code looks a bit confusing now

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
honnix: this method is shared by both polling and watcher.

danielnorberg:
An alternative solution would be to just have a cleanup grace period. I.e., only remove pod if creationTImestamp more than X minutes ago.

danielnorberg:
👍 but I guess we'll want to revisit this soon =)

honnix:
yeah, we should probably avoid this race condition at all.

honnix:
https://github.com/spotify/styx/blob/a90f9b5b8fa76a1be71ffceb7c72e899a33559ab/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java#L76-L76
😆

danielnorberg:
palm, meet face

PR #193, opened 2017-05-18
when counting currently used resources

honnix:
we need to decorate as well.

this is to fix #191 

@ulzha PTAL


danielnorberg:
👍

danielnorberg:
Would be good to have tests as well

PR #194, opened 2017-05-18
add resource usage metrics

honnix:


honnix:
@danielnorberg unfortunately coverage drops, but doesn't make much sense to cover either.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -554,6 +555,33 @@ private void setupMetrics(
...
     });
 
+    try {
+      storage.resources()
+          .forEach(resource ->
+                       stats.registerResourceCount(
+                           resource.id(),
+                           () -> {
+                             try {
+                               return storage.resource(resource.id())
```
danielnorberg: the level of indentation here shouts out for some refactoring / reformatting

danielnorberg:
Hmm. Does this actually produce resource _usage_ metrics? Looks like it's just reading the configured limits?

honnix:
The histogram shows the usage. In Scheduler class.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
honnix: Tried with reformatting, not good still. Will refactor.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -143,6 +147,8 @@ void tick() {
                 ConcurrentHashMap::new,
                 counting()));
 
+    currentResourceUsage.forEach(stats::resourceUsage);
```
ulzha: A bit disorienting to read statements in a program without any verbs. Not to address in this PR, just complaining.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -554,6 +555,33 @@ private void setupMetrics(
               .count());
     });
 
+    final Function<String, Gauge<Long>> createResourceCountGauge = (resource) -> () -> {
+      try {
+        return storage.resource(resource).map(Resource::concurrency).orElse(0L);
+      } catch (IOException e) {
+        LOG.error("Failed to get resource {}", resource);
```
ulzha: We're repeating this pattern so much all over the place. Hard to imagine what will happen when configuration storage actually intermittently fails with IOExceptions - probably will have very ugly ripple effects.

What do you think of fronting storage with a CachingStorage, that does not throw, runs with the cached values forever if there are IOExceptions in contacting the real storage, and raises alarms if the failure persists for long? That's simpler code and improved operational understanding, via decoupling and modularization of concerns.

honnix: Good comment! Working on it.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
honnix: `stats:updateResourceUsage` and so on would be better. A separated PR to keep this one smaller.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -547,19 +550,30 @@ private void setupMetrics(
...
               .count());
     });
 
-    stats.registerQueuedEvents(queuedEventsCount);
-    stats.registerWorkflowCount("all", allWorkflowsCount);
-    stats.registerWorkflowCount("configured", configuredWorkflowsCount);
-    stats.registerWorkflowCount("enabled", configuredEnabledWorkflowsCount);
-    stats.registerWorkflowCount("docker_termination_logging_enabled",
-                                dockerTerminationLoggingEnabledWorkflowsCount);
+    final Function<String, Gauge<Long>> createResourceCountGauge = (resource) -> {
```
ulzha: Shouldn't call it "resource count" if it's "resource usage" everywhere else. It even has a different meaning.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -209,6 +214,15 @@ public void pullImageError() {
     pullImageErrorMeter.mark();
   }
 
+  @Override
+  public void registerResourceCount(String resource, Gauge<Long> resourceCount) {
+    registry.register(RESOURCE_USAGE.tagged("resource-total", resource), resourceCount);
```
honnix: would `available` mean `configured` or `configured - used`?

ulzha: Or "resource-configured", dunno.

ulzha: Semantically I think these could be just two different "what"s. There's some redundancy in a metric that says `"what": "resource-usage", "resource-used": 101`, and there's some contradiction in a metric that says `"what": "resource-*usage*", "resource-*configured*": 101`.

honnix: That sounds better.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -547,19 +550,30 @@ private void setupMetrics(
...
+    try {
+      storage.resources().forEach(resource -> stats
+          .registerResourceCount(resource.id(), createResourceCountGauge.apply(resource.id())));
+    } catch (IOException e) {
+      LOG.warn("Failed to get resources, skip registering resource count");
+    }
+
+    final Supplier<Optional<Long>> globalConcurrency =
+        new CachedSupplier<>(storage::globalConcurrency, Instant::now);
+    stats.registerResourceCount(Scheduler.GLOBAL_RESOURCE_ID, () ->
```
ulzha: I don't remember if we talked about it, but here it's cumbersome to have the globalConcurrency be a special snowflake among resources. Shall we rather append a default resource like "concurrent-workflows-per-styx-scheduler" to every workflow instance, and make it be a normal resource? Not to be addressed in this PR.

honnix: Feels like a bigger change. I remember there was a concern having it the same as normal resource, could be overwritten or something by any user from API. We had two PRs to choose from and ended up choosing this approach.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/NoopStats.java
```diff
@@ -82,4 +82,12 @@ public void terminationLogInvalid() {
   @Override
   public void exitCodeMismatch() {
   }
+
+  @Override
+  public void registerResourceConfigured(String resource, Gauge<Long> resourceCount) {
```
ulzha: Should be `resourceConfigured`.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/Stats.java
```diff
@@ -57,5 +57,9 @@
 
   void exitCodeMismatch();
 
+  void registerResourceConfigured(String resource, Gauge<Long> resourceCount);
```
ulzha: Should be `resourceConfigured`.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/Stats.java
```diff
@@ -57,5 +57,9 @@
...
+
+  void resourceUsed(String resource, long used);
```
ulzha: I guess it's inconsistent to have the other method start with "register"; either all of them should or none of them should.

ulzha: Or, maybe I'm misunderstanding what's the difference between the `Gauge<>` ones and primitive ones.

honnix: This can be addressed later when adding proper verbs to other methods. It is a bit weird indeed.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -512,28 +514,29 @@ private void setupMetrics(
...
-            .count();
-      } catch (IOException e) {
-        LOG.error("Failed to read enabled status from storage", e);
-        return 0L;
-      }
+        .count());
+
+    final Supplier<Gauge<Long>> configuredEnabledWorkflowsCountGaugeSupplier = () -> {
+      final Supplier<Set<WorkflowId>> enabledWorkflowSupplier =
+          new CachedSupplier<>(storage::enabled, Instant::now);
```
ulzha: Oh, I didn't know we already had `CachedSupplier`. Much better. And it logs the failures, which allows some log based alerting...

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -547,19 +550,30 @@ private void setupMetrics(
...
     });
 
-    stats.registerQueuedEvents(queuedEventsCount);
-    stats.registerWorkflowCount("all", allWorkflowsCount);
-    stats.registerWorkflowCount("configured", configuredWorkflowsCount);
-    stats.registerWorkflowCount("enabled", configuredEnabledWorkflowsCount);
-    stats.registerWorkflowCount("docker_termination_logging_enabled",
-                                dockerTerminationLoggingEnabledWorkflowsCount);
+    final Function<String, Gauge<Long>> createResourceConfiguredGauge = (resource) -> {
+      final Supplier<Optional<Resource>> resourceSupplier =
```
ulzha: Why `Optional<Resource>`? A `CachedSupplier` pattern that prevents all `orElse` should be possible if the resource isn't meant to get removed in the meantime. Or if this is about resources that do get removed, then `0L` is a lie, the metric should stop being reported when the resource is actually gone.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -547,19 +550,30 @@ private void setupMetrics(
...
+          new CachedSupplier<>(() -> storage.resource(resource), Instant::now);
+      return () -> resourceSupplier.get().map(Resource::concurrency).orElse(0L);
+    };
+
+    try {
+      storage.resources().forEach(resource -> stats
+          .registerResourceConfigured(resource.id(), createResourceConfiguredGauge.apply(resource.id())));
+    } catch (IOException e) {
+      LOG.warn("Failed to get resources", e);
```
ulzha: Also this one wants to be prevented with `CachedSupplier`.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -547,19 +551,46 @@ private void setupMetrics(
...
+    }
+
+    final Supplier<Optional<Long>> globalConcurrencySupplier =
```
ulzha: This can be non`Optional` now, right?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -170,6 +177,14 @@ void tick() {
     eligibleInstances.forEach(limitAndDequeue);
   }
 
+  private void updateMetrics(Map<String, Resource> resources,
+                             Map<String, Long> currentResourceUsage) {
+    resources.values().forEach(r -> stats.resourceConfigured(r.id(), r.concurrency()));
+    currentResourceUsage.forEach(stats::resourceUsed);
+    Sets.difference(resources.keySet(), currentResourceUsage.keySet())
+        .forEach(r -> stats.resourceUsed(r, 0));
```
ulzha: :star: 

Just it seems in the previous commits we treated globalConcurrency as well. Not anymore?

honnix: That should have been included in `resources`. https://github.com/spotify/styx/pull/194/files#diff-a473be0c753046d2e7f56a00c3ed5b62L112

ulzha:
LGTM, @honnix says he would refactor in the next PR to improve coverage.

PR #197, opened 2017-05-19
fail fast upon receiving exit code 50

honnix:
For certain use cases, an workflow instance execution would like to
return a well-known exit code (similar as 20 for missing dependency)
to Styx, and that would instruct Styx to stop retrying and fail the
execution immediately.

With something like 50 to mean "permanently failed, don't even retry",
by following this pattern:

https://github.com/spotify/luigi/blob/master/doc/configuration.rst#retcode

@ulzha PTAL

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
@@ -28,6 +28,7 @@ import com.spotify.styx.state.StateManager;
 import com.spotify.styx.util.RetryUtil;
 import java.time.Duration;
 import java.util.Objects;
+import java.util.Optional;
 
 /**
  * A {@link OutputHandler} that manages scheduling generation of {@link Event}s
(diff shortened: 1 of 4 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
```diff
@@ -73,15 +75,28 @@ private void checkRetry(RunState state) {
...
+      final Optional<Integer> exitCode = state.data().lastExit();
+      if (shouldFailFast(exitCode)) {
+        stateManager.receiveIgnoreClosed(Event.stop(workflowInstance));
       } else {
-        delayMillis = retryUtil.calculateDelay(state.data().consecutiveFailures()).toMillis();
+        final long delayMillis;
+        if (shouldRetry(exitCode)) {
+          delayMillis = Duration.ofMinutes(MISSING_DEPS_RETRY_DELAY_MINUTES).toMillis();
+        } else {
+          delayMillis = retryUtil.calculateDelay(state.data().consecutiveFailures()).toMillis();
```
danielnorberg: if shouldRetry == false we... retry? This is a bit confusing. Maybe `shouldRetry` should be named something else?

honnix: ha, indeed, bad naming. `isMissingDependeny()` should be better.

ulzha:
LGTM

fabriziodemaria:
We have the same documentation regarding exit codes here (I suggest we remove it from one of the two places so we don't have to update it twice if we change it again): https://github.com/spotify/styx/blob/dca12c1de2fb3b71a11e709b28bc22f9e1540337/doc/design-overview.md

I also verified that we haven't had a 50 exit code internally before, so this shouldn't override behaviour of existing workflows.

Apart from the docs thing mentioned above it looks good to me +1

honnix:
Very good catch @fabriziodemaria ! I will fix that.

fabriziodemaria:
👍

PR #198, opened 2017-05-22
fix execution id creation & submission race

danielnorberg:
Create the execution id in the prepare state and add it to the run state as part of the submit event. This eliminates the race between k8s submission and execution id persistence that caused the runner to not be able to look up run states for newly started pods.

This also means that styx will not "lose" any pods that were in the middle of being created during a styx restart.

For backwards compatibility, the `executionId` field is kept in the `submitted` event to ensure that we can replay old events and do not lose execution id's when deploying this change.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
@@ -26,6 +26,7 @@ import com.github.sviperll.adt4j.Visitor;
 import com.spotify.styx.state.Message;
 import com.spotify.styx.state.Trigger;
 import java.util.Optional;
+import javax.annotation.Nullable;
 
 /**
  * Generated {@link Event} ADT for all events that can be received by RunState
@@ -37,8 +38,9 @@ public interface EventVisitor<R> {
   R triggerExecution(@Getter WorkflowInstance workflowInstance, Trigger trigger);
   R info(@Getter WorkflowInstance workflowInstance, Message message);
   R dequeue(@Getter WorkflowInstance workflowInstance);
-  R submit(@Getter WorkflowInstance workflowInstance, ExecutionDescription executionDescription);
-  R submitted(@Getter WorkflowInstance workflowInstance, String executionId);
+  R submit(@Getter WorkflowInstance workflowInstance, ExecutionDescription executionDescription,
+      @Nullable String executionId);
+  R submitted(@Getter WorkflowInstance workflowInstance, @Nullable String executionId);
   R started(@Getter WorkflowInstance workflowInstance);
   R terminate(@Getter WorkflowInstance workflowInstance, Optional<Integer> exitCode);
   R runError(@Getter WorkflowInstance workflowInstance, String message);
--- a/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
@@ -117,9 +117,11 @@ class WFIExecutionBuilder {
     }
 
     @Override
-    public Void submit(WorkflowInstance workflowInstance, ExecutionDescription executionDescription) {
+    public Void submit(WorkflowInstance workflowInstance, ExecutionDescription executionDescription,
+        String executionId) {
       currWorkflowInstance = workflowInstance;
       currDockerImg = executionDescription.dockerImage();
+      currExecutionId = executionId;
 
       return null;
     }
@@ -127,7 +129,9 @@ class WFIExecutionBuilder {
     @Override
     public Void submitted(WorkflowInstance workflowInstance, String executionId) {
       currWorkflowInstance = workflowInstance;
-      currExecutionId = executionId;
+      if (currExecutionId == null) {
+        currExecutionId = executionId;
+      }
 
       executionStatusList.add(ExecStatus.create(eventTs, "SUBMITTED"));
       return null;
--- a/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
@@ -129,9 +129,7 @@ class WFIExecutionBuilder {
     @Override
     public Void submitted(WorkflowInstance workflowInstance, String executionId) {
       currWorkflowInstance = workflowInstance;
-      if (currExecutionId == null) {
-        currExecutionId = executionId;
-      }
+      currExecutionId = executionId;
 
       executionStatusList.add(ExecStatus.create(eventTs, "SUBMITTED"));
       return null;
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -74,7 +74,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -138,11 +137,10 @@ class KubernetesDockerRunner implements DockerRunner {
   }
 
   @Override
-  public String start(WorkflowInstance workflowInstance, RunSpec runSpec) throws IOException {
+  public void start(WorkflowInstance workflowInstance, RunSpec runSpec) throws IOException {
     final KubernetesSecretSpec secretSpec = ensureSecrets(workflowInstance, runSpec);
     try {
-      final Pod pod = client.pods().create(createPod(workflowInstance, runSpec, secretSpec));
-      return pod.getMetadata().getName();
+      client.pods().create(createPod(workflowInstance, runSpec, secretSpec));
     } catch (KubernetesClientException kce) {
       throw new IOException("Failed to create Kubernetes pod", kce);
     }
@@ -204,11 +202,9 @@ class KubernetesDockerRunner implements DockerRunner {
         ? runSpec.imageName()
         : runSpec.imageName() + ":latest";
 
-    final String podName = STYX_RUN + "-" + UUID.randomUUID().toString();
-
     final PodBuilder podBuilder = new PodBuilder()
         .withNewMetadata()
-        .withName(podName)
+        .withName(runSpec.executionId())
         .addToAnnotations(STYX_WORKFLOW_INSTANCE_ANNOTATION, workflowInstance.toKey())
         .addToAnnotations(DOCKER_TERMINATION_LOGGING_ANNOTATION,
                           String.valueOf(runSpec.terminationLogging()))
@@ -221,7 +217,7 @@ class KubernetesDockerRunner implements DockerRunner {
         .withName(STYX_RUN)
         .withImage(imageWithTag)
         .withArgs(runSpec.args())
-        .withEnv(buildEnv(workflowInstance, runSpec, podName));
+        .withEnv(buildEnv(workflowInstance, runSpec));
 
     secretSpec.serviceAccountSecret().ifPresent(serviceAccountSecret -> {
       final SecretVolumeSource saVolumeSource = new SecretVolumeSourceBuilder()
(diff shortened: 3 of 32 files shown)


danielnorberg:
@ulzha @TinaRanic @bergman @fabriziodemaria @rouzwawi @honnix PTAL

danielnorberg:
Realized that an issue with this PR is that if `started` is received before `submitted`, then `RunState` does not increment `tries`.

Perhaps we should emit `submitted` _before_ starting the container.

honnix:
It is a a problem. It feels weird to have `submitted` before actual submission though.

danielnorberg:
@honnix even worse, receiving `submitted` after `started` would throw `illegalTransition`.

danielnorberg:
* Currently, receiving `submitted` throws `illegalTransition` unless the current state is `SUBMITTING`.
* The intent seems to be for the `RunState` state machine to receive `submitted` before events from the container like `started`.
* Currently, receiving `started` before `submitted` is prevented by discarding pod events.
* However, it seems that purpose of the watcher is to receive pod events and emit `started` when pods are created. Thus it seems undesirable to discard events for newly created pods just due to a race condition.

By emitting `submitted` before creating the pod we can preserve the `submitted` before `started` ordering. The `submitted` event would thus mean `pod creation has been asynchronously requested`: https://github.com/spotify/styx/pull/200

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
```diff
@@ -71,18 +70,29 @@ public void transitionInto(RunState state) {
         // Perform rate limited submission on a separate thread pool to avoid blocking the caller.
         executor.submit(() -> {
           rateLimiter.acquire();
+
+          final RunSpec runSpec;
           try {
-            final String executionId = dockerRunnerStart(state);
-            // this is racy
```
ulzha: :star:

styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
```diff
@@ -37,8 +38,9 @@
   R triggerExecution(@Getter WorkflowInstance workflowInstance, Trigger trigger);
   R info(@Getter WorkflowInstance workflowInstance, Message message);
   R dequeue(@Getter WorkflowInstance workflowInstance);
-  R submit(@Getter WorkflowInstance workflowInstance, ExecutionDescription executionDescription);
-  R submitted(@Getter WorkflowInstance workflowInstance, String executionId);
+  R submit(@Getter WorkflowInstance workflowInstance, ExecutionDescription executionDescription,
+      @Nullable String executionId);
```
ulzha: In `RunSpec` we add `executionId`, yet in `ExecutionDescription` we don't add it but carry it alongside like this... Intended?

I think `ExecutionDescription` should eventually converge with `RunSpec` to be one value class... Any reason not to? Javadocs missing for both.

styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
```diff
@@ -117,17 +117,21 @@ public Void created(WorkflowInstance workflowInstance, String executionId, Strin
...
+      currExecutionId = executionId;
 
       return null;
     }
 
     @Override
     public Void submitted(WorkflowInstance workflowInstance, String executionId) {
       currWorkflowInstance = workflowInstance;
-      currExecutionId = executionId;
+      if (currExecutionId == null) {
```
ulzha: What's the check for? Should this be an assertion that we don't get changing execution ids, and be logged as error?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -378,9 +374,12 @@ private synchronized void tryPollPods() {
       if (!workflowInstance.isPresent()) {
         continue;
       }
-      lookupPodRunState(pod, workflowInstance.get(), () -> cleanup(workflowInstance.get(),
-                                                                   pod.getMetadata().getName()))
-          .ifPresent(runState -> emitPodEvents(Watcher.Action.MODIFIED, pod, runState));
+      final Optional<RunState> runState = lookupPodRunState(pod, workflowInstance.get());
+      if (runState.isPresent()) {
```
ulzha: Nice.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
```diff
@@ -71,18 +70,29 @@ public void transitionInto(RunState state) {
...
             LOG.error("Unable to start docker procedure.", e);
             stateManager.receiveIgnoreClosed(Event.halt(state.workflowInstance()));
+            return;
+          }
+
+          // Emit submitted event first to guarantee it is observed before events from the pod
+          final Event submitted = Event.submitted(state.workflowInstance(), runSpec.executionId());
+          try {
+            stateManager.receive(submitted);
+          } catch (StateManager.IsClosed isClosed) {
```
ulzha: Kind of unsure why we need to check these. `StateManager` closing means we're exiting, right? Not sending events anymore then is normal operation. It would be fine to exit the event handler thread right there with an unchecked exception.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
```diff
@@ -71,18 +70,29 @@ public void transitionInto(RunState state) {
...
+            LOG.warn("Could not emit 'submitted' event", isClosed);
+            return;
+          }
+
+          try {
+            LOG.info("running:{} image:{} args:{} termination_logging:{}", state.workflowInstance().toKey(),
```
ulzha: Add `toString` to `runSpec`?

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
```diff
@@ -108,30 +118,29 @@ public void transitionInto(RunState state) {
...
+  private RunSpec createRunSpec(RunState state) throws ResourceNotFoundException {
     final Optional<ExecutionDescription> executionDescriptionOpt = state.data().executionDescription();
 
     final ExecutionDescription executionDescription = executionDescriptionOpt.orElseThrow(
         () -> new ResourceNotFoundException("Missing execution description for "
                                             + state.workflowInstance().toKey()));
 
+    final String executionId = state.data().executionId()
+        // Backwards compatibility fallback
+        .orElseGet(ExecutionDescriptionHandler::createExecutionId);
```
ulzha: Hard to think what will happen here... As we replay, we'll work with this real-looking unrelated ID?

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerServiceFixture.java
```diff
@@ -79,7 +81,7 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(StyxSchedulerServiceFixture.class);
 
-  static final String TEST_EXECUTION_ID = "execution_1";
+//  static final String TEST_EXECUTION_ID = "execution_1";
```
ulzha: Commented out code.

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/DockerRunnerHandlerTest.java
```diff
@@ -190,7 +195,7 @@ public void shouldFailIfDockerRunnerRaisesException() throws Exception {
     stateManager.initialize(runState);
     dockerRunnerHandler.transitionInto(runState);
 
-    verify(stateManager, timeout(60_000)).receive(any(Event.class));
+    verify(stateManager, timeout(60_000).times(2)).receive(any(Event.class));
```
ulzha: Where does 2 come from?

ulzha:
Also, please document the updated semantics of the `submitted` event. (At least the SUBMITTING state description in design-overview.md seems wrong now.)

styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
danielnorberg: Both `executionId` and `executionDescription` are (already) fields in `RunState.data()`. Adding `executionId` to `ExecutionDescription` would be redundant. Moving `executionId` from `RunState.data()` to `ExecutionDescription` could be an option, but not sure we would want that. Either way it would be a bigger and backwards incompatible change.

`RunSpec` is not part of `RunState`. It is only a part of the protocol between `DockerRunner` and `DockerRunnerHandler`.

Although `ExecutionDescription` and `RunSpec` are almost identical, I think they might live in separate bounded contexts and that DRY might not apply here. E.g. we would not want to add `trigger` to `ExecutionDescription` just because `DockerRunner` needs it and it is currently in `RunSpec`. Anyway, this might be a discussion worth having outside this PR.

styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
danielnorberg: I now believe this code to be incorrect. To be backwards compatible I think we need to keep unconditionally assigning `executionId` to `currExecutionId`.

danielnorberg: Changing execution id's is actually expected in some circumstances. E.g. a wfi that is retrying.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
danielnorberg: Yeah, these checks have been bothering me as well. Actually, I think Styx should be crash-only and implemented in such a way that sudden restart does not disturb workflow execution. I.e. idempotent at-least-once event transition handling etc.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
danielnorberg: `RunSpec` already has a `toString` as it is an `@AutoValue`. This is not new code though, just factored out from `dockerRunnerStart`, so let's do it in a separate PR.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
danielnorberg: The intent here is to handle existing active WFIs that have already transitioned through `PREPARE`. As they transition through `SUBMITTING`, they will have no `executionId` in their `RunState.data()` and it will be created here by `DockerRunnerHandler` and added to `RunState.data()` through the `submitted` event. This is identical to current behavior.

New WFIs that transition through `PREPARE` will already have an `executionId` in their `RunState.data()` and that will be used.

danielnorberg: As for pure replays, they just go through the transitions defined by `RunState`. Old WFIs will have `executionId`  added to their state by a `submitted` event, new WFIs will have it added by a `submit` event.

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerServiceFixture.java
danielnorberg: oops 👍

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/DockerRunnerHandlerTest.java
danielnorberg: The state manager receives two events:

1. `submitted`
2. `runError`

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
danielnorberg: 🎆

ulzha:
:+1:

PR #200, opened 2017-05-23
emit submitted event before creating pod

danielnorberg:
* Change the semantics of the `submitted` event to mean `workflow instance start has been asynchronously requested`. 

* Guarantee that `submitted` is received before the `started` event from the container.


danielnorberg:
Merged into https://github.com/spotify/styx/pull/198

PR #201, opened 2017-05-23
Rename to StyxCanary for consistency

ulzha:
Using --local-scheduler is not even a recommended mode of operation.


honnix:
We have addressed this in #154. I will double check that everything is covered.

honnix:
@ulzha can I close this PR?

PR #202, opened 2017-05-23
fix "unexpected newer event" spam

danielnorberg:
As it turns out, when reading the backfill run state events, we _do_ expect to concurrently get new events after reading the active states.


danielnorberg:
@honnix @ulzha PTAL

honnix:
Reformatting a bit? Too long lines. Otherwise 👍

danielnorberg:
Hmm. The checkstyle config specifies 120 char long lines so I configured my editor to honor that ¯\_(ツ)_/¯

danielnorberg:
100 chars is a bit easier to read on github. Maybe we should change the checkstyle config?

honnix:
Ah, didn't realise that. So it was OK. Just I need to wrap my eyes. Don't feel like touching that as it's a commonly agreed style.

PR #203, opened 2017-05-23
log a nicer workflow SA permission denied message

danielnorberg:
Current `styx e` cli output for an instance with wf SA permission problems is the raw json from the GoogleJsonResponseException message. That is not very user friendly.

With this change, the cli output will be something like:

```
Permission denied to service account: foo@barz.iam.google.com
```

Not sure if we should also add instructions here about granting e.g. `Key Admin` role on the wf SA for the Styx SA.


honnix:
👍

PR #204, opened 2017-05-23
do not log user error InvalidExecutionException as errors

danielnorberg:


honnix:
you are faster. add a test case covering that path?

honnix:
`-3.33%` is it because you didn't rebase master? I just merged something to master bringing `~3%` more.

honnix:
It feels like this logging in memory killed circleci

danielnorberg:
Yeah =/

danielnorberg:
Ok logback seems to work better.

honnix:
To be honest I started questioning about how useful it is to test log levels given the complexity we need to introduce. https://github.com/spotify/styx/pull/204/files#diff-0286d095fa9d41fc6932d43e2e678cbdR215, those lines are rather hard to understand.

danielnorberg:
@honnix yeah, I agree with that

honnix:
Let's not verify log level in test case then. The change is still valid to push in.

danielnorberg:
Removed the test and rebased.

PR #205, opened 2017-05-23
checkstyle: lower line length to 100

danielnorberg:
Code with 120 char long lines is harder on the eyes in the github PR UI.


danielnorberg:
I guess we'll have to reformat some code to make this pass.

honnix:
not sure we want to go down this path

PR #206, opened 2017-05-23
log user errors as warnings

danielnorberg:
The ongoing quest to reserve `error` for internal system errors.


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesGCPServiceAccountSecretManager.java
```diff
@@ -158,7 +158,7 @@ private void createSecret(String workflowId, String serviceAccount, long epoch,
       jsonKey = keyManager.createJsonKey(serviceAccount);
       p12Key = keyManager.createP12Key(serviceAccount);
     } catch (IOException e) {
-      logger.error("[AUDIT] Failed to create keys for {}", serviceAccount, e);
+      logger.warn("[AUDIT] Failed to create keys for {}", serviceAccount, e);
```
honnix: This is not that clear to get categorised, though. But maybe it's OK as for most cases, it would be user error.

honnix:
👍

PR #207, opened 2017-05-23
warning instead of error

honnix:
in most cases, this is due to workflow being removed or renamed by
users, so a warning feels better.

@fabriziodemaria PTAL


danielnorberg:
👍

PR #208, opened 2017-05-23
log missing workflow as info instead of warning

honnix:
@ulzha PTAL

It's a little bit more complicated than just lowering down log level.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
@@ -101,18 +101,18 @@ public class ExecutionDescriptionHandler implements OutputHandler {
     final Optional<List<String>> dockerArgsOpt = workflow.configuration().dockerArgs();
     if (!dockerArgsOpt.isPresent()) {
       throw new ResourceNotFoundException(format("%s has no docker args, halting %s",
-                                   workflowId, workflowInstance));
+                                                 workflowId, workflowInstance));
     }
 
     final WorkflowState workflowState = storage.workflowState(workflow.id());
 
     final Optional<String> dockerImageOpt = workflowState.dockerImage().isPresent()
-        ? workflowState.dockerImage()
-        : workflow.configuration().dockerImage(); // backwards compatibility
+                                            ? workflowState.dockerImage()
+                                            : workflow.configuration().dockerImage(); // backwards compatibility
 
     if (!dockerImageOpt.isPresent()) {
       throw new ResourceNotFoundException(format("%s has no docker image, halting %s",
-                                   workflowId, workflowInstance));
+                                                 workflowId, workflowInstance));
     }
 
     return ExecutionDescription.create(
@@ -123,6 +123,7 @@ public class ExecutionDescriptionHandler implements OutputHandler {
         workflow.configuration().serviceAccount(),
         workflowState.commitSha());
   }
+
 
   static String createExecutionId() {
     return STYX_RUN + "-" + UUID.randomUUID().toString();
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
@@ -33,6 +33,7 @@ import com.spotify.styx.state.OutputHandler;
 import com.spotify.styx.state.RunState;
 import com.spotify.styx.state.StateManager;
 import com.spotify.styx.storage.Storage;
+import com.spotify.styx.util.MissingRequiredPropertyException;
 import com.spotify.styx.util.ResourceNotFoundException;
 import java.io.IOException;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


honnix:
More context about this change. As we discussed missing workflow in this case is a normal operation than exception, so we would like to handle it in a more peaceful way and log info instead of warning.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
```diff
@@ -61,12 +61,19 @@ public void transitionInto(RunState state) {
...
+            } catch (StateManager.IsClosed isClosed) {
+              LOG.warn("Could not send 'created' event", isClosed);
+            }
+          } else {
+            LOG.info(format("Workflow %s does not exist, halting %s", workflowInstance.workflowId(),
+                            workflowInstance));
+            stateManager.receiveIgnoreClosed(Event.halt(workflowInstance));
           }
         } catch (ResourceNotFoundException e) {
           LOG.warn("Failed to prepare execution description for "
```
danielnorberg: It seems all of the `ResourceNotFoundException`s thrown by getExecDescription are due to user error.

Instead of the above change, how about just changing the logging level here from `warn` to `info`?

danielnorberg:
Btw, please put unrelated commits like 04fc83d in separate PR's, makes it easier to review and merge =)

honnix:
We said user error is warning, right? Missing workflow is not a user error but rather a normal change by the user

honnix:
That commit is to add test case for the change. Should that better be a separated one?

danielnorberg:
@honnix `shouldNotTransitIfStateManagerIsClosed` sounded to me like it was unrelated to changing the logging level

honnix:
@danielnorberg Yeah, that is a bit off indeed. Codecov complains slightly however.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
```diff
@@ -61,12 +61,19 @@ public void transitionInto(RunState state) {
...
       case PREPARE:
         try {
-          final Event createdEvent =
-              Event.submit(state.workflowInstance(), getExecDescription(workflowInstance));
-          try {
-            stateManager.receive(createdEvent);
-          } catch (StateManager.IsClosed isClosed) {
-            LOG.warn("Could not send 'created' event", isClosed);
+          final Optional<Workflow> workflow = storage.workflow(workflowInstance.workflowId());
+          if (workflow.isPresent()) {
```
ulzha: I suspect this is still racy? It doesn't guarantee that ResourceNotFoundException won't occur in handling this state transition, does it? This is only partially a "peaceful way" - a code path is added without removing the objectionable (exception on non-existent workflow) path, so :-1:, for understandability it's better to have just one path even if dirty.

Fully peaceful handling of non-existent workflows would require effectively atomic state handling. (Or serializable, pardon my non-rigorous argument.) I believe we can have that, but that's an even more involved change.

honnix: No it doesn't change any existing logic, so yes still racy. This PR is about logging missing workflow as info instead of warning without trying to fix the racy problem. To fix the bigger problem, another PR? Or maybe we should just close this one as it doesn't bring much actually.

honnix: It is `ResourceNotFoundException ` misleading. We shouldn't throw this exception for missing a field of a resource. I will fix that.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
honnix: We agreed to keep it as warning for now.

styx-common/src/main/java/com/spotify/styx/util/MissingImportantPropertyException.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.spotify.styx.util;
+
+import java.io.IOException;
+
+/**
+ * An exception thrown when an important property is missing in a resource.
+ */
+public class MissingImportantPropertyException extends IOException {
```
danielnorberg: `extends IOException` seems odd

danielnorberg: Also, how `important` does not strictly imply required. How about calling it `MissingRequiredPropertyException` ?

honnix: Yeah, copy pasta.

ulzha:
`MissingRequiredPropertyException` clears things up a lot, yey!

I'm still not seeing this change as obviously "logging missing workflow as info instead of warning". We didn't remove any occurrences of `LOG.warn` as far as I can tell; so to my understanding this does add a new code path (the new `if` is new internal logic - that's what I'm complaining about; agreed that it doesn't change the existing logic that's observable in state machine transitions) and does not prevent the old one (whereby a missing workflow ends up with a warning). Or does it?

honnix:
Yeah, it doesn't change any existing logic, but just logs info instead of warning when missing workflow: https://github.com/spotify/styx/pull/208/files#diff-fa4fe2676fe691f0a9be26a68dc9d16dR96

honnix:
It can be simpler by introducing a new exception type. Maybe we can already log everything as `info` here.

bergman:
👍 (unless @ulzha has objections)

ulzha:
Looks just fine to me now.

PR #210, opened 2017-05-24
remove unused logback config file

danielnorberg:
The logback library is not included in the styx-scheduler-service dependencies.

And instead of logging using logback when testing we might want to use slf4j-test.


honnix:
👍 

How about keeping using NOP to make maven output easier to reason about.

honnix:
Please disregard my comment. slf4j-test is something else than I thought. Good to know!

Another 👍

honnix:
Merge this PR? We have too may open PRs at this moment, it doesn't feel very healthy. 🤒

PR #214, opened 2017-05-26
Bump gcp deps + remove unneeded deps

danielnorberg:


honnix:
👍 
I really hope Google is gonna stablize their groupIds and artifactIds...

PR #215, opened 2017-05-26
WFIExecutionData returns FAILED execution even if no exec id

fabriziodemaria:
Even if a `runError` event is issued before the `submitted` event, hence before the execution id is available, the `WorkflowInstanceExecutionData` now shows the related error for the failed execution.


fabriziodemaria:
PR opened from another fork prevents running of tests/coverage. Closing.

PR #216, opened 2017-05-26
WFIExecutionData reports Execution even if failure happens before "submitted" event

fabriziodemaria:
Even if a `runError` or `halt` events are issued before the `submitted` event, hence before the execution id and/or docker image are made available, the `WorkflowInstanceExecutionData` now reports the failed `Execution` and shows the related error message (in case of runError) or the HALTED status.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
@@ -49,10 +49,11 @@ class WFIExecutionBuilder {
   private Instant eventTs;
 
   private void closeExecution() {
-    if (currExecutionId != null && currDockerImg != null) {
-      Execution execution = Execution.create(currExecutionId, currDockerImg, executionStatusList);
-      executionList.add(execution);
-    }
+    Execution execution = Execution.create(
+        Optional.ofNullable(currExecutionId),
+        Optional.ofNullable(currDockerImg),
+        executionStatusList);
+    executionList.add(execution);
 
     executionStatusList = new ArrayList<>();
     currExecutionId = null;
--- a/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
@@ -148,7 +148,7 @@ class WFIExecutionBuilder {
     public Void terminate(WorkflowInstance workflowInstance, Optional<Integer> exitCode) {
       currWorkflowInstance = workflowInstance;
 
-      String status = exitCode.map(c -> {
+      final String status = exitCode.map(c -> {
         if (c == 0) {
           return "SUCCESS";
         } else if (c == RunState.MISSING_DEPS_EXIT_CODE) {
@@ -158,7 +158,16 @@ class WFIExecutionBuilder {
         }
       }).orElse("FAILED");
 
-      executionStatusList.add(ExecStatus.create(eventTs, status, Optional.empty()));
+      final Optional<String> message;
+      if (status.equals("FAILED")) {
+        message = exitCode
+            .map(c -> Optional.of("Exit code: " + c))
+            .orElseGet(() -> Optional.of("Exit code unknown"));
+      } else {
+        message = Optional.empty();
+      }
+
+      executionStatusList.add(ExecStatus.create(eventTs, status, message));
 
       closeExecution();
       return null;
--- a/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
@@ -49,7 +49,7 @@ class WFIExecutionBuilder {
   private Instant eventTs;
 
   private void closeExecution() {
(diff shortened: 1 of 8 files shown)


fabriziodemaria:
Should we mark this entry in the documentation to be optional somehow?
https://github.com/spotify/styx/blob/a1da3fad99b99041c355bd122daf05258c373451/doc/api.apib#L79

honnix:
How would this impact #198?

ulzha:
Introducing optionals and thus increasing possible states of a system doesn't seem warranted here. The executions that didn't really occur (no ID, failed before `submitted`) should perhaps just not be executions at all in the API.

fabriziodemaria:
> How would this impact #198?

No impact, the modified `WorkflowInstanceExecutionData` is only used in the APIs.

> The executions that didn't really occur (no ID, failed before submitted) should perhaps just not be executions at all in the API.

Indeed this PR slightly changes the concept of a Styx _execution_: I actually tend to like more this version where an execution is an actual attempt to run, including the preparation steps prior to the starting of the container.

ulzha:
My view is that the benefit of reporting all the error messages is better achieved without introducing optional properties and related conditional code (they complicate the model, thus are prone to degrade quality).

:+1: to this in immediate term.

You say "execution is an actual attempt to run" and the class even is documented as "Value representing an execution attempt" - the language is giving many cues that the concept and the class name should be `ExecutionAttempt` not `Execution` ;)

Commit messages should be worded in imperative, rationale e.g. here https://chris.beams.io/posts/git-commit/#imperative. When present tense verbs and the word "now" is used, it tends to leave one confused whether this refers to the state in master or the state after applying the change.

bergman:
should this be a new api version perhaps? it is not backwards compatible, is it?

bergman:
The JSON output wouldn't change much, let's keep the version the same.

styx-common/src/test/java/com/spotify/styx/model/data/deprecated/WorkflowInstanceExecutionDataTest.java
```diff
@@ -136,41 +68,41 @@ public void shouldDeserializeExecutionData() throws Exception {
         + "\"workflow_instance\":{"
           + "\"workflow_id\":{"
             + "\"component_id\":\"component1\","
-            + "\"id\":\"endpoint1\""
+            + "\"endpoint_id\":\"endpoint1\""
```
fabriziodemaria: Most tests in this deprecated class seem to be the same of the non deprecated one. Shouldn't this class test the deprecated API (that uses `endpoint_id`)?

styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
```diff
@@ -157,7 +158,16 @@ public Void terminate(WorkflowInstance workflowInstance, Optional<Integer> exitC
         }
       }).orElse("FAILED");
 
-      executionStatusList.add(ExecStatus.create(eventTs, status));
+      final Optional<String> message;
+      if (status.equals("FAILED")) {
+        message = exitCode
+            .map(c -> Optional.of("Exit code: " + c))
+            .orElseGet(() -> Optional.of("Exit code unknown"));
```
bergman: Shouldn't `orElse(Optional.of(…))` work here?

styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
```diff
@@ -167,7 +177,7 @@ public Void terminate(WorkflowInstance workflowInstance, Optional<Integer> exitC
     public Void runError(WorkflowInstance workflowInstance, String message) {
       currWorkflowInstance = workflowInstance;
 
-      executionStatusList.add(ExecStatus.create(eventTs, message));
+      executionStatusList.add(ExecStatus.create(eventTs, "FAILED", Optional.of(message)));
```
bergman: Can message be null here? Maybe you should use `Optional.ofNullable` in that case.

styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
```diff
@@ -49,10 +49,11 @@
   private Instant eventTs;
 
   private void closeExecution() {
-    if (currExecutionId != null && currDockerImg != null) {
-      Execution execution = Execution.create(currExecutionId, currDockerImg, executionStatusList);
-      executionList.add(execution);
-    }
+    Execution execution = Execution.create(
```
bergman: final or inline this variable

styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
fabriziodemaria: I followed intelliJ suggestions but `orElse` is better 👍

styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
fabriziodemaria: We don't expect it to be null, but it is good to handle that case anyway.

bergman:
👍

PR #218, opened 2017-05-31
fix missing log value

bergman:


bergman:
@danielnorberg you seem to be the last person to touch this code before this, did you just miss including the workflow id?

danielnorberg:
@bergman yeah probably just missed it. Good catch! 👍

PR #220, opened 2017-05-31
log replay events at debug level

danielnorberg:


styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
```diff
@@ -59,7 +59,7 @@ private ReplayEvents() {
       final long lastConsumedEvent = entry.getValue();
       final SettableTime time = new SettableTime();
       if (printLogs) {
-        LOG.info("Replaying {} up to #{}", workflowInstance.toKey(), lastConsumedEvent);
+        LOG.debug("Replaying {} up to #{}", workflowInstance.toKey(), lastConsumedEvent);
```
bergman: does the `if (printLogs)` still make sense if we're logging to debug anyway?

danielnorberg: Not sure. Need to look at the call sites and understand why it `printLogs` was introduced in the first place I guess.

danielnorberg:
@bergman `printLogs` still seems useful in this PR: https://github.com/spotify/styx/pull/225

ulzha:
Merging, codecov failure spurious.

PR #223, opened 2017-06-01
factor Middlewares class to styx-common

honnix:
It can be shared by both styx-api and styx-scheduler.

Also rename the middleware a bit to preare for token validation.

@danielnorberg @bergman PTAL

As a starting point. Feel free to continue or close it.

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/StyxApi.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/StyxApi.java
@@ -20,8 +20,8 @@
 
 package com.spotify.styx;
 
-import static com.spotify.styx.api.Middlewares.auditLogging;
 import static com.spotify.styx.api.Middlewares.clientValidator;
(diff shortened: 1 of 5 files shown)


danielnorberg:
Could token validation and audit logging be separate middlewares?

honnix:
They could. Didn't have time to figure out how to pass in user account to audit logging middleware and it's not good to retrieve that twice.

honnix:
Maybe after token validation we can annotate request object for example setting a new header.

danielnorberg:
@honnix this seems to be the recommended way to decorate the request context: https://github.com/spotify/apollo/blob/master/apollo-api/docs/middleware.md#advanced-creating-custom-context-types

honnix:
Doesn't look very straight forward how to pass info from out handler to inner handler though.

danielnorberg:
Yeah, ideally it would be easy.

honnix:
As I see this PR has been covered by #230. So close.

PR #225, opened 2017-06-02
silence spam for /status/activeStates replay

danielnorberg:
This "error" is expected and harmless while styx is up and getting api requests, so silence it. The `/status/activeStates` handler calls `replayActiveStates` with `printLogs=false`.

The error will still be logged during startup state restore replay, which is desirable it is not expected to happen then. `restoreState` calls `replayActiveStates` with `printLogs=true`.

These are the only two call sites.


danielnorberg:
@honnix @ulzha PTAL

ulzha:
:+1:

PR #228, opened 2017-06-07
re-prioritize docker image specs if docker_image_override is set

ukarlsson:
This is based on the PR #60 from @TinaRanic but since we could not merge that easily without the possibility to break existing pipelines, I have added a `docker_image_override` flag in the workflow definition that toggles this behavior.

This PR changes the order in which a docker image specification takes precedence depending on the `docker_image_override` flag:

if `docker_image_override` is false:
1. docker image as part of state in workflow instance
2. docker image as part of state in component
3. docker image specified in workflow definition (yaml)

if `docker_image_override` is true:
1. docker image specified in workflow definition (yaml)
2. docker image as part of state in workflow instance
3. docker image as part of state in component

The reason for this is that many customers find it more intuitive that a docker image specification in the yaml file should override everything else.

ping @kanterov @Xeago @fabriziodemaria @bergman @honnix

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,10 @@ A unique identifier for the workflow (lower-case-hyphenated). This identifier is
 #### `schedule[].docker_image` **[string]**:
 The Docker image that should be executed.
 
+####  `schedule[].docker_image_override` **boolean**:
+Whether the `docker_image` in the workflow schedule configuration has precedence over the
+`docker_image` in Datastore.
+
 #### `schedule[].docker_args` **[string]**
 The arguments passed to the Docker image.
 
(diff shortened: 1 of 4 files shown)


tkhduracell:
🍰

kanterov:
Do we consider just breaking existing behaviour instead of introducing a flag?

kanterov:
@bergman ping

bergman:
We are working on an alternative solution to this. I will keep you posted on that as soon as we have anything to share!

ulzha:
It looks like we'll be able to clean the design up and have one single way of setting Docker image, that is, the workflow level API endpoint: https://github.com/spotify/styx/pull/243. So the entire concept of "prioritization" in Styx will disappear. It was accidental complexity.

PR #230, opened 2017-06-14
Token logging

rouzwawi:


styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -79,39 +90,94 @@ private Middlewares() {
...
+  private static final GoogleIdTokenVerifier GOOGLE_ID_TOKEN_VERIFIER;
+
+  static {
+    final NetHttpTransport transport;
+    try {
+      transport = GoogleNetHttpTransport.newTrustedTransport();
+    } catch (GeneralSecurityException | IOException e) {
+      throw new RuntimeException(e);
+    }
+    GOOGLE_ID_TOKEN_VERIFIER = new GoogleIdTokenVerifier
```
bergman: Will this client reconnect/reinitialise if it breaks?

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -79,39 +90,94 @@ private Middlewares() {
...
+    }
+  }
+
+  public static <T> Middleware<AsyncHandler<Response<T>>, AsyncHandler<Response<T>>> auditLogger() {
     return innerHandler -> requestContext -> {
       final Request request = requestContext.request();
+      final AuthContext authContext = auth(requestContext);
       if (!"GET".equals(request.method())) {
-        LOG.info("[AUDIT] {} {} with headers {} parameters {} and payload {}",
+        LOG.info("[AUDIT] {} {} from {} with headers {} parameters {} and payload {}",
```
bergman: On the wording, I'd rather make `from x` mean REMOTE_ADDRESS or whatever the name for remote ip is and another word for the user. Perhaps `by foo@bar.com`?

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -79,39 +90,94 @@ private Middlewares() {
...
+        .header(HttpHeaders.AUTHORIZATION)
+        .filter(s -> s.startsWith(BEARER_PREFIX))
+        .map(s -> s.substring(BEARER_PREFIX.length()))
+        .map(Middlewares::verifyIdToken)
+        .filter(Objects::nonNull);
+  }
+
+  private static GoogleIdToken verifyIdToken(final String s) {
+    try {
+      return GOOGLE_ID_TOKEN_VERIFIER.verify(s);
```
bergman: Any gut feeling on the latency added by this?

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
rouzwawi: I believe so. I've never seen code that deals with transport level connections on this layer of their client libs. Seems it's all handled within the `transport` implementation.

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -79,39 +90,94 @@ private Middlewares() {
...
+  public static AuthContext auth(RequestContext requestContext) {
+    return () -> requestContext.request()
+        .header(HttpHeaders.AUTHORIZATION)
+        .filter(s -> s.startsWith(BEARER_PREFIX))
+        .map(s -> s.substring(BEARER_PREFIX.length()))
+        .map(Middlewares::verifyIdToken)
+        .filter(Objects::nonNull);
+  }
+
+  private static GoogleIdToken verifyIdToken(final String s) {
```
bergman: This codebase doesn't seem to use the `final` on parameters.

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
rouzwawi: Looking at the code, it seems to only do token parsing and local validation. One part of the code path can lead to refreshing the public keys used to verify the token signature. But that is only hit periodically when the keys need to be refreshed.

So in most cases it should be insignificant delay, and occationally a request to Google.

fabriziodemaria: Should we log cases of ID tokens failing the validation step?

bergman: What does that mean for key revocations? Is it acceptable to have offline/local verification only? Perhaps it could be enforced on the mutating (POST, DELETE etc) APIs but local validation could be fine on non-mutating (HEAD, GET)?

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -79,39 +90,94 @@ private Middlewares() {
...
+  public static <T> Middleware<Requested<Authenticated<T>>, AsyncHandler<Response<ByteString>>> authed() {
+    return ar -> jsonAsync().apply(requestContext -> {
+      final T payload = ar
+          .apply(requestContext)
+          .apply(auth(requestContext));
+      return completedFuture(Response.forPayload(payload));
+    });
+  }
+
+  public static <T> Middleware<AsyncHandler<Response<T>>, AsyncHandler<Response<T>>> authValidator() {
```
rouzwawi: Yes

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -79,39 +90,94 @@ private Middlewares() {
...
+  }
+
+  public interface AuthContext {
+    Optional<GoogleIdToken> user();
+  }
+
+  public interface Authenticated<T> extends Function<AuthContext, T> {}
+  public interface Requested<T> extends Function<RequestContext, T> {}
+
+  public static <T> Middleware<Requested<Authenticated<T>>, AsyncHandler<Response<ByteString>>> authed() {
```
rouzwawi: This is also for future use. Moved.

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -79,39 +90,94 @@ private Middlewares() {
...
-        // noinspection unchecked
-        return (CompletionStage<Response<ByteString>>) innerHandler.invoke(requestContext);
+        return innerHandler.invoke(requestContext);
       }
     };
   }
 
-  public static Middleware<AsyncHandler<? extends Response<?>>,
-      AsyncHandler<? extends Response<ByteString>>> auditLogging() {
+  private static final GoogleIdTokenVerifier GOOGLE_ID_TOKEN_VERIFIER;
```
rouzwawi: done

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -79,39 +90,94 @@ private Middlewares() {
...
+    return h -> rc -> {
+      if (!auth(rc).user().isPresent()) {
+        return completedFuture(Response.forStatus(Status.UNAUTHORIZED));
+      }
+
+      return h.invoke(rc);
+    };
+  }
+
+  public static AuthContext auth(RequestContext requestContext) {
```
rouzwawi: done

styx-api-service/src/main/java/com/spotify/styx/api/deprecated/WorkflowResource.java
```diff
@@ -52,11 +54,11 @@ public WorkflowResource(com.spotify.styx.api.WorkflowResource workflowResource)
     this.workflowResource = Objects.requireNonNull(workflowResource);
   }
 
-  public Stream<? extends Route<? extends AsyncHandler<? extends Response<ByteString>>>> routes() {
+  public Stream<Route<AsyncHandler<Response<ByteString>>>> routes() {
     final List<Route<AsyncHandler<Response<ByteString>>>> v0Routes = Arrays.asList(
         Route.with(
-            json(), "GET", BASE + "/<cid>/<eid>",
-            rc -> workflow(arg("cid", rc), arg("eid", rc))),
+            authed(), "GET", BASE + "/<cid>/<eid>",
```
rouzwawi: removed

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -155,7 +157,7 @@ public WorkflowResource(Storage storage) {
     return Response.forPayload(patchState);
   }
 
-  public Response<Workflow> workflow(String componentId, String id) {
+  public Response<Workflow> workflow(AuthContext authContext, String componentId, String id) {
```
rouzwawi: removed

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -137,6 +139,12 @@ private static GoogleIdToken verifyIdToken(String s) {
     return innerHandler -> requestContext -> {
       final Request request = requestContext.request();
       final AuthContext authContext = auth(requestContext);
+
+      if (!authContext.isValid()) {
+        return completedFuture(Response.forStatus(
+            Status.BAD_REQUEST.withReasonPhrase("Invalid Authorization token")));
```
bergman: Wouldn't 401 be more descriptive?

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -151,17 +159,27 @@ private static GoogleIdToken verifyIdToken(String s) {
     };
   }
 
+  @AutoMatter
   public interface AuthContext {
+    boolean hasAuthHeader();
+    boolean isValid();
```
bergman: When auth is enforced, this doesn't seem that useful. I was thinking we could just throw an exception and handle that somewhere else… but maybe that's harder.

PR #231, opened 2017-06-16
Add some more env variables

bergman:
- commit sha
- docker image
- docker args
- service account

commit sha wasn't available in RunSpec so had to add it. Might be worthwhile to think about skipping the RunSpec and just pass the ExecutionDescription directly.


rouzwawi:
👍

honnix:
update readme?

labianchin:
Super nice! Thanks!

PR #233, opened 2017-06-21
cli: add Google ID token authentication

ulzha:
With this, `styx` CLI tool will require a logged in gcloud user or an activated service account.

Edit: no - if the credential is not found then CLI will attempt to call API without authentication as previously.

### DIFF
--- a/styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
+++ b/styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
@@ -1,0 +1,109 @@
+/*-
+ * -\-\-
+ * Spotify Styx API Client
+ * --
+ * Copyright (C) 2016 - 2017 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.styx.client.auth;
+
+import com.google.api.client.auth.oauth2.RefreshTokenRequest;
+import com.google.api.client.auth.oauth2.TokenRequest;
+import com.google.api.client.auth.oauth2.TokenResponse;
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+import com.google.api.client.googleapis.util.Utils;
+import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.webtoken.JsonWebSignature;
+import com.google.api.client.json.webtoken.JsonWebSignature.Header;
+import com.google.api.client.json.webtoken.JsonWebToken.Payload;
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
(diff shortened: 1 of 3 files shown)


fabriziodemaria:
What is the reasoning behind having the CLI fail in case of missing default Google credentials? Should authentication be enforced at the server side instead, according to API versioning/deprecation?

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
```diff
@@ -0,0 +1,109 @@
...
+import com.google.api.client.json.webtoken.JsonWebToken.Payload;
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+public class GoogleIdTokenAuth {
+  private final HttpTransport httpTransport;
+  private static final JsonFactory JSON_FACTORY = Utils.getDefaultJsonFactory();
+
+  public GoogleIdTokenAuth(HttpTransport httpTransport) {
```
bergman: Is this one needed for a test? Maybe make it package-private then?

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
```diff
@@ -0,0 +1,109 @@
...
+  public String getToken() throws IOException, GeneralSecurityException {
+    GoogleCredential credential = GoogleCredential.getApplicationDefault();
+    if (credential.getServiceAccountId() != null) {
+      // is a service account
+      if (credential.createScopedRequired()) {
+        credential = credential.createScoped(ImmutableList.of(
+            "https://www.googleapis.com/auth/cloud-platform",
+            "email"));
+      }
+      return getServiceAccountToken(credential, "https://styx.spotify.net");
```
bergman: Internal host

ulzha: :disappointed:

ulzha:
@fabriziodemaria no reasoning, I just hadn't thought about supporting the unauthorized case. How about now?

Also, ninja Codecov coverage.

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
```diff
@@ -0,0 +1,109 @@
...
+    this(Utils.getDefaultTransport());
+  }
+
+  public String getToken(String targetAudience) throws IOException, GeneralSecurityException {
+    GoogleCredential credential = GoogleCredential.getApplicationDefault();
+    if (credential.getServiceAccountId() != null) {
+      // is a service account
+      if (credential.createScopedRequired()) {
+        credential = credential.createScoped(ImmutableList.of(
+            "https://www.googleapis.com/auth/cloud-platform",
```
bergman: this list could be a static variable since it never changes

ulzha: I find that static variables often end up less readable than in-place literals. Also in this case. When "...Scoped" in the immediate context explains what these strings are, why would I try to give them yet another name.

ulzha:
Tested that this sends the Authorization header (gets logged on the server side) both for a logged in user as well as for an activated service account.

fabriziodemaria:
👍

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
```diff
@@ -0,0 +1,109 @@
...
+
+  GoogleIdTokenAuth(HttpTransport httpTransport) {
+    this.httpTransport = httpTransport;
+  }
+
+  public GoogleIdTokenAuth() {
+    this(Utils.getDefaultTransport());
+  }
+
+  public String getToken(String targetAudience) throws IOException, GeneralSecurityException {
```
danielnorberg: Seems the GCE metadata server hands out id tokens. Might be worth supporting. 

https://cloud.google.com/compute/docs/instances/verifying-instance-identity

PR #234, opened 2017-07-03
update readme

bergman:


README.md
```diff
@@ -145,7 +144,8 @@ Example arguments for the supported schedule values:
 ```
 
 #### `schedule[].schedule` **[string]**
-How often the workflow should be triggered.
+How often the workflow should be triggered and what the `{}` placeholder will be replaced with in
```
honnix: How about "... and **how** the `{}` placeholder will be replaced"?

PR #237, opened 2017-07-06
Increase key rotation epoch to weekly

bergman:
GCP Service Accounts have a low limit of number of keys. Having the rotation epoch greater than the timeout of "running" state (24 hours) means there is no chance of having three epochs of keys active at the same time. This means no more warnings about number of keys being too high.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesGCPServiceAccountSecretManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesGCPServiceAccountSecretManager.java
@@ -64,12 +64,15 @@ class KubernetesGCPServiceAccountSecretManager {
   private static final String STYX_WORKFLOW_SA_JSON_KEY = "styx-wf-sa.json";
   private static final String STYX_WORKFLOW_SA_P12_KEY = "styx-wf-sa.p12";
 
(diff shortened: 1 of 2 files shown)


bergman:
Setting it to weekly is just an arbitrary choice greater than the timeout. I would be fine with any number greater than 24 hours. Perhaps 48 hours is a better choice.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesGCPServiceAccountSecretManager.java
```diff
@@ -64,12 +64,15 @@
   private static final String STYX_WORKFLOW_SA_JSON_KEY = "styx-wf-sa.json";
   private static final String STYX_WORKFLOW_SA_P12_KEY = "styx-wf-sa.p12";
 
-  private static final EpochProvider DEFAULT_SECRET_EPOCH_PROVIDER =
-      KubernetesGCPServiceAccountSecretManager::smearedDailyEpoch;
+  static final Duration DEFAULT_SECRET_EPOCH_PERIOD = Duration.ofDays(7);
```
honnix: I would prefer keeping the visibility and using magic numbers in unit test for sanity check. Minor issue though.

honnix:
👍

PR #238, opened 2017-07-06
Handle max number of keys created nicer

bergman:
This kind of exception isn't handled properly hence if it happens users will see Google related exception as event description, which is not so friendly.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/GcpUtil.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/GcpUtil.java
@@ -38,8 +38,21 @@ public class GcpUtil {
   }
 
   public static boolean isPermissionDenied(GoogleJsonError error) {
-    return Optional.ofNullable(error.get("status"))
-        .map("PERMISSION_DENIED"::equals)
+    return "PERMISSION_DENIED".equals(error.get("status"));
+  }
+
+  public static boolean isResourceExhausted(Throwable t) {
+    return t instanceof GoogleJsonResponseException
+           && isResourceExhausted((GoogleJsonResponseException) t);
+  }
+
+  public static boolean isResourceExhausted(GoogleJsonResponseException e) {
+    return e.getStatusCode() == 429 && Optional.ofNullable(e.getDetails())
+        .map(GcpUtil::isResourceExhausted)
         .orElse(false);
+  }
+
+  public static boolean isResourceExhausted(GoogleJsonError error) {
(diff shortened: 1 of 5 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesGCPServiceAccountSecretManager.java
```diff
@@ -107,12 +108,14 @@ String ensureServiceAccountKeySecret(String workflowId,
     try {
       return serviceAccountSecretCache.get(serviceAccount, () ->
           getOrCreateSecret(workflowId, serviceAccount, epoch, secretName));
-    } catch (Exception e) {
+    } catch (ExecutionException | UncheckedExecutionException e) {
```
honnix: Why do we need to be specific here?

ulzha:
Missing description, and the test doesn't explain in what way exactly it becomes "nicer"...

honnix:
I guess it will be nicer for user, otherwise `styx e ...` will show a long Google exception message which we'd better hide.

Somewhere here: https://github.com/spotify/styx/blob/7b9b0d934069afb24b24b3baf4f2a54226e5168d/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java#L105-L105

styx-common/src/main/java/com/spotify/styx/util/GcpUtil.java
```diff
@@ -38,8 +38,21 @@ public static boolean isPermissionDenied(GoogleJsonResponseException e) {
...
-        .map("PERMISSION_DENIED"::equals)
+    return "PERMISSION_DENIED".equals(error.get("status"));
+  }
+
+  public static boolean isResourceExhausted(Throwable t) {
+    return t instanceof GoogleJsonResponseException
+           && isResourceExhausted((GoogleJsonResponseException) t);
+  }
+
+  public static boolean isResourceExhausted(GoogleJsonResponseException e) {
```
fabriziodemaria: A few of these methods can be private

fabriziodemaria:
Smaller comment, looks good to me.

styx-common/src/main/java/com/spotify/styx/util/GcpUtil.java
honnix: I think the idea was to have them public as help methods (_might be used later but not now_ stuff).

honnix:
@ulzha @fabriziodemaria are we good to go?

PR #240, opened 2017-07-13
Enable V3 api

honnix:
Keep backward compatibility for older api versions while attaching
authValidator middleware to V3.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
+++ b/styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
@@ -161,15 +161,16 @@ public final class Middlewares {
       final Request request = requestContext.request();
 
       if (!"GET".equals(request.method())) {
-        final AuthContext authContext = auth(requestContext);
         LOG.info("[AUDIT] {} {} by {} with headers {} parameters {} and payload {}",
-            request.method(),
-            request.uri(),
-            authContext.user().map(idToken -> idToken.getPayload().getEmail()).orElse("anonymous"),
-            filterHeaders(request.headers()),
-            request.parameters(),
-            request.payload().map(ByteString::utf8).orElse("")
-                .replaceAll("\n", " "));
+                 request.method(),
+                 request.uri(),
+                 auth(requestContext).user().map(idToken -> idToken.getPayload()
+                     .getEmail())
+                     .orElse("anonymous"),
+                 filterHeaders(request.headers()),
+                 request.parameters(),
+                 request.payload().map(ByteString::utf8).orElse("")
+                     .replaceAll("\n", " "));
       }
       return innerHandler.invoke(requestContext);
     };
--- a/styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
+++ b/styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
@@ -159,9 +159,9 @@ public final class Middlewares {
   public static <T> Middleware<AsyncHandler<Response<T>>, AsyncHandler<Response<T>>> auditLogger() {
     return innerHandler -> requestContext -> {
       final Request request = requestContext.request();
-      final AuthContext authContext = auth(requestContext);
 
       if (!"GET".equals(request.method())) {
+        final AuthContext authContext = auth(requestContext);
         LOG.info("[AUDIT] {} {} by {} with headers {} parameters {} and payload {}",
             request.method(),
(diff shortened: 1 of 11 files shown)


honnix:
@axelri @NatashaL @ulzha PTAL

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -159,9 +159,9 @@ private static GoogleIdToken verifyIdToken(String s) {
   public static <T> Middleware<AsyncHandler<Response<T>>, AsyncHandler<Response<T>>> auditLogger() {
     return innerHandler -> requestContext -> {
       final Request request = requestContext.request();
-      final AuthContext authContext = auth(requestContext);
 
       if (!"GET".equals(request.method())) {
+        final AuthContext authContext = auth(requestContext);
```
ulzha: What does this do? Doesn't seem used.

NatashaL: It's used, see line 168

honnix: Yeah. Can probably just be inline.

ulzha: Ok, I was just not seeing the obvious and too lazy to press Alt+Shift+7

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -159,17 +159,18 @@ private static GoogleIdToken verifyIdToken(String s) {
...
-            request.method(),
-            request.uri(),
-            authContext.user().map(idToken -> idToken.getPayload().getEmail()).orElse("anonymous"),
-            filterHeaders(request.headers()),
-            request.parameters(),
-            request.payload().map(ByteString::utf8).orElse("")
-                .replaceAll("\n", " "));
+                 request.method(),
+                 request.uri(),
+                 auth(requestContext).user().map(idToken -> idToken.getPayload()
```
ulzha: Hm, we're inconsistent with this comment https://github.com/spotify/styx/pull/240/files#diff-7f72d8b0fc3e9e2d8512aa93f0d1e3c4R221

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -231,10 +232,14 @@ private static AuthContext auth(RequestContext requestContext) {
...
 
-  private static <T> Middleware<AsyncHandler<Response<T>>, AsyncHandler<Response<T>>> authValidator() {
+  public static <T> Middleware<AsyncHandler<Response<T>>, AsyncHandler<Response<T>>> authValidator() {
     return h -> rc -> {
-      // ensure auth context is valid (throws otherwise)
-      auth(rc);
+      if (!"GET".equals(rc.request().method())) {
+        if (!auth(rc).user().isPresent()) {
+          return completedFuture(
+              Response.forStatus(Status.UNAUTHORIZED.withReasonPhrase("Unauthorized access")));
```
axelri: Is there any convention on when to return a completedFuture (like this case) or throw a ResponseException (like in the called auth function)?

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
ulzha: As far as I understand the idea is to give a distinguishing type to the authenticated handlers, so they can't accidentally be invoked before authentication?

honnix: Not sure of the whole story. I thought that middleware shall be used if the actual handler needs to access `user`, and so far we don't have that case.

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
honnix: Not really. Normally returning a future is needed, and in this case it might be just to make `auth` easier to call and handle.

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
honnix: That middleware currently is not useful unless in the future we may have a handler needs to access `user`.

ulzha:
:+1:

PR #241, opened 2017-07-14
use v3

honnix:
depends on #240


honnix:
covered by #240 to make it simpler to roll out.

PR #245, opened 2017-07-24
having a flag to disable update from schedule source

honnix:
@axelri @ulzha

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/model/Workflow.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/Workflow.java
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
+import com.spotify.styx.api.Api;
 import java.net.URI;
 import java.util.Optional;
 
@@ -41,6 +42,11 @@ public abstract class Workflow {
   @JsonProperty
   public abstract Optional<URI> componentUri();
 
+  // this gives a hint that the workflow registration/modification came from
+  // a certain version of Styx API instead of any schedule source
+  @JsonProperty("__from_api")
+  public abstract Optional<Api.Version> fromApi();
+
   @JsonProperty
   public abstract WorkflowConfiguration configuration();
 
@@ -52,8 +58,12 @@ public abstract class Workflow {
   public static Workflow create(
       @JsonProperty("component_id") String componentId,
       @JsonProperty("component_uri") Optional<URI> componentUri,
+      @JsonProperty("__from_api") Optional<Api.Version> fromApi,
       @JsonProperty("configuration") WorkflowConfiguration configuration) {
-    return new AutoValue_Workflow(componentId, configuration.id(), componentUri,
+    return new AutoValue_Workflow(componentId,
+                                  configuration.id(),
+                                  componentUri,
+                                  fromApi,
                                   configuration);
   }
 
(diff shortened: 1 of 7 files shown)


honnix:
Not super happy with https://github.com/spotify/styx/pull/245/files#diff-a660bb778e38936e1c4f442324f41e64R140

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -570,6 +571,15 @@ private void setupMetrics(
           workflow.id(),
           () -> stateManager.getActiveStatesCount(workflow.id()));
 
+      final Optional<Workflow> existingWorkflow = cache.workflow(workflow.id());
+      if (existingWorkflow.isPresent()) {
+        // if we have ever registered/updated the workflow from API,
+        // we ignore any subsequent updates via schedule source
+        if (existingWorkflow.get().fromApi().isPresent() && !workflow.fromApi().isPresent()) {
```
axelri: This line reads a bit hard if you're not familiar with the history of this field. What you want to say is, "was this call made through the new API?", but I would read this as "was this call made through an unknown API version?". Ideally we would want to easily compare API versions, both to clarify intent and for future-proofing. For example:

```java
if (isHigher(existingWorkflow.get().fromApi(), workflow.fromApi())) {
  return;
}
```

styx-common/src/main/java/com/spotify/styx/model/Workflow.java
```diff
@@ -41,6 +41,11 @@
   @JsonProperty
   public abstract Optional<URI> componentUri();
 
+  // this gives a hint that the workflow registration/modification came from Styx API
+  // instead of any schedule source
+  @JsonProperty("__from_api")
+  public abstract Optional<String> fromApi();
```
axelri: String seems like a suboptimal type for the reasons I laid out in my other comment. But maybe it's more convenient for serialization compared to for example Enums.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
honnix: yeah, better to have a properly named method.

styx-common/src/main/java/com/spotify/styx/model/Workflow.java
honnix: good point, as API is part of styx-command anyway, doesn't make sense to go for string.

axelri:
👍

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -570,11 +571,28 @@ private void setupMetrics(
           workflow.id(),
           () -> stateManager.getActiveStatesCount(workflow.id()));
 
+      final Optional<Workflow> existingWorkflow = cache.workflow(workflow.id());
+      if (existingWorkflow.isPresent()) {
+        if (!isGreaterOrEqualApiVersion(workflow, existingWorkflow.get())) {
+          return;
```
ulzha: Silently returning with a 200 OK? I think that's lying.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -570,11 +571,28 @@ private void setupMetrics(
...
+        }
+      }
+
       cache.store(workflow);
       workflowInitializer.inspectChange(workflow);
     };
   }
 
+  // if we have ever registered/updated the workflow via API,
+  // we ignore any subsequent updates via schedule source or lower API version
```
ulzha: This describes "what" but doesn't describe "why". Harrumph. "We ignore _in order to_ enable gradual rollout of workflow configuration API endpoint, while still supporting the polling schedule source for users that haven't migrated".

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
honnix: Indeed. Shall we address this later when validating a workflow, e.g. cron syntax? We would need to introduce a new exception and throw it out.

I will anyway put FIXME here.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
honnix: Sure.

PR #246, opened 2017-07-24
choose a more random port

honnix:
8080 is too well-known not to collide with others.

@axelri


axelri:
Seems fine! I could not find any other instances of the old port in test classes.

PR #247, opened 2017-07-24
something better than Optional.empty

honnix:


honnix:
@axelri PTAL

btw, this coverage check will never succeed since we have no test case :(

ulzha:
:+1:

axelri:
Looks good to me

PR #248, opened 2017-07-25
Guard triggering of each workflow

honnix:
The triggering process may fail with an unchecked exception and hence it
can affect rest of the workflows in the map.

Remediation for INCIDENT-6128.

@ulzha @fabriziodemaria PTAL

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
@@ -20,6 +20,7 @@
 
 package com.spotify.styx;
 
+import static com.spotify.styx.util.GuardedRunnable.guard;
 import static com.spotify.styx.util.TimeUtil.nextInstant;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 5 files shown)


fabriziodemaria:
LGTM

honnix:
That is a a very good place to do some validation, for both schedule source and api.

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -71,30 +72,30 @@ void tick() {
       }
     } catch (IOException e) {
       LOG.warn("Couldn't fetch global enabled status, skipping this run.");
-      return ;
+      return;
     }
 
-    final Map<Workflow, TriggerInstantSpec> map;
-    final Set<WorkflowId> enabled;
+    final Map<Workflow, TriggerInstantSpec> canBeTriggeredWorkflows;
```
ulzha: :star:

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -71,30 +72,30 @@ void tick() {
...
-    map.entrySet().forEach(entry -> {
-      final Workflow workflow = entry.getKey();
-      final Schedule schedule = workflow.configuration().schedule();
-      final TriggerInstantSpec instantSpec = entry.getValue();
-
-      if (now.isBefore(instantSpec.offsetInstant())) {
-        return;
-      }
+    canBeTriggeredWorkflows.entrySet().stream()
+        .filter(entry -> !now.isBefore(entry.getValue().offsetInstant()))
```
ulzha: Hard to read. Can we avoid negated logic, turning it around?

honnix: yeah, thought about this. that would slightly change the original logic if `it is now exactly`.

honnix: That wouldn't matter. Let's change it.

honnix: It was amazing we have test cases replying on exact time matching. Good catch of that!

ulzha: Ah, somewhat unfortunate, didn't intend to change the logic. Anyway fine.

ulzha:
:+1:

PR #249, opened 2017-07-25
patch at component level is gone from v3

honnix:
historically this has been causing more troubles than benefit.

@ulzha PTAL


ulzha:
:+1:

PR #251, opened 2017-07-26
V3 api description

honnix:
component level state PATCH only exists for V2 and lower, will be
deprecated soon.

@axelri PTAL

After being merged, we will generate html version to gh-pages branch.


axelri:
👍

PR #255, opened 2017-08-01
remove workflow mutating APIs from v3

bergman:
The idea is that these things should be handled in the ScheduleSource implementations. Styx should treat ScheduleSources as truth and not own the workflow definitions. Overrides through the various patch calls are still possible (for now).


honnix:
Please kill those endpoints in api blueprint as well.

Do we want to delete though from schedule api as well? Or maybe a separated PR.

bergman:
Good point @honnix , removing that too

doc/api.apib
```diff
@@ -68,18 +68,6 @@ Resources related to the Workflows registered in Styx.
...
-
-+ Parameters
-    + version: `v3` (enum[string]) - API version
-        + Members
-            + `v3`
-
-+ Response 200
-+ Response 404
-
 ## Workflows [/{version}/workflows/{component_id}]
```
honnix: The whole section can be gone I think.

honnix:
Can we hold on merging it for a while as we discussed yesterday in order to have a more peaceful migration?

honnix:
+1 let’s ship it?

PR #256, opened 2017-08-03
Remove docker image requirement

axelri:
Users should be able to create workflows without specifying a docker image.


axelri:
PTAL @honnix

honnix:
👍 when it is green

PR #257, opened 2017-08-03
treat iam 403s as 404

bergman:
It seems Google's IAM API is returning 403's even though the service account used to make the call has ServiceAccountKeyAdmin permission on the service account. Not sure if this is new but haven't seen this problem before.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
@@ -64,7 +64,8 @@ public class ServiceAccountKeyManager {
           .execute();
       return true;
     } catch (GoogleJsonResponseException e) {
-      if (e.getStatusCode() == 404) {
+      // TODO: handle 403 correctly once google fixes their API
+      if (e.getStatusCode() == 403 || e.getStatusCode() == 404) {
         return false;
       }
       throw e;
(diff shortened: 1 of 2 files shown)


honnix:
It's time to have a test case for `ServiceAccountKeyManager` ;)

fabriziodemaria:
Can we add some `todo` comments to remind that we will change this back when Google fixes the problem?

fabriziodemaria:
Following the discussions irl, I think exception handling and related LOGs for the specific Google `Iam` methods should live as much as possible in `ServiceAccountKeyManager` class. But this is not part of this PR anyway.

bergman:
damn codecov… i hit both changed codepaths and the rest of the diff are tests. maybe tests for the tests?

styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
```diff
@@ -64,7 +64,8 @@ public boolean keyExists(String keyName) throws IOException {
           .execute();
       return true;
     } catch (GoogleJsonResponseException e) {
-      if (e.getStatusCode() == 404) {
+      // TODO: handle 403 correctly once google fixes their API
+      if (e.getStatusCode() == 403 || e.getStatusCode() == 404) {
         return false;
       }
       throw e;
```
honnix: probably this is not covered ;)

honnix:
+1

PR #258, opened 2017-08-03
inject resources to container as env

honnix:
This PR changes dequeue event.


honnix:
Postpone the feature.

PR #259, opened 2017-08-07
inject resources to container as env, not via event

honnix:
this is an alternative solution to #258


honnix:
Postpone the feature.

PR #260, opened 2017-08-09
add the ability to scan instances with a range

honnix:
by specifying start and stop, it is possible to query a given range;
this can ease the work for a certain way of doing pagination.

@axelri

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
@@ -136,16 +136,40 @@ public class BigtableStorage {
         }
       }
 
-      return workflowInstancesSet.parallelStream()
-          .map(workflowInstance -> {
-            try {
-              return executionData(workflowInstance);
-            } catch (IOException e) {
-              throw Throwables.propagate(e);
-            }
-          })
-          .sorted(WorkflowInstanceExecutionData.COMPARATOR)
-          .collect(Collectors.toList());
+      return executionData(workflowInstancesSet);
+    }
+  }
+
+  List<WorkflowInstanceExecutionData> executionData(WorkflowId workflowId, String start, String stop)
+      throws IOException {
+    try (final Table eventsTable = connection.getTable(EVENTS_TABLE_NAME)) {
+      final Scan scan = new Scan()
+          .setRowPrefixFilter(Bytes.toBytes(workflowId.toKey() + '#'))
+          .setFilter(new FirstKeyOnlyFilter());
+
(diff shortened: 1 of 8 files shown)


styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
```diff
@@ -169,6 +193,20 @@ WorkflowInstanceExecutionData executionData(WorkflowInstance workflowInstance) t
...
+      Set<WorkflowInstance> workflowInstancesSet) {
+    return workflowInstancesSet.parallelStream()
+        .map(workflowInstance -> {
+          try {
+            return executionData(workflowInstance);
+          } catch (IOException e) {
+            throw Throwables.propagate(e);
+          }
+        })
+        .sorted(WorkflowInstanceExecutionData.COMPARATOR)
```
axelri: This is for another PR, but it's confusing that we sort the instances on timestamp when the order appears guaranteed by BigTable. I was thinking this could relate to `parallelStream` but the collection order in Java 8 streams are guaranteed to be the same as the original, if the streams are created from Lists (see [these docs](https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html#Ordering)). Maybe it's a parallel optimization thing, but in that case the speed benefit is probably not worth the mental burden.

honnix: @rouzwawi Can you recall the reasoning behind?

axelri:
LGTM, except for the sorting thing which should be addressed in another PR

styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
rouzwawi: Isn't it because the input collection is a Set, that does not guarantee any order? It might be a SortedSet that is actually passed in here, but I would guess that the additional sorting here is to make sure that any Set that is passed in would actually be correctly sorted.

honnix: It's a hash set. But don't understand why a set? Do we expect duplicates? Using a list would do the job, right.

rouzwawi: The reason it's a set is because we're iterating through all events from Bigtable, and each WorkflowInstance will have several events. So the set is needed.

honnix: Ah OK. I thought we store events in columns of the same row.

PR #261, opened 2017-08-10
Do not delete workflow if:

honnix:
the deletion request comes from a lower API or via polling


danielnorberg:
Some tests would be good. Apart from that, 👍

honnix:
Yeah that should be there. Working on it.

PR #262, opened 2017-08-11
I could be wrong but nobody seems using scopes

honnix:
@rouzwawi @danielnorberg  PTAL


danielnorberg:
👎  This is required for service accounts.

honnix:
I thought so too, but actually not really.

danielnorberg:
Interesting.

PR #264, opened 2017-08-17
move rate limiting to scheduler

danielnorberg:
Avoid WFI's timing out waiting on submission rate limiting.

This can cause a long backlog of duplicate pod submissions to build up, produce a lot of duplicate executions and delayed executions.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -160,9 +160,9 @@ public class Scheduler {
     timedOutInstances.forEach(this::sendTimeout);
 
     for (InstanceState eligibleInstance : eligibleInstances) {
-      final boolean stop = limitAndDequeue(
+      final boolean proceed = limitAndDequeue(
           resources, workflowResourceReferences, currentResourceUsage, eligibleInstance);
-      if (stop) {
+      if (!proceed) {
         break;
       }
     }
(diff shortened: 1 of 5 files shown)


danielnorberg:
@rouzwawi PTAL

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -263,6 +266,11 @@ private boolean shouldExecute(RunState runState) {
   }
 
   private void sendDequeue(InstanceState instanceState) {
+    if (!dequeueRateLimiter.tryAcquire()) {
+      LOG.debug("Dequeue rate limited");
+      return;
```
rouzwawi: This would cause resource usage to be incremented without the instance being dequeued. This happens on the call side of `sendDequeue`. Either return a boolean from this, or check the limited before calling?

danielnorberg: Good catch. I put the rate limiting inside `sendDequeue` because of the code path fork in [tick()]( https://github.com/spotify/styx/pull/264/files/672e94bc80a01b2bdb1a31519b422e1b1be17f70#diff-a473be0c753046d2e7f56a00c3ed5b62R172). Maybe we can eliminate that fork and have all dequeuing go through `evaluateResourcesForDequeue` and have rate limiting happen before the resource usage counters are modified.

danielnorberg: I removed the code path fork, promoted `limitAndDequeue` to be an actual method and merged `evaluateResourcesForDequeue` into it. Now `tick` calls `limitAndDequeue` on each eligible instance until it returns false to signal that no further dequeue is currently possible.

wdyt?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -148,33 +150,24 @@ void tick() {
                 ConcurrentHashMap::new,
                 counting()));
 
-    updateStats(resources, currentResourceUsage);
```
rouzwawi: Does this mean that the stats update always was one tick behind?

rouzwawi:
LGTM! 👍

PR #265, opened 2017-08-25
exit code counter metric

bergman:


danielnorberg:
I think the reason we closed this the last time is that the meter itself already contains a monotonically increasing counter.

danielnorberg:
How about making semantic-metrics report the meter count instead? https://github.com/spotify/semantic-metrics/pull/18

danielnorberg:
This should be addressed in #361

PR #266, opened 2017-08-28
remove v0-v2 API endpoints

honnix:
Deprecated Java classes are still there. We either remove them
completely or keep them there, but will address that in a different PR.

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
@@ -20,7 +20,6 @@
 
 package com.spotify.styx.api;
 
-import static com.spotify.styx.api.Api.Version.V2;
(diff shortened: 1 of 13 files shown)


styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -232,10 +220,6 @@ public WorkflowResource(Storage storage, String schedulerServiceBaseUrl) {
     }
   }
 
-  private String schedulerApiUrl(CharSequence... parts) {
```
axelri: Is this strictly related to this PR or rather a rest from the prior POST api?

honnix: Yeah, it is a clean up.

axelri:
LGTM

PR #267, opened 2017-08-29
removed deprecated api endpoints code

honnix:
@fabriziodemaria PTAL


fabriziodemaria:
We should probably delete this as well:
https://github.com/spotify/styx/blob/gh-pages/api-deprecated.html

fabriziodemaria:
There is still some `v2` references in the apib file:
https://github.com/spotify/styx/blob/remove-deprecated-code/doc/api.apib#L223

honnix:
Yeah, right, should remove those as well. Will fix. gh-pages branch I will just delete after this PR gets merged.

fabriziodemaria:
Probably more related to #266:
- `v0` found in classes `ApiVersionTest` and `Api`
- Should we get rid of this since there is no more old API versions available? https://github.com/spotify/styx/blob/remove-deprecated-code/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java#L553

honnix:
Removing them from enum seems ok. That method is still useful as it was designed to check API version and disallow lower versions, but the comment should be revised. Let's address those in this PR.

honnix:
We can keep the enum untouched as removing it will break a few things, so probably not worth.

fabriziodemaria:
> That method is still useful as it was designed to check API version and disallow lower versions, but the comment should be revised.

I believe we don't need to check for lower versions, if V3 is the only one available. However, we want to still disregard the polled workflow definition when a `fromApi` value is present. This can be kept as the default behaviour until we reach a final decisions on how to manage Workflow configurations in Styx (via SchedulerSource, via API, or both).

honnix:
That's true. We will only have V3 after this is merged. However we may have V4, and etc. later, so keeping this method might be still useful. But again, as you commented, the future is not clear yet, let's keep it for now.

fabriziodemaria:
👍

PR #268, opened 2017-08-29
rename the artifactId

honnix:
as it pollutes com/spotify/


danielnorberg:
👍 

Not sure we even want that report artifact created and uploaded.

PR #269, opened 2017-08-29
Revert

honnix:
revert undeployable changes.


honnix:
This is to revert the breaking changes in case we need to fix something and get the fix deployed quickly.

PR #270, opened 2017-08-29
Remove old endpoints

honnix:
This is basically a redo of #266 and #267.

DO NOT merge until the last minute to deploy.


styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -85,15 +83,8 @@ public WorkflowResource(Storage storage, DockerImageValidator dockerImageValidat
             rc -> patchState(arg("cid", rc), arg("wfid", rc), rc.request()))
     );
 
-    final List<Route<AsyncHandler<Response<ByteString>>>> sunsetRoutes = Collections.singletonList(
-        Route.with(
-            json(), "PATCH", BASE + "/<cid>/state",
-            rc -> patchState(arg("cid", rc), rc.request()))
```
danielnorberg: Should this `patchState()` method also be removed?

honnix: Yeah, it should.

honnix: Gone now.

PR #274, opened 2017-09-01
register active states while starting up

honnix:
@fabriziodemaria PTAL


fabriziodemaria:
Is there any possible race condition due to the fact that we call `setupMetrics` after `startScheduleSources`? We can get a workflow changed via ScheduleSource while registering all the workflows in the cache.

Also I think in `workflowChanged` method, the registering code should happen after the `isGreaterOrEqualApiVersion` check.

honnix:
Race condition is fine I think. It is a ConcurrentMap and `.all()` returns of copy. The metric registration is also ConcurrentMap.

Yeah, that should be moved after version check.

fabriziodemaria:
👍

danielnorberg:
👍

PR #275, opened 2017-09-01
proxy: remove auth header to avoid duplicate headr

danielnorberg:
The apollo http client decorates the outgoing request with the auth header from the incoming request without bothering to check whether there already is a auth header. This causes nginx on the scheduler to fail the proxied request with 400 Bad Request.


honnix:
👍

PR #276, opened 2017-09-01
proxy: do not use request scoped client

danielnorberg:


danielnorberg:
@honnix PTAL

honnix:
There are other cases we forward request to scheduler. Need to fix as well.

danielnorberg:
As per offline discussion we'll keep `BackfillResource` as is for now in anticipation of future refactoring.

PR #277, opened 2017-09-06
Add delay for pod cleanup

NatashaL:


fabriziodemaria:
It would be good to allow us setting the `DEFAULT_POD_NON_DELETE_SECONDS` in Datastore rather than re-deploying the service every time we need to change the value.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -294,13 +308,46 @@ static EnvVar envVar(String name, String value) {
   @Override
   public void cleanup(WorkflowInstance workflowInstance, String executionId) {
     if (!debug.get()) {
-      LOG.info("Cleaning up {} pod: {}", workflowInstance.toKey(), executionId);
-      client.pods().withName(executionId).delete();
+      Pod pod = client.pods().withName(executionId).get();
```
honnix: `final`

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -294,13 +308,46 @@ static EnvVar envVar(String name, String value) {
...
+
+      if (pod != null) {
```
honnix: We can probably just streamline this using `Optional` and skip that debug message.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -294,13 +308,46 @@ static EnvVar envVar(String name, String value) {
...
+    Optional<ContainerStatus> optionalTerminatedStatus =
+        pod.getStatus()
+            .getContainerStatuses()
+            .stream()
+            .filter(containerStatus -> containerStatus.getName().equals(STYX_RUN)
+                                       && Optional
+                                           .ofNullable(containerStatus.getState().getTerminated())
+                                           .isPresent()
+                                       && Optional.ofNullable(
+                containerStatus.getState().getTerminated().getFinishedAt()).isPresent())
```
honnix: These two conditions can be merged I think. It's already `Optional`.

```
            .filter(containerStatus -> containerStatus.getName().equals(STYX_RUN)
                                       && Optional
                                           .ofNullable(containerStatus.getState().getTerminated())
                                           .map(ContainerStateTerminated::getFinishedAt).isPresent())
```

honnix: Maybe separated filters is clearer to read:
```
    Optional<ContainerStatus> optionalTerminatedStatus =
        pod.getStatus()
            .getContainerStatuses()
            .stream()
            .filter(containerStatus -> containerStatus.getName().equals(STYX_RUN))
            .filter(containerStatus -> Optional
                .ofNullable(containerStatus.getState().getTerminated())
                .map(ContainerStateTerminated::getFinishedAt)
                .isPresent())
            .findFirst();
```

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -294,13 +308,46 @@ static EnvVar envVar(String name, String value) {
...
+            .findFirst();
+
+    if (optionalTerminatedStatus.isPresent()) {
```
honnix: How about
```
    return optionalTerminatedStatus
        .map(x -> Instant
            .parse(x.getState().getTerminated().getFinishedAt())
            .isBefore(clock.instant().minus(Duration.ofSeconds(podNonDeleteSeconds))))
        .orElse(true);
```

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -294,13 +308,46 @@ static EnvVar envVar(String name, String value) {
...
+      } else {
+        LOG.debug("Cannot cleanup pod {} because not found", executionId);
+      }
     } else {
       LOG.info("Keeping {} pod: {}", workflowInstance.toKey(), executionId);
     }
   }
 
+  private boolean isNonDeletePeriodExpired(Pod pod) {
+    Optional<ContainerStatus> optionalTerminatedStatus =
```
honnix: final

honnix:
Regarding the comment of @fabriziodemaria `StyxScheduler::updateRuntimeConfig` can be used as reference.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -102,6 +106,8 @@
   static final String TRIGGER_ID = "STYX_TRIGGER_ID";
   static final String TRIGGER_TYPE = "STYX_TRIGGER_TYPE";
   private static final int DEFAULT_POLL_PODS_INTERVAL_SECONDS = 60;
+  private static final int DEFAULT_POD_NON_DELETE_SECONDS = 120;
```
danielnorberg: This kind of thing is usually called `grace period`, could we use that here as well?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -122,22 +128,30 @@
   private final KubernetesGCPServiceAccountSecretManager serviceAccountSecretManager;
   private final Debug debug;
   private final int pollPodsIntervalSeconds;
+  private final int podNonDeleteSeconds;
```
danielnorberg: How about `podCleanupGracePeriodSeconds` ?

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
```diff
@@ -594,49 +594,6 @@ public void doNotDeleteWorkflowFromNoneAPI() throws Exception {
   }
 
   @Test
-  public void cleansUpDockerRunsWhenTerminating() throws Exception {
```
danielnorberg: Should this test be updated to verify that cleanup happens after some grace period using `timePasses()` instead?

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
```diff
@@ -594,49 +594,6 @@ public void doNotDeleteWorkflowFromNoneAPI() throws Exception {
...
-    injectEvent(Event.started(workflowInstance));
-    injectEvent(Event.terminate(workflowInstance, Optional.of(20)));
-    awaitWorkflowInstanceState(workflowInstance, RunState.State.QUEUED);
-
-    assertThat(dockerCleans, contains(runSpec.executionId()));
-  }
-
-  @Test
-  public void cleansUpDockerRunsWhenFailing() throws Exception {
-    givenTheTimeIs("2016-03-14T15:59:01Z");
```
danielnorberg: Same as above

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
NatashaL: There's already a `grace period` parameter in kubectl delete that doesn't do what we need it to do. So we opted for this naming in order to avoid confusion and people thinking they're the same thing

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/DockerRunnerHandlerTest.java
```diff
@@ -164,36 +178,6 @@ public void shouldHaltIfMissingExecutionDescription() throws Exception {
     assertThat(stateManager.get(workflowInstance), is(nullValue()));
   }
 
-  @Test
-  public void shouldPerformCleanupOnFailed() throws Exception {
```
danielnorberg: Should these tests instead be updated to verify that cleanup happens after grace period?

danielnorberg:
For debugging purposes, how about keeping failing pods around longer if exit code is not `0` or `20` ?

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
danielnorberg: Ok, seems these tests are now invalid, nvm.

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/DockerRunnerHandlerTest.java
NatashaL: As we talked, we can't do that here due to the DockerRunner being mocked. But the isolated  behaviour is tested in KubernetesDockerRunnerTest

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
NatashaL: tnx, fixed

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -294,13 +309,42 @@ static EnvVar envVar(String name, String value) {
   @Override
   public void cleanup(WorkflowInstance workflowInstance, String executionId) {
     if (!debug.get()) {
-      LOG.info("Cleaning up {} pod: {}", workflowInstance.toKey(), executionId);
-      client.pods().withName(executionId).delete();
+      final Optional<Pod> pod = Optional.ofNullable(client.pods().withName(executionId).get());
+
+      pod.filter(this::isNonDeletePeriodExpired).ifPresent(p -> {
+        client.pods().withName(executionId).delete();
+        LOG.info("Cleaning up {} pod: {}", workflowInstance.toKey(), executionId);
```
honnix: What is the reason reordering the two statements?

honnix:
As we discussed, this change is good enough to be shipped. So let's address other comments in new PRs if needed. So 👍

PR #279, opened 2017-09-11
Fix bug in delay of pod cleanup

NatashaL:
Fix bug in positioning the cleanup within the state machine.
Also fix failing tests.

There are instances where the a workflowInstance's runState is still kept
in the stateManager even though the execution failed. For example, the runState is
removed from the stateManager only when the runState isTerminal (meaning, Done or Error),
whereas Failed and Terminated are not considered terminal states. In
the case of cleaning up pods, we want to clean the pods once a workflowInstance
enters one of all those 4 states (Failed/Terminated/Done/Error). Running cleanup(..) only
under the condition !runState.isPresent() prevented us to do so.

Now we try to cleanup the pod if runState isPresent or delete directly if the pod is not tracked by any runState. In `cleanup` we make sure deleting the pod that is in terminated state and stays after graceful period. `debug` flag still makes the final decision for deletion.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -48,7 +48,6 @@ import com.spotify.styx.state.Trigger;
 import com.spotify.styx.util.Debug;
 import com.spotify.styx.util.TriggerUtil;
 import io.fabric8.kubernetes.api.model.ContainerBuilder;
-import io.fabric8.kubernetes.api.model.ContainerStateTerminated;
 import io.fabric8.kubernetes.api.model.ContainerStatus;
 import io.fabric8.kubernetes.api.model.EnvVar;
 import io.fabric8.kubernetes.api.model.EnvVarBuilder;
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -430,8 +425,9 @@ private synchronized void tryPollPods() {
       final Optional<RunState> runState = lookupPodRunState(pod, workflowInstance.get());
       if (runState.isPresent()) {
         emitPodEvents(Watcher.Action.MODIFIED, pod, runState.get());
-      } else {
         cleanup(workflowInstance.get(), pod.getMetadata().getName());
+      } else {
+        deletePod(workflowInstance.get(), pod.getMetadata().getName());
```
danielnorberg: I guess a reason we didn't do remove pods with no matching WF runstate before is that (at the time) the pod execution id might not be  present in the WF runstate. Pod events / polling could thus race with the execution id persistence and incorrectly cause recently created pods to be deleted. This changed in https://github.com/spotify/styx/pull/198 and the execution id is now created before the actual pod. Let's hope this holds true in the future as well =)

danielnorberg:
It seems like before this PR we would only delete pods if the runstate is present. The comment above on the other hand says we only delete pods if the runstate is _not_ present. Is that just a typo or are we misunderstanding something?

danielnorberg:
LGTM 👍

PR #280, opened 2017-09-12
Read pod deletion delay from datastore

NatashaL:


honnix:
Not sure how much delaying pod cleanup can help deliver logs, so hold on.

danielnorberg:
One current suspicion is that if fluentd buffers are full, fluentd won't pick up new container log files. But in that case it could be necessary to wait a long time before deleting the pod, and we probably don't want that.

If buffers are not full fluentd seems to pick up new logs quickly enough.

bergman:
👍 after rebase

styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -30,10 +30,12 @@
 import static java.util.Optional.of;
 import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasEntry;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.lessThan;
```
bergman: is this import used?

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -167,7 +169,9 @@ ScheduledExecutorService create(
 
     private Time time = Instant::now;
     private StorageFactory storageFactory = storage(StyxScheduler::storage);
-    private DockerRunnerFactory dockerRunnerFactory = StyxScheduler::createDockerRunner;
+    private DockerRunnerFactory dockerRunnerFactory =
+        (id, environment, stateManager1, scheduler1, stats, debug, delay) -> createDockerRunner(
```
bergman: what happened here?

honnix:
I'm more concerned that whether we really need this feature. It's a complicated change without bringing in enough value. @bergman

styx-scheduler-service/src/main/java/com/spotify/styx/docker/DockerRunner.java
```diff
@@ -122,12 +123,14 @@ static DockerRunner local(ScheduledExecutorService executorService, StateManager
     return new LocalDockerRunner(executorService, stateManager);
   }
 
-  static DockerRunner kubernetes(NamespacedKubernetesClient kubernetesClient, StateManager stateManager,
-      Stats stats, ServiceAccountKeyManager serviceAccountKeyManager, Debug debug) {
+  static DockerRunner kubernetes(NamespacedKubernetesClient kubernetesClient,
+                                 StateManager stateManager,
+                                 Stats stats, ServiceAccountKeyManager serviceAccountKeyManager,
```
bergman: formatting seems off

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -46,8 +46,10 @@
 import com.spotify.styx.state.StateManager;
 import com.spotify.styx.state.Trigger;
 import com.spotify.styx.util.Debug;
+import com.spotify.styx.util.Delay;
 import com.spotify.styx.util.TriggerUtil;
 import io.fabric8.kubernetes.api.model.ContainerBuilder;
+import io.fabric8.kubernetes.api.model.ContainerStateTerminated;
```
bergman: i don't see the usage of this import in this PR?

bergman:
It's not complicated, it's just that it needs to be passed in through a few layers.

honnix:
I'm confident that we don't need the extra complexity. Close for now?

fabriziodemaria:
> I'm confident that we don't need the extra complexity. Close for now?

+1 on closing

honnix:
this changes doesn't seem to be need in practice, so close it but keep the branch.

PR #283, opened 2017-09-14
bring back the action when transiting

honnix:
It feels more correct still act upon transiting, while the actual runner
can decide not to do anything.

@NatashaL PTAL


NatashaL:
👍

PR #284, opened 2017-09-14
client: default to https://

danielnorberg:


danielnorberg:
@honnix PTAL

dflemstr:
:+1:

honnix:
A few test cases need to be fixed. We actually have very good test cases!

danielnorberg:
indeed

honnix:
+1

PR #285, opened 2017-09-20
WIP: Remove deprecated event usage in tests

danielnorberg:


honnix:
What's the plan @danielnorberg ? Do we want to keep working on this? To me this is still valuable to get rid of tech debt.

ulzha:
Discussed in fireside chat. May become irrelevant due to https://jira.spotify.net/browse/DATAEX-1515

danielnorberg:
I think we should work on this and I think it should be a ticket in our backlog.

PR #286, opened 2017-09-21
Add support for consuming internal events asynchronously

fabriziodemaria:
This allows to asynchronously consume the processed internal events. A possible application for this is to publish such events to third-party services by implementing the `EventInterceptor` interface.

EDIT: In the latest implementation the `EventInterceptor` interface has been removed in favour of the standard Java `Consumer` interface.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -57,6 +57,7 @@ import com.spotify.styx.api.SchedulerResource;
 import com.spotify.styx.docker.DockerRunner;
 import com.spotify.styx.docker.WorkflowValidator;
 import com.spotify.styx.model.Event;
+import com.spotify.styx.model.SequenceEvent;
 import com.spotify.styx.model.Workflow;
 import com.spotify.styx.model.WorkflowId;
 import com.spotify.styx.model.WorkflowInstance;
@@ -81,6 +82,7 @@ import com.spotify.styx.storage.Storage;
 import com.spotify.styx.util.CachedSupplier;
 import com.spotify.styx.util.Debug;
 import com.spotify.styx.util.DockerImageValidator;
+import com.spotify.styx.util.IsClosedException;
 import com.spotify.styx.util.RetryUtil;
 import com.spotify.styx.util.StorageFactory;
 import com.spotify.styx.util.Time;
@@ -143,6 +145,7 @@ public class StyxScheduler implements AppInit {
   public interface ScheduleSources extends Supplier<Iterable<ScheduleSourceFactory>> { }
   public interface StatsFactory extends Function<Environment, Stats> { }
   public interface PublisherFactory extends Function<Environment, Publisher> { }
+  public interface EventConsumerFactory extends Function<Environment, Consumer<SequenceEvent>> { }
 
   @FunctionalInterface
   interface DockerRunnerFactory {
@@ -173,6 +176,7 @@ public class StyxScheduler implements AppInit {
     private PublisherFactory publisherFactory = (env) -> Publisher.NOOP;
     private RetryUtil retryUtil = DEFAULT_RETRY_UTIL;
     private WorkflowResourceDecorator resourceDecorator = WorkflowResourceDecorator.NOOP;
+    private EventConsumerFactory eventConsumerFactory = (env) -> (event) -> { };
 
     public Builder setTime(Time time) {
       this.time = time;
@@ -219,6 +223,11 @@ public class StyxScheduler implements AppInit {
       return this;
     }
 
+    public Builder setEventConsumerFactory(EventConsumerFactory eventConsumerFactory) {
+      this.eventConsumerFactory = eventConsumerFactory;
+      return this;
+    }
+
     public StyxScheduler build() {
       return new StyxScheduler(this);
     }
@@ -243,6 +252,7 @@ public class StyxScheduler implements AppInit {
   private final PublisherFactory publisherFactory;
   private final RetryUtil retryUtil;
   private final WorkflowResourceDecorator resourceDecorator;
+  private final EventConsumerFactory eventConsumerFactory;
 
   private StateManager stateManager;
   private Scheduler scheduler;
@@ -259,6 +269,7 @@ public class StyxScheduler implements AppInit {
     this.publisherFactory = requireNonNull(builder.publisherFactory);
     this.retryUtil = requireNonNull(builder.retryUtil);
     this.resourceDecorator = requireNonNull(builder.resourceDecorator);
+    this.eventConsumerFactory = requireNonNull(builder.eventConsumerFactory);
   }
 
   @Override
@@ -283,9 +294,11 @@ public class StyxScheduler implements AppInit {
     closer.register(publisher);
 
     final ScheduledExecutorService executor = executorFactory.create(3, schedulerTf);
-    final ExecutorService eventWorker = Executors.newFixedThreadPool(16, eventTf);
     closer.register(executorCloser("scheduler", executor));
-    closer.register(executorCloser("event-worker", eventWorker));
+    final ExecutorService outputHandlerExecutor = Executors.newFixedThreadPool(16, eventTf);
+    closer.register(executorCloser("output-handler", outputHandlerExecutor));
+    final ExecutorService eventConsumerExecutor = Executors.newSingleThreadExecutor();
+    closer.register(executorCloser("event-consumer", eventConsumerExecutor));
 
     final Stats stats = statsFactory.apply(environment);
     final WorkflowCache workflowCache = new InMemWorkflowCache();
@@ -294,7 +307,8 @@ public class StyxScheduler implements AppInit {
     warmUpCache(workflowCache, storage);
 
     final QueuedStateManager stateManager = closer.register(
-        new QueuedStateManager(time, eventWorker, storage));
+        new QueuedStateManager(time, outputHandlerExecutor, storage,
+            eventConsumerFactory.apply(environment), eventConsumerExecutor));
 
     final Config staleStateTtlConfig = config.getConfig(STYX_STALE_STATE_TTL_CONFIG);
     final TimeoutConfig timeoutConfig = TimeoutConfig.createFromConfig(staleStateTtlConfig);
@@ -360,7 +374,7 @@ public class StyxScheduler implements AppInit {
   }
 
   @VisibleForTesting
-  void receive(Event event) throws StateManager.IsClosed {
+  void receive(Event event) throws IsClosedException {
     stateManager.receive(event);
   }
 
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/DockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/DockerRunner.java
@@ -20,14 +20,8 @@
 
 package com.spotify.styx.docker;
 
-import static java.util.Optional.empty;
-
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.auto.value.AutoValue;
-import com.google.common.collect.ImmutableList;
 import com.spotify.styx.ServiceAccountKeyManager;
 import com.spotify.styx.model.WorkflowConfiguration;
-import com.spotify.styx.model.WorkflowConfiguration.Secret;
 import com.spotify.styx.model.WorkflowInstance;
 import com.spotify.styx.monitoring.Stats;
 import com.spotify.styx.state.StateManager;
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
@@ -31,6 +31,7 @@ import com.spotify.styx.model.WorkflowId;
 import com.spotify.styx.model.WorkflowInstance;
 import com.spotify.styx.storage.Storage;
 import com.spotify.styx.util.AlreadyInitializedException;
+import com.spotify.styx.util.IsClosedException;
 import com.spotify.styx.util.Time;
 import java.io.IOException;
 import java.util.Map;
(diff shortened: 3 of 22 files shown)


danielnorberg:
How about decoupling and publishing a curated stream of events meant for public consumption instead of the internal state machine events?

It would be good if future changes to the internal state machine would not require changing the consumers of this event stream.

fabriziodemaria:
> How about decoupling and publishing a curated stream of events meant for public consumption instead of the internal state machine events?

It could be up to the `EventInterceptor` implementation to manipulate the events according to the different needs.

styx-common/src/main/java/com/spotify/styx/publisher/EventInterceptor.java
```diff
@@ -0,0 +1,40 @@
...
+import com.spotify.styx.model.SequenceEvent;
+
+/**
+ * Interface for acting on internal Styx events. For example, this can be used to publish internal
+ * Styx events to third-party databases.
+ */
+public interface EventInterceptor {
+
+  /**
+   * Called when a state machine transition happens due to a new internal Styx event.
```
ulzha: "Called _when_ transition happens" is uninformative. It's actually called _after_ I guess? not before? and not concurrently? Also `interceptedEvent` is uninformative in the same manner. We could reflect more exact semantics in the name easily.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -173,6 +177,7 @@ ScheduledExecutorService create(
     private PublisherFactory publisherFactory = (env) -> Publisher.NOOP;
     private RetryUtil retryUtil = DEFAULT_RETRY_UTIL;
     private WorkflowResourceDecorator resourceDecorator = WorkflowResourceDecorator.NOOP;
+    private EventInterceptorFactory eventInterceptorFactory  = (env) -> EventInterceptor.NOOP;
```
ulzha: Odd whitespace

ulzha:
What's the conceptual difference between `EventConsumer` and `EventInterceptor`?

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -82,10 +83,15 @@
   private AtomicInteger activeEvents = new AtomicInteger(0);
   private volatile boolean running = true;
 
-  public QueuedStateManager(Time time, Executor workerPool, Storage storage) {
+  public QueuedStateManager(
+      Time time,
+      Executor workerPool,
+      Storage storage,
+      EventConsumer eventConsumer) {
```
ulzha: How come we daisy chain consumption in an ad-hoc way into here? It seems to add an unneeded responsibility to the state manager. The `StateManager` already quacks like an `EventConsumer` (it has a `CompletionStage<Void> receive(Event event)` method). It reacts to events. Just like event consumers do. I would propose a unifying refactor along the lines of `interface StateManager extends EventConsumer`.

fabriziodemaria:
> What's the conceptual difference between EventConsumer and EventInterceptor?

The `EventConsumer` takes care of the way events coming from the `StateManager` are consumed, and the current implementation `QueuedEventConsumer` allows for single-threaded asynchronous event queuing and consumption.

The `EventInterceptor` only takes care of the code to be executed when an event is actually consumed  by the `EventConsumer`.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedEventConsumer.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Objects;
+import java.util.Queue;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Single threaded asynchronous event consumer queue. It requires a {@link EventInterceptor}
```
ulzha: Lots of ad-hoc code when there seem to be primitives in Java for this... I'm looking at the snippet with `FinalizableDelegatedExecutorService` in https://stackoverflow.com/questions/30208786/run-only-one-thread-with-threadpoolexecutor

ulzha:
| The EventInterceptor only takes care of the code to be executed when an event is actually consumed by the EventConsumer.

I mean this relation is unclear to a person reading code. It's not enough to explain it in a comment on PR, because the future readers won't see it. Could be clarified by expanding the javadoc, or could be clarified by giving obviously related names or obviously related APIs to these classes. In `EventConsumer` there's nothing connecting it with `EventInterceptor` or vice versa.

honnix:
General comments before reviewing: we need to fix test coverage. :)

styx-scheduler-service/src/main/java/com/spotify/styx/state/EventFeeder.java
```diff
@@ -0,0 +1,89 @@
...
+  @Override
+  public void close() throws IOException {
+    if (executor.isTerminating() || executor.isShutdown()) {
+      return;
+    }
+    executor.shutdown();
+    LOG.info("Shutting down, waiting for queued events to be consumed");
+    try {
+      if (!executor.awaitTermination(SHUTDOWN_GRACE_PERIOD_SECONDS, TimeUnit.SECONDS)) {
+        executor.shutdownNow();
```
honnix: `shutdownNow` actually returns a list of unfinished tasks, so I think there is no need to get queue size (it won't work anyway because queue has been drained). Then we can use `Executors` to avoid using concrete class.

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerServiceFixture.java
```diff
@@ -358,4 +365,11 @@ private Connection setupBigTableMockTable(int numFailures) {
     }
     return bigtable;
   }
+
+  private class InjectingEventConsumer implements Consumer<SequenceEvent> {
```
honnix: Just inline this using lambda.

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
```diff
@@ -65,6 +66,8 @@
   private static final Trigger TRIGGER1 = Trigger.unknown("trig1");
   private static final Trigger TRIGGER2 = Trigger.unknown("trig2");
   private static final Trigger TRIGGER3 = Trigger.unknown("trig3");
+  private static final EventFeeder<SequenceEvent> eventFeeder =
+      new EventFeeder<>(NoopEventConsumer.NOOP);
```
honnix: inline with lambda?

styx-scheduler-service/src/test/java/com/spotify/styx/state/EventFeederTest.java
```diff
@@ -0,0 +1,126 @@
...
+      trackedEvents.add(sequenceEvent);
+    }
+  }
+
+  private class ExceptionalEventConsumer implements Consumer<SequenceEvent> {
+    @Override
+    public void accept(SequenceEvent sequenceEvent) {
+      try {
+        //Todo better
+        Thread.sleep(1000);
```
honnix: This is can be a good hint that we can probably inject executor to `EventFeeder`.

styx-scheduler-service/src/main/java/com/spotify/styx/state/EventFeeder.java
```diff
@@ -0,0 +1,89 @@
...
+  private final Consumer<T> eventConsumer;
+  private final ThreadPoolExecutor executor;
+
+  public EventFeeder(Consumer<T> eventConsumer) {
+    this.eventConsumer = Objects.requireNonNull(eventConsumer);
+    this.executor = new ThreadPoolExecutor(1, 1, 0L,
+        TimeUnit.MILLISECONDS, new LinkedBlockingQueue<>());
+  }
+
+  void enqueue(T event) throws IsClosed {
```
honnix: shall we call this `newEvent` or `onEvent` to not to disclose `queue` which is rather a internal design of this class?

styx-common/src/main/java/com/spotify/styx/util/IsClosed.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.styx.util;
+
+/**
+ * Exception that signals a closed state.
+ */
+public class IsClosed extends Exception {
```
ulzha: Inconsistent naming with other Exceptions in com.spotify.styx.util

styx-scheduler-service/src/main/java/com/spotify/styx/state/EventFeeder.java
ulzha: Or `feed` if we called the class Feeder.

But consistency would also be achieved with `EventQueue` + `enqueue()`, or `EventConsumerExecutor` + `execute()`.

My overarching point is _it's not very useful to have a Styx-specific abstraction over a vanilla ExecutorService API here_. It just makes it necessary for users to read doc (realistically, read code) to understand what added value Feeder brings over an ExecutorService.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -271,6 +278,8 @@ private void process(RunState state, long counterPosition, Event event,
 
     try {
       storeEvent(sequenceEvent);
+      // racy...
```
ulzha: Racy in what regard? Have we promised any invariant that this can violate in edge cases? Should be clarified, or the comment can go away.

honnix:
@ulzha ```My overarching point is it's not very useful to have a Styx-specific abstraction over a vanilla ExecutorService API here. It just makes it necessary for users to read doc (realistically, read code) to understand what added value Feeder brings over an ExecutorService.```

A very good comment, I almost went the same direction. So, let's get rid of `feeder` completely.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
honnix: I think it's an old comment and still valid very likely.

styx-scheduler-service/src/main/java/com/spotify/styx/state/EventFeeder.java
honnix: A very good comment, I almost went the same direction. So, let's get rid of feeder completely.

ulzha:
All in all great improvement - I think it's achieving the same in much simpler and maintainable way now that EventConsumer is replaced by Consumer<>, and the ad-hoc Thread, Queue, and CountDownLatch tinkering is replaced with ThreadPoolExecutor.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
honnix: Disregard my comment. This is a new comment. Let's make it clearer why it is racy.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/DockerRunner.java
```diff
@@ -20,14 +20,8 @@
 
 package com.spotify.styx.docker;
 
-import static java.util.Optional.empty;
-
-import com.fasterxml.jackson.annotation.JsonProperty;
```
bergman: Whoever is running without automatic `optimize imports…`, please enable it.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -71,8 +73,10 @@
   static final long NO_EVENTS_PROCESSED = -1L;
 
   private final Time time;
-  private final Executor workerPool;
+  private final Executor outputHandlerExecutor;
```
ulzha: :star:

ulzha:
Looks good to me, just codecov remains.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -288,8 +298,12 @@ private void process(RunState state, long counterPosition, Event event,
       throw t;
     }
 
+    // If the service is stopped right before this line, the event was successfully stored but never
+    // going to be consumed. Event consumption is only best effort with the current implementation.
+    consumeEvent(sequenceEvent);
```
bergman: I would inline this function unless it's used elsewhere. It looks weird to have only this one in a function and the other one inlined (the `outputHandlerExecutor.execute` one).

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -288,8 +298,16 @@ private void process(RunState state, long counterPosition, Event event,
       throw t;
     }
 
+    // If the service is stopped right before this line, the event was successfully stored but never
+    // going to be consumed. Event consumption is only best effort with the current implementation.
```
bergman: this is true for output handlers as well, right?

fabriziodemaria: It is true. Are you suggesting to remove the comment or adding more details to the existing one?

bergman: not sure, but i find it a bit misleading as it is right now

fabriziodemaria: I am ok with removing this comment altogether and if needed explain more in-depth what these racing conditions might cause and what guarantees they break in the docs, in a separate PR.

bergman: sounds good

bergman:
👍

PR #287, opened 2017-09-26
racing between poller and watcher

honnix:
We have observed cases that pod is not seen for running state. That
happens after poller has fetched a list of pods, then styx creates a
new pod, that pod transits into running and the state machine transits
into running as well. So we end up with having a running state but no
pod.

To fix this, first get a snapshot of running states and then fetch pod
list. If the state is running, the associated pod should be in the list;
while if the state is not running yet, we will not take any action.

@danielnorberg PTAL

We may just stop using watcher at all, but that has bigger impact.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -427,14 +427,17 @@ class KubernetesDockerRunner implements DockerRunner {
         .watch(new PodWatcher(Integer.parseInt(resourceVersion)));
   }
 
-  void examineRunningWFISandAssociatedPods(PodList podList) {
-    final Set<WorkflowInstance> runningWorkflowInstances = stateManager.activeStates()
+  private Set<WorkflowInstance> getRunningWorkflowInstances() {
+    return stateManager.activeStates()
         .values()
         .stream()
         .filter(runState -> runState.state().equals(RUNNING))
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -427,14 +427,17 @@ public void init() {
...
+    return stateManager.activeStates()
         .values()
         .stream()
         .filter(runState -> runState.state().equals(RUNNING))
         .map(RunState::workflowInstance)
         .collect(toSet());
+  }
 
+  private void examineRunningWFISandAssociatedPods(Set<WorkflowInstance> runningWorkflowInstances,
+                                           PodList podList) {
```
danielnorberg: indentation

danielnorberg:
Is maybe now this race possible instead?

1. `getRunningWorkflowInstances()` is called and WFI `foo` is included in `runningWorkflowInstances`
2. WFI `foo` Pod completes, RunState transitions into `DONE` and pod is removed
3. `client.pods().list()` is called, `foo` pod is not returned and not included in `PodList list`
4. The already completed `foo` WFI receives a `"No pod associated with this instance"` `runError` event and we get a `Illegal state transition` warning

honnix:
Good catch! Yeah that is a possible race indeed. It’s annoying but doesn’t do much harm as it is right now. I kind of start wondering why we did this check in the first place though.

danielnorberg:
👍

PR #290, opened 2017-09-29
remove all traces of api v0, v1, v2

bergman:


styx-common/src/main/java/com/spotify/styx/api/Api.java
```diff
@@ -38,7 +38,7 @@
 public final class Api {
 
   public enum Version {
-    V0, V1, V2, V3;
+    V3;
 
     public String prefix() {
       return "/api/v" + ordinal();
```
honnix: This would give wrong prefix I think. Surprisingly no test cases fail.

bergman: LOL, didn't see this at first. Would be fun, but also maybe we should use the string `"v3"` in some places rather than this thing.

honnix: Get name and then lower case instead of ordinal should be ok.

bergman: cool that we found this at the same time 😄

bergman:
squashed

styx-common/src/test/java/com/spotify/styx/api/ApiVersionTest.java
```diff
@@ -1,33 +0,0 @@
...
-
-import static org.junit.Assert.assertSame;
-
-import org.junit.Test;
-
-public class ApiVersionTest {
-
-  @Test
-  public void shouldParse() {
-    assertSame(Api.Version.V0, Api.Version.valueOf("V0"));
```
honnix: Shall we keep the test case for V3 and add one more to test `prefix()`?

honnix:
Minor comment, otherwise LGTM. As a note, we will need to modify internal part to be able to ship.

styx-common/src/test/java/com/spotify/styx/api/ApiVersionTest.java
bergman: but parsing? when do we need to do that?

I added back a test for the output of prefix though.

styx-local-files-source/src/main/java/com/spotify/styx/LocalFileScheduleSource.java
```diff
@@ -214,7 +214,7 @@ private void readFile(Path file) {
     final String componentId = componentId(path);
     final URI componentUri = path.toUri();
```
honnix: This URI variable is not used anymore right?

honnix:
👍

PR #291, opened 2017-10-03
check for dependencies before executing wf

danielnorberg:
Add a `WorkflowExecutionGate` scheduler plugin interface that can be used to implement blocking of workflow executions e.g. due to missing dependencies or other unfulfilled preconditions.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -91,6 +91,7 @@ class DatastoreStorage {
   public static final String PROPERTY_CONFIG_DOCKER_RUNNER_ID = "dockerRunnerId";
   public static final String PROPERTY_CONFIG_CONCURRENCY = "concurrency";
   public static final String PROPERTY_CONFIG_CLIENT_BLACKLIST = "clientBlacklist";
+  public static final String PROPERTY_CONFIG_EXECUTION_GATING = "executionGating";
 
   public static final String PROPERTY_WORKFLOW_JSON = "json";
   public static final String PROPERTY_WORKFLOW_ENABLED = "enabled";
@@ -117,6 +118,7 @@ class DatastoreStorage {
   public static final String DEFAULT_CONFIG_DOCKER_RUNNER_ID = "default";
   public static final boolean DEFAULT_WORKFLOW_ENABLED = false;
   public static final boolean DEFAULT_CONFIG_DEBUG_ENABLED = false;
+  public static final boolean DEFAULT_CONFIG_EXECUTION_GATING_ENABLED = false;
 
   public static final int MAX_RETRIES = 100;
 
@@ -151,6 +153,8 @@ class DatastoreStorage {
             read(entity, PROPERTY_CONFIG_DOCKER_RUNNER_ID, DEFAULT_CONFIG_DOCKER_RUNNER_ID))
         .clientBlacklist(this.<String>readStream(entity, PROPERTY_CONFIG_CLIENT_BLACKLIST)
             .collect(toList()))
+        .executionGatingEnabled(
+            read(entity, PROPERTY_CONFIG_EXECUTION_GATING, DEFAULT_CONFIG_EXECUTION_GATING_ENABLED))
         .build();
   }
 
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -91,7 +91,7 @@ class DatastoreStorage {
   public static final String PROPERTY_CONFIG_DOCKER_RUNNER_ID = "dockerRunnerId";
   public static final String PROPERTY_CONFIG_CONCURRENCY = "concurrency";
   public static final String PROPERTY_CONFIG_CLIENT_BLACKLIST = "clientBlacklist";
-  public static final String PROPERTY_CONFIG_EXECUTION_GATING = "executionGating";
+  public static final String PROPERTY_CONFIG_EXECUTION_GATING_ENABLED = "executionGatingEnabled";
 
   public static final String PROPERTY_WORKFLOW_JSON = "json";
   public static final String PROPERTY_WORKFLOW_ENABLED = "enabled";
@@ -154,7 +154,7 @@ class DatastoreStorage {
         .clientBlacklist(this.<String>readStream(entity, PROPERTY_CONFIG_CLIENT_BLACKLIST)
             .collect(toList()))
         .executionGatingEnabled(
-            read(entity, PROPERTY_CONFIG_EXECUTION_GATING, DEFAULT_CONFIG_EXECUTION_GATING_ENABLED))
+            read(entity, PROPERTY_CONFIG_EXECUTION_GATING_ENABLED, DEFAULT_CONFIG_EXECUTION_GATING_ENABLED))
         .build();
   }
 
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -242,7 +242,8 @@ public class Scheduler {
       LOG.debug("Interrupted");
       return false;
     } catch (ExecutionException | TimeoutException e) {
-      LOG.warn("Failed to check execution blockers for {}", instance.workflowInstance(), e);
+      LOG.warn("Failed to check execution blocker for {}, assuming there is no blocker",
+          instance.workflowInstance(), e);
     }
 
     if (blocker.isPresent()) {
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -236,15 +236,14 @@ public class Scheduler {
       CompletionStage<Optional<ExecutionBlocker>> executionBlockerFuture) {
 
     // Check for execution blockers
-    final Optional<ExecutionBlocker> blocker;
+    Optional<ExecutionBlocker> blocker = Optional.empty();
     try {
       blocker = executionBlockerFuture.toCompletableFuture().get(30, TimeUnit.SECONDS);
     } catch (InterruptedException e) {
       LOG.debug("Interrupted");
       return false;
     } catch (ExecutionException | TimeoutException e) {
-      LOG.error("Failed to check for missing dependencies", e);
-      return true;
+      LOG.warn("Failed to check execution blockers for {}", instance.workflowInstance(), e);
     }
 
     if (blocker.isPresent()) {
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -113,7 +113,7 @@ public class Scheduler {
     this.resourceDecorator = Objects.requireNonNull(resourceDecorator);
     this.stats = Objects.requireNonNull(stats);
     this.dequeueRateLimiter = Objects.requireNonNull(dequeueRateLimiter, "dequeueRateLimiter");
-    this.gate = Objects.requireNonNull(gate, "gate");;
+    this.gate = Objects.requireNonNull(gate, "gate");
   }
 
   void tick() {
@@ -234,7 +234,7 @@ public class Scheduler {
       Map<String, Long> currentResourceUsage, InstanceState instance,
       CompletionStage<Optional<ExecutionBlocker>> executionBlockerFuture) {
 
-    // Check for execution blockers
+    // Check for execution blocker
     Optional<ExecutionBlocker> blocker = Optional.empty();
     try {
       blocker = executionBlockerFuture.toCompletableFuture().get(30, TimeUnit.SECONDS);
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -241,7 +241,7 @@ public class Scheduler {
       blocker = executionBlockerFuture.toCompletableFuture().get(30, TimeUnit.SECONDS);
     } catch (InterruptedException e) {
       LOG.debug("Interrupted");
-      return true;
+      return false;
     } catch (ExecutionException | TimeoutException e) {
       LOG.error("Failed to check for missing dependencies", e);
       return true;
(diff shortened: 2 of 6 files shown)


honnix:
I included this PR as well to be part of fireside chat happening some time later.

ulzha:
Just noted in fireside chat that forced reexecution might be needed sooner than in one scheduling interval, because some workflows would change their dependencies depending on wall clock time, for example. (Nondeterministic `requires()`.)

styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowExecutionGate.java
```diff
@@ -0,0 +1,58 @@
...
+import com.spotify.styx.model.ExecutionDescription;
+import com.spotify.styx.model.WorkflowConfiguration;
+import com.spotify.styx.model.WorkflowInstance;
+import com.spotify.styx.state.RunState;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+public interface WorkflowExecutionGate {
```
ulzha: This looks like a good abstraction to have.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -180,13 +270,42 @@ private void updateStats(Map<String, Resource> resources,
 
   private boolean limitAndDequeue(Map<String, Resource> resources,
       Map<WorkflowId, Set<String>> workflowResourceReferences,
-      Map<String, Long> currentResourceUsage, InstanceState instance) {
+      Map<String, Long> currentResourceUsage, InstanceState instance,
+      CompletionStage<List<String>> missingDependenciesFuture) {
```
bergman: This one should probably be more generic, maybe just a `String` with some message about why gate was closed.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -306,4 +425,14 @@ static ResourceWithInstance create(String resource, InstanceState instanceState)
       return new AutoValue_Scheduler_ResourceWithInstance(resource, instanceState);
     }
   }
+
+  @AutoValue
+  abstract static class Execution {
```
bergman: Oooooh, another type called execution something 😄

styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowExecutionGate.java
```diff
@@ -0,0 +1,58 @@
...
+   * workflow instance is dequeued, in order to provide an opportunity to perform cheaper dependency
+   * lookup. Implementations of this method should be careful to consider whether dependency
+   * information gathered e.g. from previous executions has been invalidated by docker image and/or
+   * args changes, etc. This method must not block.
+   *
+   * @param instance The workflow instance to check dependencies for.
+   * @param configuration The {@link WorkflowConfiguration} of the workflow instance.
+   * @return A list of missing dependency resource identifiers, e.g. uris.
+   */
+  CompletionStage<List<String>> missingDependencies(
```
bergman: Calling these missing dependencies seems unnecessarily specific. Perhaps `shouldRun` or similar and just return something like a `String`.

danielnorberg:
This PR now only adds the gating interface and makes the scheduler use it. 

All logic for looking up previous execution configurations, force re-execution policies etc is now supposed to be implemented by the plugin itself.

The interface itself has been reworked to be more generic instead of only dealing with missing dependencies.

@ulzha @honnix PTAL again please

danielnorberg:
Execution gating has also been made runtime configurable so it can be disabled. Workflow instances will then be executed directly without any checking for blockers.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -91,6 +91,7 @@
   public static final String PROPERTY_CONFIG_DOCKER_RUNNER_ID = "dockerRunnerId";
   public static final String PROPERTY_CONFIG_CONCURRENCY = "concurrency";
   public static final String PROPERTY_CONFIG_CLIENT_BLACKLIST = "clientBlacklist";
+  public static final String PROPERTY_CONFIG_EXECUTION_GATING = "executionGating";
```
honnix: `PROPERTY_CONFIG_EXECUTION_GATING` naming is not very aligned with other boolean properties, such as `PROPERTY_CONFIG_DEBUG_ENABLED`.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -103,14 +114,17 @@ public Scheduler(Time time, TimeoutConfig ttls, StateManager stateManager,
     this.resourceDecorator = Objects.requireNonNull(resourceDecorator);
     this.stats = Objects.requireNonNull(stats);
     this.dequeueRateLimiter = Objects.requireNonNull(dequeueRateLimiter, "dequeueRateLimiter");
+    this.gate = Objects.requireNonNull(gate, "gate");;
   }
 
   void tick() {
     final Map<String, Resource> resources;
     final Optional<Long> globalConcurrency;
+    final StyxConfig config;
```
honnix: this part needs a bit of rework if we merge https://github.com/spotify/styx/pull/360

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -159,17 +173,44 @@ void tick() {
...
-      }
-    }
+    limitAndDequeue(config, resources, workflowResourceReferences, currentResourceUsage,
+        eligibleInstances);
 
     updateStats(resources, currentResourceUsage);
   }
 
+  private void limitAndDequeue(
+      StyxConfig config,
```
honnix: It doesn't feel very necessary to pass in the whole `StyxConfig` object, a boolean would just work.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -159,17 +173,44 @@ void tick() {
...
+      final Map<WorkflowInstance, CompletionStage<Optional<ExecutionBlocker>>> executionBlockers =
+          batch.stream()
+              .map(InstanceState::workflowInstance)
+              .collect(toMap(identity(), gate::executionBlocker));
+
+      // Evaluate each instance in the batch for dequeuing
+      for (InstanceState instance : batch) {
+        final boolean proceed = limitAndDequeue(
+            resources, workflowResourceReferences, currentResourceUsage, instance,
+            executionBlockers.getOrDefault(instance.workflowInstance(), NO_BLOCKER));
```
honnix: This map is built by iterating through the batch, in what case we might get the default? It doesn't seem necessary to use a map. Do I misunderstand something?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -159,17 +173,44 @@ void tick() {
...
+
+        // Stop processing if rate limit was hit
```
honnix: First of all this comment is now lying. Second, if one blocker execution is interrupted, are we sure we want to stop processing everything left? I guess in many cases interruption would mean the executor service is shutting down.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -191,7 +232,28 @@ private void updateStats(Map<String, Resource> resources,
...
   private boolean limitAndDequeue(Map<String, Resource> resources,
       Map<WorkflowId, Set<String>> workflowResourceReferences,
-      Map<String, Long> currentResourceUsage, InstanceState instance) {
+      Map<String, Long> currentResourceUsage, InstanceState instance,
+      CompletionStage<Optional<ExecutionBlocker>> executionBlockerFuture) {
+
+    // Check for execution blockers
+    Optional<ExecutionBlocker> blocker = Optional.empty();
+    try {
+      blocker = executionBlockerFuture.toCompletableFuture().get(30, TimeUnit.SECONDS);
```
honnix: This would cause delay if somehow blocker execution is slow? e.g. bad network, external service slowness, corner cases, etc. Is 30s too much?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -191,7 +232,28 @@ private void updateStats(Map<String, Resource> resources,
 
   private boolean limitAndDequeue(Map<String, Resource> resources,
       Map<WorkflowId, Set<String>> workflowResourceReferences,
-      Map<String, Long> currentResourceUsage, InstanceState instance) {
+      Map<String, Long> currentResourceUsage, InstanceState instance,
+      CompletionStage<Optional<ExecutionBlocker>> executionBlockerFuture) {
+
+    // Check for execution blockers
```
honnix: `s/blockers/blocker`

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -191,7 +232,28 @@ private void updateStats(Map<String, Resource> resources,
...
+    Optional<ExecutionBlocker> blocker = Optional.empty();
+    try {
+      blocker = executionBlockerFuture.toCompletableFuture().get(30, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      LOG.debug("Interrupted");
+      return false;
+    } catch (ExecutionException | TimeoutException e) {
+      LOG.warn("Failed to check execution blockers for {}", instance.workflowInstance(), e);
```
honnix: `s/blockers/blocker`, also I think the log can be clearer that we will continue as if there were no blocker.

styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowExecutionGate.java
```diff
@@ -0,0 +1,81 @@
...
+  WorkflowExecutionGate NOOP = wfi -> NO_BLOCKER;
+
+  /**
+   * Check if there is a blocker for the execution of a workflow instance. This method is called
+   * before the workflow instance is dequeued. If an {@link ExecutionBlocker} is returned, execution
+   * of the workflow instance is delayed. Implementations of this method should be careful to
+   * consider whether information gathered e.g. from previous executions has been invalidated by
+   * docker image and/or args changes, etc.
+   *
+   * <p>This method must not block.
```
honnix: ⭐️

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: I think we can hold off on #360 for now.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: The reason for storing the blockers in a separate collection is to look them up in parallel for the entire batch.

Perhaps I would have used a list instead of a map if java offered a nice way of iterating over two lists. E.g. zipping. But alas.

An alternative here would be to change this for loop into an old-school for loop with an integer counter.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: Yeah I think we want to stop processing if the thread was interrupted.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: What timeout would you prefer?

Note that the same is true where the scheduler calls `stateManager.receiveIgnoreClosed`. The scheduler might block for a long time for each workflow instance if the network is bad, the blocking datastore requests take a long time to complete, etc.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
honnix: Ok old school was the also the alternative coming to my mind. ;)

Anyway getOrDefault doesn’t seem to be necessary and can be confusing when reading the code after a while.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
honnix: Dunno :D How long would it require in our internal use case?
It is a little bit different comparing with communicating with datastore. The plugin can do arbitrary stuff and the platform has no control of that.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
honnix: +1

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: Went for old-school for loop, feels cleaner 👍

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: Doesn't feel very necessary to pull out an individual field either... =)

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: I think `30` seconds is a reasonable timeout.

And I think it is fair that a scheduler plugin can cause problems for the scheduler. It can be disabled by using the `executionGatingEnabled` runtime config flag if necessary.

honnix: Fair enough. A little bit documentation for `executionBlocker` method if you feel that's necessary?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
honnix: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
honnix: This was mostly a nitpicking, so 👍

honnix:
Awesome now! 👍

PR #292, opened 2017-10-05
storage & docker operation error metrics

danielnorberg:
* Also count failed operations
* Qualify operation `status` as either `success` or `error`
* Also emit detailed docker/k8s error meters with error status code

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MeteredProxy.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MeteredProxy.java
@@ -23,6 +23,8 @@ package com.spotify.styx.monitoring;
 import com.spotify.styx.docker.DockerRunner;
 import com.spotify.styx.storage.Storage;
 import com.spotify.styx.util.Time;
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import io.fabric8.kubernetes.client.KubernetesClientTimeoutException;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
@@ -30,6 +32,7 @@ import java.lang.reflect.Proxy;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Objects;
+import javaslang.control.Try;
 
 /**
  * A proxy for instrumenting an instance of {@link Storage} or {@link DockerRunner} using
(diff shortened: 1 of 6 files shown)


danielnorberg:
@ulzha PTAL

danielnorberg:
@NatashaL PTAL

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MeteredProxy.java
```diff
@@ -23,13 +23,16 @@
 import com.spotify.styx.docker.DockerRunner;
 import com.spotify.styx.storage.Storage;
 import com.spotify.styx.util.Time;
+import io.fabric8.kubernetes.client.KubernetesClientException;
```
honnix: This kind of dependencies feels weird to be in the proxy. But don't have a better solution. Maybe we can have a separated class handling reporting of failed operations (not limited to k8s in the future).

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MeteredProxy.java
```diff
@@ -62,20 +65,59 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
+        throw e.getTargetException();
+      }
+    });
+
     final long durationMillis = t0.until(time.get(), ChronoUnit.MILLIS);
 
+    // TODO: Instrument at a lower level to catch errors with greater granularity. At this level we might not see
+    //       all errors as the implementation might swallow exceptions, perform retries, etc.
+    result.onFailure(error -> {
+      if (delegate instanceof DockerRunner) {
```
honnix: Move this to L90? We already check `instaceof` there. Or if we have a separated helper class handling reporting failure, we can pass `delegate` to that class and do all the dirty check there. So instead of `reportDockerOperationError` we can do `reportOperationError`.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MeteredProxy.java
danielnorberg: Yeah, I would rather wrap the k8s client directly, e.g. in the `KubernetesDockerRunner` but I wanted to get something in place quickly and the `MeteredProxy` seemed like a convenient place to quickly get some metrics into place with a minimal amount of code.

honnix: If the TODO will be addressed later (not staying forever:), ok for me.

honnix:
+1 ship it so we can capture data for a few hours.

styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MeteredProxy.java
danielnorberg: Yeah, I think I'd like to maybe work on moving this to `KubernetesDockerRunner`.

PR #293, opened 2017-10-06
k8s restore: retry harder

danielnorberg:


honnix:
👍

ulzha:
+1. Could only perhaps improve with some diagnostic logging onRetry? Trying hard and not telling the world makes the operator confused, like, what is this program doing for so long...

danielnorberg:
@ulzha agreed

danielnorberg:
@ulzha now pod polling failures during startup are logged, good enough?

PR #294, opened 2017-10-06
[Hot-Fix] Handle GKE nodes becoming unresponsive

fabriziodemaria:


honnix:
We might get lots of false positive. So hold on.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesPodEventTranslator.java
```diff
@@ -226,6 +226,12 @@ public TerminationLogMessage(
   private static Optional<Event> isInErrorState(WorkflowInstance workflowInstance, Pod pod) {
     final PodStatus status = pod.getStatus();
     final String phase = status.getPhase();
+    final String reason = status.getReason();
+
+    if (reason != null && "NodeLost".equals(reason)) {
```
bergman: You write the equals in that backwards way to get rid of null checks 😄

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesPodEventTranslatorTest.java
```diff
@@ -130,6 +130,16 @@ public void runErrorOnUnexpectedTerminatedStatus() throws Exception {
   }
 
   @Test
+  public void runErrorOnLostNode() throws Exception {
+    pod.setStatus(running(true));
+    pod.getStatus().setReason("NodeLost");
+
+    assertGeneratesEventsAndTransitions(
+        RunState.State.RUNNING, pod,
+        Event.runError(WFI, "Node became unresponsive"));
```
bergman: The code seems to return `Kubernetes node became unresponsive` and not `Node became unresponsive`?

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesPodEventTranslator.java
fabriziodemaria: 👍 :)

bergman:
👍

fabriziodemaria:
In our production systems we found Kubernetes nodes that were unresponsive (`NotReady` state) for some time (minutes) and then went back to `Ready` state. Checking the node `state` might indeed cause a lot of false positive as @honnix pointed out. I think we should close this.

PR #295, opened 2017-10-13
[HOLD] workflow container memory requests & limits

danielnorberg:
Memory requests & limits are useful in k8s clusters to limit the impact of non well-behaved containers on the node and other containers.

* Configure k8s memory requests and limits for WFI containers
* Default to `512Mi` request and `1024Mi` limit
* TODO: make global default mem req/lims runtime configurable

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/model/ExecutionDescription.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/ExecutionDescription.java
@@ -20,52 +20,45 @@
 
 package com.spotify.styx.model;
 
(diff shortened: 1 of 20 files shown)


danielnorberg:
Perhaps just have a `JsonNode docker_resources` field in `WorkflowConfiguration` that gets propagated as-is to the k8s container.resources field? 🤔 Avoid a lot of tedious code mirroring.

honnix:
Time to rethink this? We didn't reach any agreement last time.

PR #296, opened 2017-10-13
remove PATCH on component level

bergman:


danielnorberg:
👍 if this is ready to be merge, release and deploy

PR #298, opened 2017-10-16
restore workflow crud api

danielnorberg:
* This reverts PR #255 (3aac3da) and reintroduces workflow CRUD REST API endpoints.

* The Styx workflow storage is now the source of truth.

* Integrations with CD platform and ingestion from external truth sources can be performed directly using the workflow CRUD REST API endpoints.

* Unavailability of CD platforms and external truth sources does not affect the workflows that are stored by Styx itself.

### DIFF
--- a/doc/api.apib
+++ b/doc/api.apib
@@ -104,14 +104,14 @@ Resources related to the Workflows registered in Styx.
 
 ### Delete Workflow [DELETE]
 
-Returns 200 OK if the Workflow was found an deleted, 404 Not Found otherwise.
+Returns 204 No Content if the Workflow was found and deleted, 404 Not Found otherwise.
 
 + Parameters
     + version: `v3` (enum[string]) - API version
         + Members
             + `v3`
 
-+ Response 200
++ Response 204
 + Response 404
 
 ## Workflows [/{version}/workflows/{component_id}]
--- a/doc/api.apib
+++ b/doc/api.apib
@@ -102,6 +102,75 @@ Resources related to the Workflows registered in Styx.
               }
         }
 
+### Delete Workflow [DELETE]
+
+Returns 200 OK if the Workflow was found an deleted, 404 Not Found otherwise.
+
++ Parameters
+    + version: `v3` (enum[string]) - API version
+        + Members
+            + `v3`
+
++ Response 200
++ Response 404
+
+## Workflows [/{version}/workflows/{component_id}]
+
+Accepts a Workflow definition as a JSON object. Returns 200 OK if the Workflow was registered
+or updated. Note that a non-null value is required for docker_image.
+
++ Parameters
+    + version: `v3` (enum[string]) - API version
+        + Members
+            + `v3`
+    + component_id: `styx-canary` (string) - Workflow Component
+
+### Create or update Workflow [POST]
+
++ Request (application/json)
+
+        {
+          "id": "StyxCanary",
+          "schedule": "days",
+          "offset": null,
+          "docker_image": "styx-canary-dummy:dummy",
+          "docker_args": [
+            "luigi",
+            "--module",
+            "canary_job",
+            "CanaryJob"
+          ],
+          "docker_termination_logging": false,
+          "secret": null,
+          "service_account": null,
+          "resources": []
+        }
+
++ Response 200 (application/json)
+
+        {
+          "component_id": "styx-canary",
+          "workflow_id": "StyxCanary",
+          "component_uri": "file:///etc/styx/schedule.yaml",
+          "configuration": {
+            "id": "StyxCanary",
+            "schedule": "days",
+            "offset": null,
+            "docker_image": "styx-canary-dummy:dummy",
+            "docker_args": [
+              "luigi",
+              "--module",
+              "canary_job",
+              "CanaryJob"
+            ],
+            "docker_termination_logging": false,
+            "secret": null,
+            "service_account": null,
+            "resources": []
+          },
+          "__from_api": "V3"
+        }
+
 ## Workflow Instances [/{version}/workflows/{component_id}/{workflow_id}/instances{?offset,limit,start,stop}]
 
 Query can be done in two different styles: `offset+limit` or `start+stop`.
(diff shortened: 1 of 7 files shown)


doc/api.apib
```diff
@@ -102,6 +102,75 @@ Resources related to the Workflows registered in Styx.
               }
         }
 
+### Delete Workflow [DELETE]
+
+Returns 200 OK if the Workflow was found an deleted, 404 Not Found otherwise.
```
honnix: There is now `204`.

danielnorberg: good catch!

bergman:
Looks good, but what about validation of input? I expected some `dockerImageValidator` calls in the POST too, not only the PATCH.

danielnorberg:
@bergman good catch!

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -62,10 +67,14 @@
   public SchedulerResource(
       StateManager stateManager,
       TriggerListener triggerListener,
+      Consumer<Workflow> workflowChangeListener,
+      Consumer<Workflow> workflowRemoveListener,
       Storage storage,
       Time time) {
     this.stateManager = Objects.requireNonNull(stateManager);
     this.triggerListener = Objects.requireNonNull(triggerListener);
+    this.workflowChangeListener = workflowChangeListener;
```
honnix: missing null check.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -86,11 +94,61 @@ public WorkflowResource(Storage storage, DockerImageValidator dockerImageValidat
...
+      return CompletableFuture.completedFuture(
+          Response.forStatus(Status.BAD_REQUEST.withReasonPhrase("Missing payload.")));
+    }
+    final WorkflowConfiguration workflowConfig;
+    try {
+      workflowConfig = OBJECT_MAPPER
+          .readValue(payload.get().toByteArray(), WorkflowConfiguration.class);
+    } catch (IOException e) {
+      return CompletableFuture.completedFuture(
+          Response.forStatus(Status.BAD_REQUEST.withReasonPhrase("Invalid payload.")));
```
honnix: We can improve the error message, I think. Instead of `Invalid payload` we can be more specific, like `Invalid workflow configuration` for example.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -86,11 +94,61 @@ public WorkflowResource(Storage storage, DockerImageValidator dockerImageValidat
...
+    }
+
+    final Workflow workflow = Workflow.create(componentId, workflowConfig);
+
+    final ByteString requestPayload;
+    try {
+      requestPayload = serialize(workflow);
+    } catch (JsonProcessingException e) {
+      return CompletableFuture.completedFuture(Response.forStatus(
+          Status.INTERNAL_SERVER_ERROR.withReasonPhrase("Failed to serialize proxy payload.")));
```
honnix: Let's get away from the `proxy` since it is completely internal to how we implement it.

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -82,11 +91,42 @@ public SchedulerResource(
...
+    }
+    if (!workflowOpt.isPresent()) {
+      return Response.forStatus(Status.NOT_FOUND.withReasonPhrase("Workflow does not exist"));
+    }
+    workflowRemoveListener.accept(workflowOpt.get());
+    return Response.forStatus(Status.NO_CONTENT);
+  }
+
+  private Response<Workflow> createOrUpdateWorkflow(Workflow workflow) {
+    workflowChangeListener.accept(workflow);
```
honnix: It's better to have a big `TODO` here: validate the workflow and throw something more reasonable than 500 if `accept` failed with `RuntimeException`. This can be address in a different PR to keep this one clean.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
danielnorberg: Might be completely unparsable payload though... 🤔

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
danielnorberg: lol how did this go through the first PR

danielnorberg: also why is it re-serializing instead of just passing on inputs...

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
honnix: We need a flaky network for this to happen...

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
honnix: not sure. I thought we fixed this.

danielnorberg:
Btw, the use of `POST` for upsert is a bit unusual. `PUT` is usually used for that. Do we care?

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
danielnorberg: We should validate the workflow as much as possible, and return 4XX for bad requests, but if something fails after validation I think that's an internal error that should be surfaced as a 500.

honnix: There was a corner case that invalid workflows are surfaced later when parsing cron syntax to get next natural trigger. We can address that in a separated PR.

honnix:
If using PUT I think we will need to include workflow Id in the url.

honnix:
Maybe not worth it.

danielnorberg:
Yeah, PUT would entail having workflow id in the url. I'm fine with POST for now though. If we want we can change it later.

danielnorberg:
Good to merge?

honnix:
Sorry for being late. Sure. 👍

PR #300, opened 2017-10-16
cli: workflow create & delete

danielnorberg:
Usage: 

**create**
```sh
# From file
styx workflow create <cid> -f workflow.yaml

# From stdin
cat workflow.yaml | styx workflow create <cid>
cat workflow.yaml | styx workflow create <cid> -f -
```

*workflow.yaml*: 
```yaml
id: bar
schedule: daily
service_account: foo@bar.com
docker_image: busybox
docker_args: 
  - echo
  - hello world
```

**delete** 
```
styx workflow delete <cid> <wfid>
```

### DIFF
--- a/styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
+++ b/styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
@@ -20,6 +20,13 @@
 
 package com.spotify.styx.cli;
 
+import static com.spotify.apollo.Status.NOT_FOUND;
+import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toList;
+import static net.sourceforge.argparse4j.impl.Arguments.fileType;
+
+import com.fasterxml.jackson.databind.MappingIterator;
+import com.fasterxml.jackson.databind.ObjectReader;
 import com.google.common.collect.ImmutableList;
 import com.spotify.apollo.Client;
 import com.spotify.apollo.core.Service;
@@ -38,9 +45,13 @@ import com.spotify.styx.client.StyxClientFactory;
 import com.spotify.styx.model.Backfill;
 import com.spotify.styx.model.Resource;
 import com.spotify.styx.model.Workflow;
+import com.spotify.styx.model.WorkflowConfiguration;
 import com.spotify.styx.model.WorkflowState;
 import com.spotify.styx.model.data.EventInfo;
+import com.spotify.styx.serialization.Json;
 import com.spotify.styx.util.ParameterUtil;
+import java.io.File;
+import java.io.IOException;
 import java.time.format.DateTimeParseException;
 import java.util.Collection;
 import java.util.Collections;
@@ -49,6 +60,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import javaslang.Tuple;
 import javaslang.Tuple2;
@@ -238,6 +250,12 @@ public final class CliMain {
             case SHOW:
               workflowShow();
               break;
+            case CREATE:
+              workflowCreate();
+              break;
+            case DELETE:
+              workflowDelete();
+              break;
             default:
               // parsing unknown command will fail so this would only catch non-exhaustive switches
               throw new ArgumentParserException(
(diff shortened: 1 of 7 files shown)


danielnorberg:
The structure for the workflow input is a bit clunky:

```

**Comment by honnix** [comment]
still trying to figure how we would use it internally, given a sysmodel file.

also we seem to support multiple workflows in one file: https://github.com/spotify/styx/blob/master/styx-local-files-source/src/main/java/com/spotify/styx/LocalFileScheduleSource.java#L206 , does it make sense to keep the consistency? (though schedule source will be gone soon)

**Comment by danielnorberg** [comment]
Yeah, supporting multiple workflows in one file seems useful.

It also seems useful to allow a single workflow in one file. 

How should this work? Maybe something like this?

```
# Creates a single workflow
styx workflow create -f workflow.yaml

# Creates multiple workflows.
styx create -f workflows.yaml
```

*workflows.yaml*:
```
workflows:
  - id: foo
    ...
  - id: bar
    ...
```

Note that `workflows` is not included in the command line in the second case. Instead that identifier is in `workflows.yaml`. This allows us to explicitly separate between the case of creating a single workflow and multiple workflows. It also gives us the flexibility of extending and allowing other entity types (e.g. resources) in the file in the future.

This would also allow us to support the existing "schedules" file almost as-is, just change `schedules:` to `workflows:` and tell the user to `styx create -f` instead of putting it on the styx scheduler host and wait for it to be picked up.

Wdyt?

**Comment by honnix** [comment]
LGTM

File: styx-cli/src/main/java/com/spotify/styx/cli/Main.java [hunk 1]
```diff
@@ -252,6 +265,41 @@ private void run() {
...
+    final String component = namespace.getString(parser.workflowDeleteComponentId.getDest());
+    final String workflow = namespace.getString(parser.workflowDeleteWorkflowId.getDest());
+    final boolean force = namespace.getBoolean(parser.workflowDeleteForce.getDest());
+
+    if (System.console() != null && !force) {
+      final String reply = System.console().readLine(
+          "Sure you want to delete the workflow " + component + " " + workflow + " ? [yN]").trim();
+      if (!reply.equals("y")) {
+        System.exit(1);
+        return;
```
honnix: unreachable code.

bergman: iirc you have to do this to get rid of some warning, like intellij doesn't understand that `System.exit()` will not return

styx-cli/src/main/java/com/spotify/styx/cli/Main.java
```diff
@@ -489,6 +537,25 @@ private void retryWorkflowInstance() throws ExecutionException, InterruptedExcep
...
+            .type(fileType().acceptSystemIn().verifyCanRead())
+            .help("Workflow configuration file");
+
+    final Subparser workflowDelete = WorkflowCommand.DELETE.parser(workflowParser);
+    final Argument workflowDeleteComponentId =
+        workflowDelete.addArgument("component").help("Component ID");
+    final Argument workflowDeleteWorkflowId =
+        workflowDelete.addArgument("workflow").help("Workflow ID");
+    final Argument workflowDeleteForce =
+        workflowDelete.addArgument("-f", "--force")
```
bergman: `-f` means "force" in deletion but "file" in create? I'd be fine with having to type out `--force` here.

danielnorberg: 👍

danielnorberg:
Another simpler solution would be to allow for multiple workflows in one yaml separated by `---`, i.e. like k8s: https://kubernetes.io/docs/concepts/cluster-administration/manage-deployment/#organizing-resource-configurations

danielnorberg:
The cli now supports creating and deleting multiple workflows in one invocation:

**workflows.yaml:**
```
id: foo
schedule: daily
---
id: bar
schedule: daily
...
---
...
```

JSON is also supported.

**workflows.json:**
```javascript
[
  {
    "id": "foo",
    "schedule": "daily",
    // ...
  },
  {
    "id": "bar",
    "schedule": "daily",
    // ...
  }
]
```

**create:**
```
$ styx workflow create quux -f workflows.yaml
Workflow foo in component quux created.
Workflow bar in component quux created.
```

**delete**:
```
$ styx workflow delete quux foo baz bar
Workflow foo in component quux deleted.
Workflow baz in component quux not found.
Workflow bar in component quux deleted.
```

danielnorberg:
TODO: add tests for these commands

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -269,6 +287,76 @@ private void run(CliContext cliContext) {
...
+        return;
+      }
+    }
+
+    final List<Tuple2<String, CompletionStage<Void>>> futures = workflows.stream()
+        .map(workflow -> Tuple.of(workflow, styxClient.deleteWorkflow(component, workflow)))
+        .collect(toList());
+
+    for (Tuple2<String, CompletionStage<Void>> future : futures) {
+      String workflow = future._1;
```
honnix: final

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -269,6 +287,76 @@ private void run(CliContext cliContext) {
...
+      try {
+        future._2.toCompletableFuture().get();
+        cliOutput.printMessage("Workflow " + workflow + " in component " + component + " deleted.");
+      } catch (ExecutionException e) {
+        if (e.getCause() instanceof ApiErrorException) {
+          ApiErrorException apiError = (ApiErrorException) e.getCause();
```
honnix: final

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -269,6 +287,76 @@ private void run(CliContext cliContext) {
...
+
+    // TODO: allow reading workflow ID's from file?
+
+    if (System.console() != null && !force) {
+      final String reply = System.console().readLine(
+          "Sure you want to delete the workflow" + (workflows.size() > 1 ? "s" : "") + " "
+              + workflows.stream().collect(joining(", "))
+              + "in component " + component + "? [yN]").trim();
+      if (!reply.equals("y")) {
+        System.exit(1);
```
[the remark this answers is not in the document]
danielnorberg: hmm rebase fail?

honnix:
+1

bergman:
The delete command is kind of weird that the first argument is component and the rest are workflow id's. I feel it may be easy to make mistakes? Maybe not.

danielnorberg:
@bergman Yeah, I agree that it might be easy to make mistakes. Hence the deletion confirmation prompt that clearly states what will be deleted and defaults to `no`.

Alternatives I guess could look like this, not sure if it is much better.

```
styx workflow create -c quux -f workflows.yaml
styx workflow delete -c quux bar baz
```

Any other ideas?

honnix:
If we want to make it super safe, `-c` for component id then.

danielnorberg:
@honnix is `styx workflow delete -c quux bar baz` that much clearer?

bergman:
Hmm, that's almost worse, it makes it seem like it's optional to supply component. I'd go for first version. Would be funny if someone has a workflow named the same as a component and forgot to specify their own component. Whaaaa?

PR #301, opened 2017-10-16
client: workflow crud methods

danielnorberg:


styx-cli/src/test/java/com/spotify/styx/client/StyxApolloClientTest.java
```diff
@@ -20,19 +20,30 @@
...
+import com.spotify.apollo.Status;
+import com.spotify.styx.api.Api;
+import com.spotify.styx.model.Schedule;
+import com.spotify.styx.model.Workflow;
+import com.spotify.styx.model.WorkflowConfiguration;
+import com.spotify.styx.serialization.Json;
 import java.util.concurrent.CompletableFuture;
 import junitparams.JUnitParamsRunner;
 import junitparams.Parameters;
+import okhttp3.HttpUrl;
```
bergman: com.squareup.okhttp.HttpUrl

bergman: Not sure what the difference is...

honnix: We have both okhttp and okhttp3 as dependency. :O

honnix:
I don't see why we had to introduce okhttp3. I verified styx-cli still works using okhttp. Let's remove `okhttp3` for now?

honnix:
https://github.com/spotify/styx/pull/302

danielnorberg:
Having only one okhttp dep sgtm =)

honnix:
👍

PR #302, opened 2017-10-16
use 2 series

honnix:
@fabriziodemaria @danielnorberg Let's take it as a separated PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -200,13 +200,6 @@
         <version>1.7.25</version>
       </dependency>
 
-      <!-- styx client libs -->
-      <dependency>
-        <groupId>com.squareup.okhttp3</groupId>
-        <artifactId>okhttp</artifactId>
-        <version>3.7.0</version>
-      </dependency>
-
       <!-- external clients libs -->
       <dependency>
         <groupId>com.spotify</groupId>
--- a/styx-client/pom.xml
+++ b/styx-client/pom.xml
@@ -18,7 +18,7 @@
     </dependency>
 
     <dependency>
-      <groupId>com.squareup.okhttp3</groupId>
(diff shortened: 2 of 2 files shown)


danielnorberg:
okhttp 2 is obsolete tho: https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-300-rc1

> This release obsoletes OkHttp 2.x, and all code that uses OkHttp's com.squareup.okhttp package should upgrade to the okhttp3 package. Libraries that depend on OkHttp should upgrade quickly to prevent applications from being stuck on the old version.

honnix:
Yeah unfortunately. But we probably want to go with Apollo.

danielnorberg:
We should submit a PR to apollo to upgrade to okhttp3 as well I guess.

honnix:
@danielnorberg feel free to merge this if it's OK.

honnix:
https://github.com/spotify/apollo/pull/220 :D

danielnorberg:
@honnix thanks =)

honnix:
OK to merge for now? This jar is now included for nothing.

danielnorberg:
But then we end up with an explicit dependency on the obsolete okhttp...

honnix:
That's true. I think there is little chance that Apollo 1.x will upgrade to okhttp3 https://github.com/spotify/apollo/pull/224

If we want to use okhttp3 explicitly here, let's fix the imported packages.

honnix:
https://github.com/spotify/styx/pull/316

danielnorberg:
I interpreted https://github.com/spotify/apollo/pull/220#issuecomment-338992850 as upgrading to okhttp3 being positively considered for Apollo 1.x as well

honnix:
in favor of #316

PR #304, opened 2017-10-18
scoped client cannot be used when nginx is in front of scheduler

honnix:
scoped client would add another Authorization header and that makes
nginx sad.

@bergman PTAL


bergman:
👍

danielnorberg:
👍 good catch 😅

PR #308, opened 2017-10-24
add commitSha to workflow configuration

bergman:
- PATCH state will also update workflow configuration
 - POST workflow will also update workflow state


NatashaL:
👍

PR #312, opened 2017-10-26
Remove docker_image and commit_sha from api docs

bergman:


honnix:
Not sure we want to return docker_image and commit_sha though. But that's gonna be another PR. So 👍

PR #313, opened 2017-10-26
add /workflows endpoint to api doc

bergman:


doc/api.apib
```diff
@@ -30,13 +30,44 @@ To know more details, please check
 # Group Workflows
 
 Resources related to the Workflows registered in Styx.
+## Workflows [/{version}/workflows]
+
+### Get Workflows [GET]
+
++ Parameters
+    + version: `v3` (enum[string]) - API version
```
fabriziodemaria: We could move this up one layer to `## Workflows [/{version}/workflows]`

PR #315, opened 2017-10-26
remove dockerImage and commitSha from WF State

bergman:


bergman:
closing in favor of #318

PR #317, opened 2017-10-27
Make workflow json be stored as long unindexed text

NatashaL:
There's no way to store the property as a [`Text`](https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/Text) object per se. But it seems that Text is just a wrapper around unindexed strings, so this should accomplish the same thing.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -33,6 +33,7 @@ import com.google.cloud.datastore.KeyFactory;
 import com.google.cloud.datastore.PathElement;
 import com.google.cloud.datastore.Query;
 import com.google.cloud.datastore.QueryResults;
+import com.google.cloud.datastore.StringValue;
 import com.google.cloud.datastore.StructuredQuery.CompositeFilter;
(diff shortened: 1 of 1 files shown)


honnix:
Why not

```Consider using a standard Java string and storing it with the PropertyContainer.setUnindexedProperty(java.lang.String, java.lang.Object) method instead.```

NatashaL:
`Text`, as well as `PropertyContainer` are part of the `com.google.appengine` libraries and those, along with `com.google.appengine.api.datastore.Entity` are not compatible with `com.google.cloud.(datastore.Entity)` . 

The reading part should be compatible for both, since the size only affects indexing on the server side.

honnix:
How about the recommended way I mentioned above?

NatashaL:
That's the `PropertyContainer` I mentioned that is part of the appengine libraries and is not compatible with the google cloud libraries.

honnix:
Sorry I missed that. Let’s talk about it tmr.

honnix:
Can we somehow verify this against real datastore instead of the emulator?

NatashaL:
Yeah, it's verifiable if you test it against a real datastore instance. The reading part is identical for both reading indexed and unindexed fields.

honnix:
👍

PR #318, opened 2017-10-31
Remove dockerImage and commitSha from WorkflowState

bergman:
~~this is not as harsh as #315 , keeping wf state for reading still but populating from wf configuration~~

this is just as harsh now


styx-scheduler-service/src/main/java/com/spotify/styx/StateInitializingTrigger.java
```diff
@@ -49,17 +48,15 @@
...
-  StateInitializingTrigger(StateFactory stateFactory, StateManager stateManager, Storage storage) {
+  StateInitializingTrigger(StateFactory stateFactory, StateManager stateManager) {
     this.stateFactory = Objects.requireNonNull(stateFactory);
     this.stateManager = Objects.requireNonNull(stateManager);
-    this.storage = Objects.requireNonNull(storage);
   }
 
   @Override
   public CompletionStage<Void> event(Workflow workflow, Trigger trigger, Instant instant) {
-    if (!WorkflowValidator.hasDockerConfiguration(workflow, storage)) {
```
fabriziodemaria: We now solely rely on the Workflow object and never check Storage at this stage.
We can see that the Workflow object passed here from various places in the code is created by either reading Storage (from `SchedulerResource` and `TriggerManager`) or just retrieved from WorkflowCache (from `BackfillTriggerManager`). Storage and WorkflowCache _should_ be in sync if data is not manually changed in Datastore.

With this comment I wanted to point out that the `hasDockerConfiguration` check slightly changes now for the `BackfillTriggerManager` case. We should consolidate or remove WorkflowCache (in a different PR of course).

styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
```diff
@@ -203,61 +187,44 @@ public void shouldSucceedWithFullPatchStatePerWorkflow() throws Exception {
...
     sinceVersion(Api.Version.V3);
 
-    storage.patchState(WORKFLOW.id(), patchDockerImage("preset:image"));
-
     Response<ByteString> response =
         awaitResponse(serviceHelper.request("PATCH", path("/foo/bar/state"),
                                             STATEPAYLOAD_OTHER_FIELD));
 
     assertThat(response, hasStatus(withCode(Status.OK)));
-    assertThat(response, hasHeader("Content-Type", equalTo("application/json")));
```
fabriziodemaria: Did you mean to remove this assertion?

styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -452,15 +372,15 @@ public void allFieldsAreSetWhenRetrievingWorkflowState() throws Exception {
...
+            .commitSha(COMMIT_SHA)
             .build()));
     Instant instant = Instant.parse("2016-03-14T14:00:00Z");
     Instant offset = instant.plus(1, ChronoUnit.DAYS);
     TriggerInstantSpec spec = TriggerInstantSpec.create(instant, offset);
     storage.updateNextNaturalTrigger(WORKFLOW_ID1, spec);
     WorkflowState state = WorkflowState.builder()
         .enabled(true)
-        .dockerImage(DOCKER_IMAGE.get())
-        .commitSha(COMMIT_SHA)
```
fabriziodemaria: We should also check that the `Workflow` returned from Storage is the same as the one constructed above (so we also verify that dockerImage and commitSha have been set properly by `store` operation).

styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -182,97 +179,20 @@ public void shouldPersistNextScheduledRun() throws Exception {
   }
 
   @Test
-  public void shouldNotRemoveWorkflowWhenSettingDockerImage() throws Exception {
```
fabriziodemaria: I don't see a test where a Workflow is stored once, and then a second time with just a modified field (i.e. the Docker image), to then check that `storage.workflow(...)` does return the updated version.

styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
bergman: it just didn't seem necessary, there are other checks for that. i can add back if you want.

styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
bergman: that is not a valid case. everything should be overwritten.

styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
bergman: not sure about that, there are other tests that already do that

styx-scheduler-service/src/main/java/com/spotify/styx/StateInitializingTrigger.java
bergman: agreed, we should rework the workflow cache soon

styx-scheduler-service/src/test/java/com/spotify/styx/StateInitializingTriggerTest.java
```diff
@@ -111,8 +104,11 @@ public void shouldInjectTriggerExecutionEventWithBackfillTrigger() throws Except
 
   @Test
   public void shouldDoNothingIfDockerInfoMissing() throws Exception {
-    Workflow workflow = Workflow.create("id", TestData.DAILY_WORKFLOW_CONFIGURATION);
-    setDockerImage(workflow.id(), workflow.configuration());
+    final WorkflowConfiguration configuration =
```
fabriziodemaria: This test was checking that nothing is done when docker args are missing, while now it checks that nothing is done when docker image is missing. Let's rename this `shouldDoNothingIfDockerImageMissing`. Maybe it's good to add a test case that tests initialization succeed even with no docker args provided.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/WorkflowValidator.java
```diff
@@ -37,19 +32,6 @@ private WorkflowValidator() {
   }
 
   public static boolean hasDockerConfiguration(Workflow workflow) {
-    return workflow.configuration().dockerArgs().isPresent();
-  }
-
-  public static boolean hasDockerConfiguration(Workflow workflow, Storage storage) {
-    final Optional<List<String>> dockerArgs = workflow.configuration().dockerArgs();
-    if (!dockerArgs.isPresent()) {
```
fabriziodemaria: I believe this was an intentional removal to allow scheduling docker containers with no docker args?

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandlerTest.java
```diff
@@ -116,12 +116,13 @@ public void shouldTransitionIntoSubmittingIfMissingDockerArgs() throws Exception
...
-    Workflow workflow = Workflow.create("id", schedule("--date", "{}", "--bar"));
-    WorkflowState workflowState = WorkflowState.builder()
-        .enabled(true)
-        .dockerImage(DOCKER_IMAGE)
-        .commitSha(COMMIT_SHA)
-        .build();
+    final WorkflowConfiguration workflowConfiguration =
+        WorkflowConfigurationBuilder.from(schedule("--date", "{}", "--bar"))
+            .commitSha(COMMIT_SHA)
+            .dockerImage(DOCKER_IMAGE)
```
fabriziodemaria: Two fields above are redundant now since set by `schedule()`

styx-scheduler-service/src/main/java/com/spotify/styx/docker/WorkflowValidator.java
bergman: yes

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandlerTest.java
bergman: right, thanks!

fabriziodemaria:
`shouldFallbackToDockerImageInScheduleDefinition` test in `ExecutionDescriptionHandlerTest` might not have much sense anymore.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -416,8 +413,6 @@ void patchState(WorkflowId workflowId, WorkflowState state) throws IOException {
 
       final Entity.Builder builder = Entity.newBuilder(workflowOpt.get());
       state.enabled().ifPresent(x -> builder.set(PROPERTY_WORKFLOW_ENABLED, x));
-      state.dockerImage().ifPresent(x -> builder.set(PROPERTY_DOCKER_IMAGE, x));
-      state.commitSha().ifPresent(x -> builder.set(PROPERTY_COMMIT_SHA, x));
```
fabriziodemaria: Constants are not used anymore, can be removed from declarations

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -426,24 +421,6 @@ void patchState(WorkflowId workflowId, WorkflowState state) throws IOException {
...
   }
 
-  Optional<String> getDockerImage(WorkflowId workflowId) throws IOException {
-    final Optional<Entity> workflowEntity = getOpt(datastore, workflowKey(workflowId));
-    Optional<String> dockerImage = getOptStringProperty(workflowEntity, PROPERTY_DOCKER_IMAGE);
-    if (dockerImage.isPresent()) {
-      return dockerImage;
-    }
-
-    dockerImage = getOptStringProperty(datastore,
```
fabriziodemaria: Two signatures of this method are not used anymore, we could remove them.

fabriziodemaria:
Minor comments on test modifications. Overall LGTM.

bergman:
rebased

fabriziodemaria:
👍

PR #319, opened 2017-11-02
CLI should use wf conf for docker image and commit sha

bergman:


bergman:
closing in favour of #318

PR #320, opened 2017-11-02
Add stats to eventConsumerFactory

fabriziodemaria:


styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -111,6 +113,10 @@
       .tagged("what", "submission-rate-limit")
       .tagged("unit", "submission/s");
 
+  private static final MetricId EVENT_CONSUMER_ERROR = BASE
```
ulzha: Inconsistent naming; in the above code the ones that are `rate` are named `_RATE`.

ulzha:
:+1: except for that one remark.

PR #323, opened 2017-11-06
Cleanup RunState fields when unnecessary

fabriziodemaria:


bergman:
👍

styx-common/src/main/java/com/spotify/styx/state/RunState.java
```diff
@@ -357,6 +361,8 @@ public RunState retryAfter(WorkflowInstance workflowInstance, long delayMillis)
               QUEUED,
               data().builder()
                   .retryDelayMillis(delayMillis)
+                  .executionId(empty())
```
danielnorberg: I was using these in my dependency checking branch (T-T)
I guess I can recreate them by replaying the event stream instead, but I'd like to not introduce more event replays if possible.

PR #324, opened 2017-11-08
Change LOG level for Pod events

fabriziodemaria:
With the current implementation of keeping the Pods around for some time after the Styx executions finish, we have an even stronger decoupling between the states in Styx and the Pods in the Kubernetes cluster. Receiving a Pod event that is not relevant for the current states within Styx is expected by design, making the related logging not belonging to the `warn` level.
Logging setup reference: https://github.com/spotify/styx/blob/master/CONTRIBUTING.md

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -518,19 +518,19 @@ class KubernetesDockerRunner implements DockerRunner {
 
     final RunState runState = stateManager.get(workflowInstance);
     if (runState == null) {
-      LOG.warn("Pod event for unknown or inactive workflow instance {}", workflowInstance);
+      LOG.debug("Pod event for unknown or inactive workflow instance {}", workflowInstance);
       return Optional.empty();
     }
 
     final Optional<String> executionIdOpt = runState.data().executionId();
     if (!executionIdOpt.isPresent()) {
-      LOG.warn("Pod event for state with no current executionId: {}", podName);
+      LOG.debug("Pod event for state with no current executionId: {}", podName);
       return Optional.empty();
     }
 
     final String executionId = executionIdOpt.get();
     if (!podName.equals(executionId)) {
-      LOG.warn("Pod event not matching current exec id, current:{} != pod:{}",
+      LOG.debug("Pod event not matching current exec id, current:{} != pod:{}",
           executionId, podName);
       return Optional.empty();
     }


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -518,19 +518,19 @@ private synchronized void tryPollPods() {
 
     final RunState runState = stateManager.get(workflowInstance);
     if (runState == null) {
-      LOG.warn("Pod event for unknown or inactive workflow instance {}", workflowInstance);
+      LOG.debug("Pod event for unknown or inactive workflow instance {}", workflowInstance);
```
bergman: This one doesn't happen nearly as often, does it? I think I'd keep this one and let the others be debug.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -518,19 +518,19 @@ private synchronized void tryPollPods() {
...
     if (!executionIdOpt.isPresent()) {
-      LOG.warn("Pod event for state with no current executionId: {}", podName);
+      LOG.debug("Pod event for state with no current executionId: {}", podName);
       return Optional.empty();
     }
 
     final String executionId = executionIdOpt.get();
     if (!podName.equals(executionId)) {
-      LOG.warn("Pod event not matching current exec id, current:{} != pod:{}",
+      LOG.debug("Pod event not matching current exec id, current:{} != pod:{}",
```
bergman: How often do we see this one? Timeout is 10 minutes and cleanup delay is 2 minutes. Should be a very rare thing.

fabriziodemaria: This could happen also if `retryAfter` is quick enough to start a new retry.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
fabriziodemaria: This happens quite frequently actually, at least every time a workflow instance becomes not active.

fabriziodemaria:
I don't think it matters how frequent the logs are, the point of this PR being an incorrect usage of the log levels specifications. Note that these logs are not removed, the PR only changes their criticality level.

bergman:
My point is just that if we put these into DEBUG, we will never see them again and would not notice problems with these growing.

bergman:
👍

PR #325, opened 2017-11-09
Add ?status query parameter for GET backfill

fabriziodemaria:
Also make use of the "status" query parameter in `backfillEditConcurrency` in the Styx client, to improve performance.

### DIFF
--- a/doc/api.apib
+++ b/doc/api.apib
@@ -526,13 +526,82 @@ Workflow Instance belonging to a Backfill entity.
             }
         }
 
-## Backfill [/{version}/backfills/{backfill_id}]
+## Backfill [/{version}/backfills/{backfill_id}{?status}]
 
 + Parameters
     + version: `v3` (enum[string]) - API version
         + Members
             + `v3`
     + backfill_id: `backfill-1489054446085-52684` (string) - Backfill ID
+
+### Get Backfill [GET]
+
++ Parameters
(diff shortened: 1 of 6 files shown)


fabriziodemaria:
Side comment about the PR: I added `shouldReturnNotFoundForMissingBackfill` test to make CodeCov pass the review.

doc/api.apib
```diff
@@ -526,14 +526,82 @@ Workflow Instance belonging to a Backfill entity.
...
 + Parameters
     + version: `v3` (enum[string]) - API version
         + Members
             + `v3`
     + backfill_id: `backfill-1489054446085-52684` (string) - Backfill ID
 
+### Get Backfill [GET]
+
++ Parameters
+    + status: `true` (optional, boolean) - If to return the status list per backfill
```
bergman: include default value

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
```diff
@@ -339,10 +339,11 @@
   }
 
   @Override
-  public CompletionStage<BackfillPayload> backfill(String backfillId) {
+  public CompletionStage<BackfillPayload> backfill(String backfillId, boolean status) {
```
bergman: name this `includeStatus` or similar so you can tell what will happen

bergman:
👍

PR #329, opened 2017-11-16
scheduler: configurable k8s namespace

danielnorberg:
*Note*: This makes the gke `namespace` a required configuration parameter. The scheduler will not start without it.

### DIFF
--- a/styx-standalone-service/src/main/resources/styx-standalone.conf
+++ b/styx-standalone-service/src/main/resources/styx-standalone.conf
@@ -18,6 +18,7 @@ styx.stale-state-ttls = {
 styx.gke.default.project-id = ""
 styx.gke.default.cluster-zone = ""
 styx.gke.default.cluster-id = ""
+styx.gke.default.namespace = ""
 
 # bigtable instance
 styx.bigtable.project-id = ""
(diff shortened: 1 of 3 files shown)


danielnorberg:
@honnix PTAL

styx-standalone-service/src/main/resources/styx-standalone.conf
```diff
@@ -18,6 +18,7 @@ styx.stale-state-ttls = {
 styx.gke.default.project-id = ""
 styx.gke.default.cluster-zone = ""
 styx.gke.default.cluster-id = ""
+styx.gke.default.namespace = "default"
```
honnix: Add to internal conf afterwards?

danielnorberg: sure, there is a default fallback in the code though

honnix:
Would be nice to have some test cases to cover the config trick.

styx-standalone-service/src/main/resources/styx-standalone.conf
honnix: Better to be more explicit. :)

honnix:
This part is largely untested and hard to do. It is where all the injections happen. So take the previous comment as bonus.

PR #331, opened 2017-11-20
Add WorkflowConsumer

fabriziodemaria:


styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -244,6 +253,7 @@ public static StyxScheduler createDefault() {
   private final RetryUtil retryUtil;
   private final WorkflowResourceDecorator resourceDecorator;
   private final EventConsumerFactory eventConsumerFactory;
+  private final WorkflowConsumerFactory workflowConsumerFactory;
```
honnix: Not a big fan of `consumer` naming here, since it doesn't really consume workflow.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -553,28 +567,34 @@ private void setupMetrics(
...
     return (workflow) -> {
       stats.registerActiveStatesMetric(
           workflow.id(),
           () -> stateManager.getActiveStatesCount(workflow.id()));
 
-      cache.store(workflow);
+      final Optional<Workflow> oldWorkflowOptional = cache.workflow(workflow.id());
+
       workflowInitializer.inspectChange(workflow);
+      cache.store(workflow);
```
honnix: This slightly changed the behaviour. if `inspectChange` fails, we will not update cache, which makes more sense.

honnix:
LGTM with minor issues.

ulzha:
:+1:

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
fabriziodemaria: The change was intentional ;)

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
fabriziodemaria: This was chosen also to be consistent with `EventConsumerFactory`. Any suggestion for a better name?

honnix: Not sure. This consumes the state change, or update, not the workflow itself. I really have no good suggestion. So it is ok.

PR #328, opened 2017-11-16
cli: refactor exit codes + add api & auth err code

danielnorberg:

PR #326, opened 2017-11-15
helpful credentials error message

danielnorberg:
Display a helpful error message telling the user how to set up credentials when getting a `401 Unauthorized` from Styx and there are no credentials available:

```
API error: Unauthorized: Please set up Application Default Credentials or set the GOOGLE_APPLICATION_CREDENTIALS env var to point to a file defining the credentials.

Hint: Try setting up credentials using gcloud: 

	$ gcloud auth application-default login
```

PR #332, opened 2017-11-20
OkHttp client in cli

bergman:
pros:
- no service, instance and all that

cons:
- apollo client not dog fooded

### DIFF
--- a/styx-cli/pom.xml
+++ b/styx-cli/pom.xml
@@ -23,19 +23,6 @@
     </dependency>
 
     <dependency>
-      <groupId>com.spotify</groupId>
-      <artifactId>apollo-core</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.spotify</groupId>
-      <artifactId>apollo-environment</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.spotify</groupId>
-      <artifactId>apollo-okhttp-client</artifactId>
-    </dependency>
-
-    <dependency>
       <groupId>org.fusesource.jansi</groupId>
       <artifactId>jansi</artifactId>
     </dependency>
--- a/styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
+++ b/styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
@@ -21,11 +21,10 @@
 package com.spotify.styx.cli;
 
 import static com.google.common.base.Throwables.getStackTraceAsString;
-import static com.spotify.apollo.Status.UNAUTHORIZED;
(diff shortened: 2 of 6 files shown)


honnix:
Wait for #328 and #326 ?

bergman:
sure, this can wait

danielnorberg:
Does this make the cli faster?

Tests would be great.

bergman:
I don't know about any improvements in speed but the starting-a-service-to-make-a-request dance is embarrassing.

danielnorberg:
Yeah, the Apollo scaffolding does seem a bit excessive and although we do get some configuration for free I'm not sure we need it.

However, would be good to not have both the `StyxOkHttpClient` and `StyxApolloClient` duplicate implementations of `StyxClient`. How about keeping `StyxApolloClient` and passing it a `OkHttpClient` in a apollo `Client` [wrapper](https://github.com/spotify/apollo/blob/master/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java)?

bergman:
Hmm, yeah, thanks for the pointer to the apollo wrapper thing, will check out!

bergman:
That apollo http client wrapper is private 😢

danielnorberg:
I thought it would be a trivial amount of code to replicate that wrapper in styx, but alas, it seems to not be so 😫  

I'd be happy with not using the apollo client at all. I.e., remove `StyxApolloClient` etc. It would be nice if we do that in a way that keeps git history. I.e., rename `StyxApolloClient` to `StyxOkHttpClient` and make surgical changes to replace the apollo client with the okhttp client in a single commit 🤓 

How about switching to grpc at some point? 😄

bergman:
The apollo client seems useful to any apollo services that want to integrate with styx but I guess it's easier to get a http client from an apollo client than the other way around… I'll take a stab at going apollo->http and if that's easy enough we could go http only for the client.

grpc would be amazing! Any ideas on how hard it would be for others to integrate our client if we go that route? Since we added auth, it's pretty hard to curl anyway… but any web/javascript integrations would still need http i guess.

danielnorberg:
What would be the impact to apollo services if the styx client doesn't use the apollo client?

Afaik the apollo client is typically not used when talking to external services using client libraries. E.g. google pubsub, gcs, dataflow, etc. If that's fine for those services, maybe it would be fine for styx as well?

danielnorberg:
I guess grpc would complicate things for talking to styx directly from a browser. Are we planning to do that? Maybe e.g. the [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) can be used to handle that when/if necessary.

bergman:
Closing as stale.

PR #333, opened 2017-11-20
Add stats for workflowConsumer

fabriziodemaria:


styx-scheduler-service/src/test/java/com/spotify/styx/monitoring/MetricsStatsTest.java
```diff
@@ -43,90 +65,8 @@
 
 @RunWith(MockitoJUnitRunner.class)
 public class MetricsStatsTest {
-
```
bergman: Nice removal! 😃

fabriziodemaria: Added back :)

PR #334, opened 2017-11-21
validate workflow before storing and update

honnix:
initializing next natural trigger could throw RuntimeException due to
wrong syntax of offset or schedule.

@ulzha PTAL

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/workflow/WorkflowInitializer.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/workflow/WorkflowInitializer.java
@@ -31,8 +31,11 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.Objects;
 import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class WorkflowInitializer {
+  private static final Logger LOG = LoggerFactory.getLogger(WorkflowInitializer.class);
 
   private final Storage storage;
   private final Time time;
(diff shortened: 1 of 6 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/workflow/WorkflowInitializer.java
```diff
@@ -46,32 +49,43 @@ public void inspectChange(Workflow workflow) {
...
+      } catch (Exception e) {
+        LOG.info("could not compute next natural trigger for workflow {}", workflow, e);
+        throw new WorkflowInitializationException(e);
+      }
+    }
+
+    try {
+      storage.storeWorkflow(workflow);
+      if (nextSpec.isPresent()) {
+        storage.updateNextNaturalTrigger(workflow.id(), nextSpec.get());
```
ulzha: I wonder why there's no "racy" comment here. If we storeWorkflow but fail to updateNextNaturalTrigger, then I guess the workflow stays in Styx as a zombie.

Anyway seems a minor concern, and not introduced in this PR.

ulzha:
:+1:

styx-scheduler-service/src/main/java/com/spotify/styx/workflow/WorkflowInitializer.java
honnix: It's not strictly a zombie. If the workflow was there, the worst case is next natural trigger not updated; if it was not there, we will have an empty next natural trigger. Either case, the client will receive 500.

PR #335, opened 2017-11-21
fix auth token in apollo client

bergman:


honnix:
👍  Let's merge and release

danielnorberg:
Good catch! =)

PR #336, opened 2017-11-21
Remove last execution id from 'styx ls'

fabriziodemaria:
#323 made the execution id disappear when replaying the events for workflow instances in QUEUED or PREPARED state, so in those cases the execution id do not show up with the CLI command `styx ls`. However, such information is not very relevant for the users and the execution id can still be retrieved via `styx events ...` so we can remove it completely from `styx ls`.

EDIT: we can also keep displaying the execution id for current executions


honnix:
👍

PR #337, opened 2017-11-23
styx runtime configuration object

danielnorberg:
Have a single object containing all runtime configuration parameters instead of a method per configuration parameter on the `Storage` interface.

Also removed the (effectively unused?) [`StyxConfigResource`](https://github.com/spotify/styx/blob/master/styx-api-service/src/main/java/com/spotify/styx/api/StyxConfigResource.java) and [`setGlobalEnabled`](https://github.com/spotify/styx/blob/master/styx-common/src/main/java/com/spotify/styx/storage/Storage.java#L93) method.

TODO:
* Centralize configuration polling, e.g. have an [`Observable<StyxConfig>`](http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html)
* Use `StyxConfig` throughout instead of specialized suppliers like [`Debug`](https://github.com/spotify/styx/blob/master/styx-common/src/main/java/com/spotify/styx/util/Debug.java) ?
* Add `{set,path}Config(StyxConfig)` methods to `Storage` ? (later when / if necessary ?)


bergman:
Looks good! Do you want to address the TODO's in this PR or another?

styx-api-service/src/main/java/com/spotify/styx/api/StyxConfigResource.java
```diff
@@ -1,104 +0,0 @@
...
-    final boolean enabled;
-    try {
-      enabled = storage.globalEnabled();
-    } catch (IOException e) {
-      throw Throwables.propagate(e);
-    }
-    return Response.forPayload(StyxConfig.create(enabled));
-  }
-
-  private Response<StyxConfig> patchStyxConfig(Request request) {
```
bergman: I think this is for having a big red halt button if things go crazy. Not sure if needed or if changing in datastore through cloud console is good enough.

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -111,10 +111,11 @@ private Middlewares() {
   }
 
   public static <T> Middleware<AsyncHandler<Response<T>>, AsyncHandler<Response<T>>> clientValidator(
-      Supplier<Optional<List<String>>> supplier) {
+      Supplier<List<String>> supplier) {
     return innerHandler -> requestContext -> {
       if (requestContext.request().header("User-Agent")
-          .map(header -> supplier.get().orElse(ImmutableList.of()).contains(header))
+          // TODO: should the blacklist be a set so this lookup is O(1) instead of O(n) ?
```
bergman: If it seems easy to do, sure.

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerServiceFixture.java
```diff
@@ -244,14 +247,6 @@ void givenWorkflowEnabledStateIs(Workflow workflow, boolean enabled) {
     }
   }
 
-  void givenTheGlobalEnableFlagIs(boolean enabled) {
```
bergman: Is there a test on system level that tests that things are not scheduled when this is disabled?

styx-api-service/src/main/java/com/spotify/styx/api/StyxConfigResource.java
danielnorberg: I did not know about this red button, it's not documented anywhere I can find (including our internal ops docs) and it seems scary to me to have this be a "public" api that anyone can use to take down styx without authorization.

Having this red button be documented and in the cloud console datastore UI (like so many of the other runtime configuration options atm) seems like an improvement & good enough to me for now.

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerServiceFixture.java
danielnorberg: No, there were no tests that verified that things are not scheduled if the global enable flag is disabled.

danielnorberg:
@bergman maybe the TODO's can be done in later PRs, otherwise this might never get merged =)

bergman:
👍

PR #338, opened 2017-11-30
k8s: watch from latest

danielnorberg:
In a high volume k8s cluster, resource versions go stale quickly enough that watching from a resource version returned by a pod list operation often fails with a "too old resource version" error.

Omitting the resource version parameter from the watch call causes k8s to watch from the latest resource version. This has been tested empirically using our GKE clusters but the docs are ambiguous on this point.

Styx might miss events when the watcher is re-established, but that is already the case.

In the high-volume case it does not seem likely that there is a way to not miss events using only the watcher, thus polling acts as a fallback.

For the low-volume case we could keep storing the last observed resource version etc in an attempt to not miss any events. However, I did not think that useful enough to warrant the complexity, given that the polling mechanism will likely observe terminated pods before they are GC'd in this case.

Note: Even omitting the resource version parameter from the watch call has been observed to immediately fail with a "too old resource version" error on occasion.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -438,10 +438,8 @@ class KubernetesDockerRunner implements DockerRunner {
         pollPodsIntervalSeconds,
         TimeUnit.SECONDS);
 
-    final String resourceVersion = client.pods().list().getMetadata().getResourceVersion();
     watch = client.pods()
-        .withResourceVersion(resourceVersion)
-        .watch(new PodWatcher(Integer.parseInt(resourceVersion)));
+        .watch(new PodWatcher());
   }
 
   private Set<WorkflowInstance> getRunningWorkflowInstances() {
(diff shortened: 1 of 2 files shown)


danielnorberg:
@ulzha @honnix PTAL

danielnorberg:
Related k8s issue: https://github.com/kubernetes/kubernetes/issues/7162

honnix:
I can’t think of any extremely bad corner cases. LGTM. @ulzha please review as well.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -555,7 +552,7 @@ private void emitPodEvents(Watcher.Action action, Pod pod, RunState runState) {
     }
   }
 
-  private void logEvent(Watcher.Action action, Pod pod, int resourceVersion,
+  private void logEvent(Watcher.Action action, Pod pod, String resourceVersion,
```
ulzha: :star:

ulzha:
Great to see TODOs and FIXMEs go.

Funny how event stream processing in reliable fashion is such a hard problem across the world.

PR #339, opened 2017-12-01
attach details to reason phrase

honnix:
Append whatever jackson spits out to reason phrase. Not entirely sure
this is very readable, but currently don't have any better idea.

@NatashaL PTAL


NatashaL:
👍

PR #340, opened 2017-12-07
upgrade cron-utils

honnix:
Basically to get this fix in: https://github.com/jmrozanec/cron-utils/issues/144

Due to a change introduced by
https://github.com/jmrozanec/cron-utils/issues/155, we throw exception
if get an empty optional. This doesn't change the orignal behavior at
all, and we don't exepct this thing to happen in our use cases.

This change https://github.com/jmrozanec/cron-utils/issues/200 breaks
our previous assumption, and that's why there are so many fixes in this PR.

@danielnorberg PTAL

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,7 +173,7 @@
       <dependency>
         <groupId>com.cronutils</groupId>
         <artifactId>cron-utils</artifactId>
-        <version>5.0.5</version>
+        <version>7.0.0</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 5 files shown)


honnix:
This change https://github.com/jmrozanec/cron-utils/issues/200 breaks our previous assumption.

honnix:
How could it be `100% of diff hit`? It doesn't seem to check lambda...

danielnorberg:
Good stuff! 👍

PR #341, opened 2017-12-11
Command "styx retry" do not bypass resources check

fabriziodemaria:
That's done by sending `retryAfter` event that sets the delay to 0, instead of the `dequeue` event. A check is also added for the `/events` API endpoint to translate `dequeue` events into `retryAfter` events.

This PR also includes a refactoring of `RunState` that allows to have a `transition` method and a `backwardsCompatibleTransition` method, so that backwards compatible transitions can be allowed only where they actually make sense (mostly events' replying).
The reason the afore-mentioned refactoring is introduced as part of this PR is to be able to have `/events` API endpoint to refuse events that would cause old transitions; more specifically, sending a `retry` or a `submit` events to a Workflow Instance that is in QUEUED state is now not allowed.


honnix:
Is this ready for review? @danielnorberg PTAL anyway since it doesn't seem like a trivial change.

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -161,9 +162,15 @@ public SchedulerResource(
     }
 
     try {
-      stateManager.receive(event);
+      if ("dequeue".equals(EventUtil.name(event))) {
```
danielnorberg: How about only allowing `dequeue` here for backwards compatibility and rejecting all other event types?

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
```diff
@@ -255,7 +255,7 @@
         WorkflowId.create(componentId, workflowId),
         parameter);
     try {
-      final ByteString payload = serialize(Event.dequeue(workflowInstance));
+      final ByteString payload = serialize(Event.retryAfter(workflowInstance, 0L));
```
danielnorberg: How about introducing a more specific API for instructing styx to `retry` or `halt` a workflow instance instead of injecting raw events?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -160,13 +160,12 @@ void tick() {
     timedOutInstances.forEach(this::sendTimeout);
 
     for (InstanceState eligibleInstance : eligibleInstances) {
-      final boolean proceed = limitAndDequeue(
-          resources, workflowResourceReferences, currentResourceUsage, eligibleInstance);
-      if (!proceed) {
+      if (!dequeueRateLimiter.tryAcquire()) {
```
danielnorberg: Why is the dequeue rate limiting moved here? It seems undesirable to count e.g. dequeues that didn't happen because they were resource limited towards the rate limit.

styx-common/src/main/java/com/spotify/styx/state/RunState.java
```diff
@@ -373,41 +352,182 @@ public RunState retryAfter(WorkflowInstance workflowInstance, long delayMillis)
...
+      return state(FAILED);
+    }
+
+    @Override
+    public RunState halt(WorkflowInstance workflowInstance) {
+      return state(ERROR);
+    }
+  }
+
+  private class BackwardsCompatibleTransitionVisitor implements EventVisitor<RunState> {
```
danielnorberg: How do we ensure that `BackwardsCompatibleTransitionVisitor` is kept in sync with `TransitionVisitor`? I.e., when we add a transition to `TransitionVisitor`, we need to ensure that the same transition is added to `BackwardsCompatibleTransitionVisitor`. A sequence of events successfully replayed using `TransitionVisitor` should produce the same sequence of states when replayed using `BackwardsCompatibleTransitionVisitor`.

I would prefer to avoid the problem entirely by eliminating the need to perform replay of historical workflow instances and events with deprecated transitions.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
fabriziodemaria: Good point, I misinterpreted how the RateLimiter works. Reverting the commit.

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
fabriziodemaria: Done

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
fabriziodemaria: Added

styx-common/src/main/java/com/spotify/styx/state/RunState.java
fabriziodemaria: > I would prefer to avoid the problem entirely by eliminating the need to perform replay of historical workflow instances and events with deprecated transitions.

I agree on this, but I think this objective falls outside the scope of this PR.

What I like about having two visitors is that we can isolate the backwards compatible operations only for the replaying cases, while the "online" operations are guaranteed to operate properly (this also forces us to maintain test cases up-to-date for online operations). The downside is indeed that when we change the state machine logic we need to update both visitors. 
I see both pros and cons, how do you think we should proceed @danielnorberg?

doc/api.apib
```diff
@@ -739,26 +739,45 @@ Resources related to inspect detailed status of Workflow Instances.
...
-## Events [/{version}/scheduler/events]
+## Retry [/{version}/scheduler/retry]
 
 + Parameters
     + version: `v3` (enum[string]) - API version
         + Members
             + `v3`
 
-### Inject Event [POST]
+### Send Retry [POST]
```
danielnorberg: Document the `delay` parameter as well?

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -154,19 +168,48 @@ public SchedulerResource(
...
-      return Response.forStatus(BAD_REQUEST.withReasonPhrase("Workflow instance not found"));
+    if ("dequeue".equals(EventUtil.name(event))) {
+      // For backwards compatibility
+      return Response.forStatus(eventInjectorHelper(
+          Event.retryAfter(event.workflowInstance(), 0L))).withPayload(event);
+    } else if ("halt".equals(EventUtil.name(event))) {
+      // For backwards compatibility
+      return Response.forStatus(eventInjectorHelper(event));
+    } else {
+      return Response.forStatus(NOT_ACCEPTABLE.withReasonPhrase(
```
danielnorberg: `406 Not Acceptable` seems to have to do with content type negotiation and accept headers, which I guess is not applicable here.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.7

Use `400 Bad Request` instead ?

danielnorberg:
👍  LGTM apart from a few small things

honnix:
How about we put those new endpoints to v4?

honnix:
I’m a bit surprised we pushed in API changes in this PR. :)

danielnorberg:
@honnix afaik there's effectively no breaking changes; the api is still backwards compatible wrt the `retry` and `halt` events that existing clients send.

Adding these endpoints is how I imagined we would tackle this issue, I like it. What are the changes you want to see in the client and cli? =)

honnix:
Wouldn't we call the new endpoints `/retry` and `/halt` from the client?

honnix:
Sorry I missed the client change in this PR. :(

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
fabriziodemaria: Good point, changed it to 400

doc/api.apib
fabriziodemaria: Fixed

honnix:
👍

PR #342, opened 2017-12-13
use mock GoogleIdTokenAuth

honnix:
Let's not do real auth for unit test.

Apart from this, our code doesn't handle GCE default SA properly.
https://github.com/spotify/styx/blob/4c21ca1d6ae5894e15103e339c36b4a8f6b40c30/styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java#L61
For GCE default SA, this will return null, and we treat it as user
account!!

Those test cases failed in GCP Jenkins.

@danielnorberg PTAL


honnix:
This however makes Codecov sad. :\

danielnorberg:
👍

PR #343, opened 2017-12-14
Improve error handling for events injection APIs

fabriziodemaria:
The `SyncStateManager` used in some tests might not behave exactly as the `QueuedStateManager` used online. This makes parts of the tests relying on `SyncStateManager` potentially wrong or irrelevant. 
In this specific case, different error handling in `SyncStateManager` with respect to `QueuedStateManager` caused unexpected API error codes being returned from the event injecting APIs.
In future PRs we could remove `SyncStateManager` from these tests and think about a different approach.


styx-scheduler-service/src/test/java/com/spotify/styx/api/SchedulerResourceTest.java
```diff
@@ -103,6 +103,9 @@
 
   private final DockerImageValidator dockerImageValidator = mock(DockerImageValidator.class);
 
+  public SchedulerResourceTest() throws InterruptedException {
```
honnix: why?

fabriziodemaria: Some IntelliJ suggestion that got forgotten in the process. Removed :)

honnix:
+1 Thanks for fixing this!

PR #344, opened 2017-12-15
clean up workflow deletion prompt

danielnorberg:
Before: 
`Sure you want to delete the workflow Fooin component foo? [yN]`

After: 
`Sure you want to delete the workflow Foo in component foo? [y/N] `


bergman:
👍

PR #347, opened 2017-12-19
Resource usage count fix

NatashaL:
We suspect there is a race condition between when `currentResourceUsage` is calculated (by assuming workflows in `State.NEW` consume resources) and the thread of `TriggerManager` creating workflows with `State.NEW`. `QueuedStateManager` then takes a little while till it transitions those to `State.QUEUED`. This we think causes sometimes inconsistent metrics on resource usage.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -170,7 +170,14 @@ public class Scheduler {
     updateStats(resources, currentResourceUsage);
   }
 
-  private boolean isConsumingResources(State state) {
+  /**
+   * We'll keep counting terminal states as if they consume resources. They are transient states and
+   * should go away fairly quickly. If they don't, then we might be having some trouble cleaning up
+   * the containers. In that case it's better to be conservative on resource usage.
+   *
+   * @return true if the state consumes resources, otherwise false.
+   */
+  private static boolean isConsumingResources(State state) {
     return state != State.NEW && state != State.QUEUED;
   }
 
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -143,7 +143,7 @@ public class Scheduler {
     final Map<String, Long> currentResourceUsage =
         activeStates.parallelStream()
             .filter(entry -> !timedOutInstances.contains(entry.workflowInstance()))
-            .filter(entry -> entry.runState().state() != State.QUEUED)
+            .filter(entry -> isConsumingResources(entry.runState().state()))
             .flatMap(instanceState -> pairWithResources(globalConcurrency, instanceState))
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -170,6 +170,10 @@ void tick() {
     updateStats(resources, currentResourceUsage);
   }
 
+  private boolean isConsumingResources(State state) {
+    return state != State.NEW && state != State.QUEUED;
```
danielnorberg: Should we also exclude `TERMINATED`, `FAILED`, `ERROR`, and `DONE` ?

danielnorberg:
Looks good overall to me apart a comment on the states included in the resource usage count.

This also made me realize that the Styx resource counts cannot currently account for resources (e.g. dataflow jobs etc) left behind by terminated workflows, which seems to me like another reason to make Styx actively terminate such jobs. But that's for another PR.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
NatashaL: We could, although as @honnix mentioned once, those are transient states and should quickly get removed. If they don't get cleaned up then that's a reason for concern, and it's good to be able to see it (although maybe with another metric). 
Also the terminal states don't cause spikes in currentResourceUsage the way `State.NEW` does.

honnix: Yeah, let's keep it simple. Those transient states will go away fairly quickly, and if they don't we might be having some trouble and in that case be conservative on resource usage might be better.

styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
```diff
@@ -365,22 +366,29 @@ public void shouldLimitConcurrencyForResource() throws Exception {
 
     assertThat(countInState(State.QUEUED), is(1));
     assertThat(countInState(State.PREPARE), is(3));
+    verify(stats).recordResourceUsed("r1", 3L);
   }
 
   @Test
-  public void shouldCountResourcesOnNonQueuedStates() throws Exception {
+  public void shouldCountResourcesOnStatesNotConsumingResources() throws Exception {
```
bergman: i think the name of this test is wrong, there's a `not` in there

NatashaL: good catch :D

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: SGTM, how about a comment saying that?

honnix: Will fix.

honnix:
👍

PR #348, opened 2017-12-20
Support for regional GKE clusters

bergman:


fabriziodemaria:
This was tested in a staging environment since no integration tests are currently available for fabric8 and GKE. +1

PR #350, opened 2017-12-22
remove redundant @JsonProperty annotations

danielnorberg:
Our ObjectMapper is already configured to use snake case.


bergman:
LGTM! I guess these could be useful for people using their own object mapper but what's to say they would even be using jackson and then we would be in the same place anyway…

honnix:
another 👍

PR #352, opened 2018-01-03
add CLI support to enable and disable workflows

honnix:
the client side is more generic with the ability to handle a complete
workflow state including `enabled`, `nextNaturalTrigger`, and
`nextNaturalOffsetTrigger`.

@danielnorberg PTAL

_Github has signed me out due to long time idling so I decided to write some code._

### DIFF
--- a/styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
+++ b/styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
@@ -381,16 +381,28 @@ public final class CliMain {
     final String component = namespace.getString(parser.workflowEnableComponentId.getDest());
     final List<String> workflows = namespace.getList(parser.workflowEnableWorkflowId.getDest());
     workflowStateUpdate(component, workflows, WorkflowState.patchEnabled(true))
-        .forEach(t -> cliOutput
-            .printMessage("Workflow " + t._1 + " in component " + component + " enabled."));
+        .forEach(t -> {
+          if (t._2.enabled().orElse(false)) {
+            cliOutput.printMessage("Workflow " + t._1 + " in component " + component + " enabled.");
+          } else {
+            cliOutput.printMessage(
+                "Failed to enable workflow " + t._1 + " in component " + component + ".");
+          }
+        });
   }
 
   private void workflowDisable() throws ExecutionException, InterruptedException {
     final String component = namespace.getString(parser.workflowDisableComponentId.getDest());
     final List<String> workflows = namespace.getList(parser.workflowDisableWorkflowId.getDest());
     workflowStateUpdate(component, workflows, WorkflowState.patchEnabled(false))
-        .forEach(t -> cliOutput
-            .printMessage("Workflow " + t._1 + " in component " + component + " disabled."));
+        .forEach(t -> {
+          if (!t._2.enabled().orElse(false)) {
+            cliOutput.printMessage("Workflow " + t._1 + " in component " + component + " disabled.");
+          } else {
+            cliOutput.printMessage(
+                "Failed to disable workflow " + t._1 + " in component " + component + ".");
+          }
+        });
   }
 
   private List<Tuple2<String, WorkflowState>> workflowStateUpdate(String component,
--- a/styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
+++ b/styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
@@ -30,6 +30,7 @@ import static net.sourceforge.argparse4j.impl.Arguments.fileType;
 import com.fasterxml.jackson.databind.MappingIterator;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import com.spotify.apollo.Client;
 import com.spotify.apollo.core.Service;
 import com.spotify.apollo.core.Services;
@@ -258,6 +259,12 @@ public final class CliMain {
             case DELETE:
               workflowDelete();
               break;
+            case ENABLE:
+              workflowEnable();
+              break;
+            case DISABLE:
+              workflowDisable();
+              break;
             default:
               // parsing unknown command will fail so this would only catch non-exhaustive switches
               throw new ArgumentParserException(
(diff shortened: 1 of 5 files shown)


honnix:
circle ci is lightening fast!!

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -380,6 +377,74 @@ private void workflowCreate() throws IOException, ExecutionException, Interrupte
...
+      throws ExecutionException, InterruptedException {
+    final List<Tuple2<String, CompletionStage<WorkflowState>>> futures = workflows.stream()
+        .map(workflow -> Tuple.of(workflow,
+                                  styxClient.updateWorkflowState(component,
+                                                                 workflow,
+                                                                 workflowState)))
+        .collect(toList());
+
+    final List<Tuple2<String, WorkflowState>> updatedWorkflowStates =
+        Lists.newArrayListWithCapacity(futures.size());
```
bergman: why not just `new ArrayList<>(futures.size())`?

honnix: yeah actually should. this is deprecated way.

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -771,7 +848,9 @@ public Subparser parser(Subparsers subCommands) {
   private enum WorkflowCommand {
     SHOW("get", "Show info about a specific workflow"),
     CREATE("", "Create or update a workflow"),
-    DELETE("", "Delete a workflow");
+    DELETE("", "Delete a workflow"),
```
bergman: since these take multiple workflows, maybe it should say `Delete workflow(s)` and so on

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -380,6 +377,74 @@ private void workflowCreate() throws IOException, ExecutionException, Interrupte
...
+                                                                  WorkflowState workflowState)
+      throws ExecutionException, InterruptedException {
+    final List<Tuple2<String, CompletionStage<WorkflowState>>> futures = workflows.stream()
+        .map(workflow -> Tuple.of(workflow,
+                                  styxClient.updateWorkflowState(component,
+                                                                 workflow,
+                                                                 workflowState)))
+        .collect(toList());
+
+    final List<Tuple2<String, WorkflowState>> updatedWorkflowStates =
```
bergman: I guess this whole thing could be made into some kind of `map/flatMap` to wait for and process these futures but the checked exceptions make it quite ugly.

styx-cli/src/test/java/com/spotify/styx/cli/CliMainTest.java
```diff
@@ -205,7 +207,109 @@ public void testWorkflowDeleteInteractiveForce() throws Exception {
...
+    CliMain.run(cliContext, "workflow", "disable", component, "foo", "bar");
+
+    verify(client).updateWorkflowState(component, "foo", workflowState);
+    verify(client).updateWorkflowState(component, "bar", workflowState);
+    verify(cliOutput).printMessage("Workflow foo in component " + component + " disabled.");
+    verify(cliOutput).printMessage("Workflow bar in component " + component + " disabled.");
+  }
+
+  @Test
+  public void shouldFailToDisableWorkflow() {
```
bergman: I had to read this test a few times to understand what it was doing, perhaps name it `shouldFailWhenReturningUnexpectedState`

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -380,6 +377,74 @@ private void workflowCreate() throws IOException, ExecutionException, Interrupte
...
+          }
+        });
+  }
+
+  private void workflowDisable() throws ExecutionException, InterruptedException {
+    final String component = namespace.getString(parser.workflowDisableComponentId.getDest());
+    final List<String> workflows = namespace.getList(parser.workflowDisableWorkflowId.getDest());
+    workflowStateUpdate(component, workflows, WorkflowState.patchEnabled(false))
+        .forEach(t -> {
+          if (!t._2.enabled().orElse(false)) {
```
bergman: this seems clunky… shouldn't we be able to trust the http status code to say whether the operation succeeded or not?

honnix: Maybe we can. Our server side code doesn't really check whether the change is successful, and in case of concurrent modification http code can be lying. Maybe too paranoid.

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
honnix: Yeah, that's the main reason.

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
honnix: Good points and will fix.

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
fabriziodemaria: I tend to prefer relying on the http status code as the standard unique way to assess correctness of an API call and move all the verification steps on the server side (before determining the status code to return).
I'd be happy with just removing this check from the CLI for now and address the verification step for `WorkflowState` cahnges on the server side in a future PR, if really needed.

fabriziodemaria:
As a generic comment, I find it quite annoying that we are restricted to batch this enabled changes per component and that we still need to manually indicate the component id in the first place. Hopefully components will be become just a way to _optionally_ group workflows when needed.

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
honnix: Cool let’s do that then. Two votes now. 😀

honnix:
rebased from master.

fabriziodemaria:
LGTM

PR #354, opened 2018-01-11
Backfill description

honnix:
Add one optional field when creating a backfill to describe the purpose of it.


styx-common/src/main/java/com/spotify/styx/model/Backfill.java
```diff
@@ -44,6 +45,9 @@
   @JsonProperty
   int concurrency();
 
+  @JsonProperty
```
danielnorberg: redundant `@JsonProperty`

danielnorberg:
👍

PR #358, opened 2018-01-15
Improved CLI Readability

Zatine:
Made column sizes more appropriate when listing workflows using CLI, and also truncating backfill descriptions by default. To not truncate you supply the command "--no-trunc" similar to Docker, and then you will get the full descriptions.

### DIFF
--- a/styx-cli/src/test/java/com/spotify/styx/cli/PrettyCliOutputTest.java
+++ b/styx-cli/src/test/java/com/spotify/styx/cli/PrettyCliOutputTest.java
@@ -37,16 +37,24 @@ import org.junit.Test;
 
 public class PrettyCliOutputTest {
 
-  private static final Backfill BACKFILL = Backfill.newBuilder()
-      .id("backfill-2")
-      .start(Instant.parse("2017-01-01T00:00:00Z"))
-      .end(Instant.parse("2017-01-02T00:00:00Z"))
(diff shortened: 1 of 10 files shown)


styx-cli/src/test/java/com/spotify/styx/cli/PrettyCliOutputTest.java
```diff
@@ -37,16 +37,25 @@
...
-      .id("backfill-2")
-      .start(Instant.parse("2017-01-01T00:00:00Z"))
-      .end(Instant.parse("2017-01-02T00:00:00Z"))
-      .workflowId(WorkflowId.create("component", "workflow2"))
-      .concurrency(2)
-      .description("Description")
-      .nextTrigger(Instant.parse("2017-01-01T00:00:00Z"))
-      .schedule(Schedule.DAYS)
-      .build();
+  private Backfill backfill(String description) {
```
honnix: Let’s move this method down after those properties and make it static.

danielnorberg:
@honnix I think codecov is correctly indicating that changes are not covered by tests...

honnix:
Hmm, I think we misunderstood it. We checked those red blocks in overview page and that’s why it is weird. Should instead check the diff tab. Will fix.

honnix:
Let's add test cases to cover all the changes.

honnix:
👍 awesome!

PR #359, opened 2018-01-16
[hold] scheduler: retry later on resource exhaustion

danielnorberg:
Avoiding spinning on the same resource exhausted instances every tick.

*Note*: This will cause a `retryAfter` event to be written every minute when an instance is resource limited. This might cause a rather large number of events to accumulate in bigtable and make replay of the instance event log to become expensive and fail.

Ideally there would be some mechanism to postpone processing of a WFI that does not involve writing a `retryAfter` event.

Mitigating by doing exponentially longer `retryAfter` intervals would result in less events but might result in unacceptably long execution delays.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -49,6 +49,7 @@ import com.spotify.styx.state.TimeoutConfig;
 import com.spotify.styx.storage.Storage;
 import com.spotify.styx.util.Time;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


honnix:
Let's fix codacy first.

As you have pointed out, this change has quite a bit side effects. So not entirely sure...

danielnorberg:
Yeah, I think I'd like to figure out a way to do this without writing a new `retryAfter` event each time.

Perhaps it'd be good enough to keep some throttling state in the scheduler instead. 

I think this becomes more urgent as we roll out the scheduler HA work and will start re-reading the workflow and it's sharded resource usage counters on each tick.

PR #360, opened 2018-01-16
[hold] scheduler: cache resources and config

danielnorberg:
Avoid re-reading the resource and config entities on each tick.


honnix:
LGTM apart from a small concern: would this impact user experience? E.g. I change concurrency of a resource, but it will take max half a minute to actually work. This would be hardly noticeable I guess.

danielnorberg:
I think that delay is ok. Anyway if a user lowers the limit e.g. because too many workflows are running, the already running workflows will not be stopped.

honnix:
+1

honnix:
I remember @danielnorberg preferred to hold this for now.

PR #361, opened 2018-01-16
use semantic-metrics 0.11.2

danielnorberg:
This version ffwd exposes the meter counter which can be useful to do derivation in the metrics platform instead of relying on the in-process aggregation.


honnix:
👍 

This artifact is introduced by `com.spotify:apollo-metrics:jar:1.3.2`, will this version be compatible? Looks OK from the repo's commit history.

https://github.com/spotify/styx/pull/265 will not be needed after this is merged.

danielnorberg:
Should be compatible. It will also be in the next apollo 1.x release, but not sure when/if that will happen. Apollo hasn't had a release in months.

PR #362, opened 2018-01-16
Store WFI state in datastore

danielnorberg:
Start migrating WFI state to datastore by writing it to datastore as part of the state machine transitions. Also read state from datastore if present, falling back to event replay if not.


danielnorberg:
@NatashaL @ulzha @fabriziodemaria PTAL

danielnorberg:
rebased again

styx-common/src/main/java/com/spotify/styx/serialization/PersistentWorkflowInstanceState.java
```diff
@@ -0,0 +1,65 @@
...
+import io.norberg.automatter.AutoMatter;
+import java.time.Instant;
+import javax.annotation.Nullable;
+
+@AutoMatter
+public interface PersistentWorkflowInstanceState {
+
+  long counter();
+
+  // TODO: remove @Nullable when all persisted active states in datastore contains the runstate
```
ulzha: TODO remove counter() as well, I guess?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -112,6 +121,19 @@
   public static final String PROPERTY_CONFIG_DEBUG_ENABLED = "debug";
   public static final String PROPERTY_SUBMISSION_RATE_LIMIT = "submissionRateLimit";
 
+  public static final String PROPERTY_STATE = "state";
+  public static final String PROPERTY_STATE_TIMESTAMP = "stateTimestamp";
```
ulzha: Seems a bit unexpected to have this list of constants mirroring StateData members but living outside of StateData.java.

Checked for analogues of Jackson annotations for Datastore, and found [@PersistenceCapable](https://cloud.google.com/appengine/docs/standard/java/datastore/jdo/dataclasses#Class_and_Field_Annotations)  - something worth exploring or already explored?

styx-common/src/main/java/com/spotify/styx/serialization/PersistentWorkflowInstanceState.java
danielnorberg: Not sure about that. Might still need it to be able to index events we write to bigtable or datastore.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: I agree that it's a bit annoying code but I don't think it's a big deal. And I think the decoupling of datastore entity representation from the RunState/StateData layout (StateData in particular is exposed through the client api and is used in the cli) is a good thing.

Using JDO doesn't seem recommended, it's listed under `Superseded Storage Solutions` in the app engine docs. The cloud datastore docs do not seem to mention JDO at all. 

[Objectify](https://github.com/objectify/objectify) seems like a recommended alternative, but is apparently [app engine only](https://github.com/objectify/objectify/wiki/FrequentlyAskedQuestions#can-i-use-objectify-from-a-gae-flexible-runtime). I guess the same might be true of JDO.

styx-common/src/test/java/com/spotify/styx/util/ReplayEventsTest.java
```diff
@@ -152,4 +158,34 @@ public void restoreRunStateForActiveInstance(long counter, State expectedState,
     assertThat(restoredRunState.data().trigger().get(), is(Trigger.natural()));
   }
 
+  @Test
+  @Parameters({"true", "false",})
+  public void restoreRunStateForActiveInstanceFromPersistentSate(boolean printLogs) throws Exception {
```
ulzha: Typo

styx-scheduler-service/src/test/java/com/spotify/styx/SystemTest.java
```diff
@@ -609,23 +631,39 @@ public void restoredStatesUseOriginalTimestamps() throws Exception {
   }
 
   @Test
-  public void restoresActiveStatesFromBigtable() throws Exception {
+  @Parameters({
+      "true, true",
+      "true, false",
+      "false, true",})
+  public void restoresActiveStatesFromBigtable(boolean withEvents, boolean withPersistentState) throws Exception {
```
ulzha: A bit misleading name now that in "false, true" scenario we ignore Bigtable.

styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
```diff
@@ -56,7 +57,19 @@ private ReplayEvents() {
...
     return instances.entrySet().parallelStream().map(entry -> {
       final WorkflowInstance workflowInstance = entry.getKey();
-      final long lastConsumedEvent = entry.getValue();
+      final PersistentWorkflowInstanceState persistentState = entry.getValue();
+      final long lastConsumedEvent = persistentState.counter();
+
+      // TODO: always use the persistent state from datastore when it has been migrated
+      if (persistentState.state() != null) {
+        if (printLogs) {
+          LOG.debug("Using persistent state instead of replaying events");
```
fabriziodemaria: This logging is not very useful in this form, let's add the processed WorkflowInstance as well?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -354,31 +376,85 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, Instant instant) thr
...
+      final Entity.Builder entity = Entity.newBuilder(key)
           .set(PROPERTY_COMPONENT, workflowInstance.workflowId().componentId())
           .set(PROPERTY_WORKFLOW, workflowInstance.workflowId().id())
           .set(PROPERTY_PARAMETER, workflowInstance.parameter())
-          .set(PROPERTY_COUNTER, counter)
-          .build();
+          .set(PROPERTY_COUNTER, state.counter());
+
+      // TODO: always write these fields when event-only replay tests have been removed
+      if (state.state() != null) {
```
ulzha: Where in WFI lifecycle does this become nonnull? I kind of expected the state to be unconditionally _written_ to Datastore here - then it would be clear to me that in 24 hours all the active states are dark-initialized in Datastore alright, but now I'm a bit confused.

ulzha:
Overall LGTM - the changes seem nicely contained and regular thus far.

styx-common/src/main/java/com/spotify/styx/storage/InMemStorage.java
```diff
@@ -325,6 +328,7 @@ public void deleteActiveState(WorkflowInstance workflowInstance) {
 
   public Optional<Long> getCounterFromActiveStates(WorkflowInstance workflowInstance)
       throws IOException {
-    return Optional.ofNullable(activeStatesMap.get(workflowInstance));
+    return Optional.ofNullable(activeStatesMap.get(workflowInstance))
+        .map(PersistentWorkflowInstanceState::counter);
```
fabriziodemaria: I think `getLatestStoredCounter()` from AggregateStorage (and InMemStorage) should read from ActiveStates instead of from the events, especially now that we are migrating away from events as source of truth.

NatashaL:
+1 to merge, but before adding this to a release, I'd like to look at making https://github.com/spotify/styx/blob/master/styx-scheduler-service/src/main/java/com/spotify/styx/monitoring/MonitoringHandler.java stateless. It feels like the timestamps needed there, for the metrics that are emitted there, belong in `RunState` / `PersistentWorkflowInstanceState`. If we can add/move these fields there before starting to write to Datastore, we can avoid a migration/ or adding backwards compatible quirks so early on.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: During normal operations the state should always be non-null. 

Like the TODO mentions, some tests write `PersistentWorkflowInstanceState` instances with null `state` though in order to test scenarios where we fall back to event replay from bigtable due to state not being in datastore.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -701,7 +777,20 @@ private Entity backfillToEntity(Backfill backfill) {
         : Optional.empty();
   }
 
+  private <T> Optional<T> readOptJson(Entity entity, String property, Class<T> cls) throws IOException {
```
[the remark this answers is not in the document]
danielnorberg: Refactoring `parseWorkflowJson` is out of scope for this PR.

styx-common/src/main/java/com/spotify/styx/storage/InMemStorage.java
danielnorberg: `getLatestStoredCounter` is used when re-initializing a WFI that has been deactivated and removed from datastore. Thus we need to count the events in bigtable.

I'd be all for keeping WFI information/state in datastore even after deactivation, but not sure that should go into this PR.

styx-common/src/main/java/com/spotify/styx/storage/Storage.java
```diff
@@ -145,26 +146,25 @@
...
    * <p>A {@link WorkflowInstance} is active if there has been at least one call to
-   *
-   * {@link #writeActiveState(WorkflowInstance, long)} and no calls to
+   * {@link #writeActiveState(WorkflowInstance, PersistentWorkflowInstanceState)} and no calls to
    * {@link #deleteActiveState(WorkflowInstance)}.
    *
    * @return The map of workflow instances to sequence counts
    */
-  Map<WorkflowInstance, Long> readActiveWorkflowInstances() throws IOException;
+  Map<WorkflowInstance, PersistentWorkflowInstanceState> readActiveWorkflowInstances() throws IOException;
```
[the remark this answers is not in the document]
danielnorberg: Yeah, I decided against refactoring `RunState` in order to keep this PR manageable and not get stuck in refactoring the world. Incremental steps.

danielnorberg:
@NatashaL wdyt about https://github.com/spotify/styx/pull/371 for making `MonitoringHandler` stateless?

danielnorberg:
Might be good to also merge either https://github.com/spotify/styx/pull/372 or https://github.com/spotify/styx/pull/367 to avoid running into the entity size limit for WFIs with a lot of messages.

styx-common/src/main/java/com/spotify/styx/storage/InMemStorage.java
fabriziodemaria: Let's not include this work in this PR, but I agree we should eventually remove such operational requirement from BigTable.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
ulzha: Oh so it's only null in tests...

ulzha:
:+1:

PR #363, opened 2018-01-17
Add TransactionalStorage

fabriziodemaria:
This provides a generic interface to execute various Storage operations belonging to the same transaction and manually commit the transaction, in order to avoid having too much scheduler logic in the Storage class.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -753,7 +753,8 @@ class DatastoreStorage {
       return value;
     } catch (DatastoreException ex) {
       tx.rollback();
-      throw new TransactionException(false, ex);
+      final boolean conflict = ex.getCode() == 10;
+      throw new TransactionException(conflict, ex);
     } finally {
       if (tx.isActive()) {
         tx.rollback();
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -40,6 +40,7 @@ import com.google.cloud.datastore.StringValue;
 import com.google.cloud.datastore.StructuredQuery.CompositeFilter;
 import com.google.cloud.datastore.StructuredQuery.Filter;
 import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
+import com.google.cloud.datastore.Transaction;
 import com.google.cloud.datastore.Value;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
@@ -78,6 +79,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Stream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -146,16 +148,21 @@ class DatastoreStorage {
 
   private final Datastore datastore;
   private final Duration retryBaseDelay;
-  private final KeyFactory componentKeyFactory;
+  private final Function<Transaction, DatastoreTransactionalStorage> transactionalStorageFactory;
 
   @VisibleForTesting
   final Key globalConfigKey;
 
   DatastoreStorage(Datastore datastore, Duration retryBaseDelay) {
+    this(datastore, retryBaseDelay, DatastoreTransactionalStorage::new);
+  }
+
+  DatastoreStorage(Datastore datastore, Duration retryBaseDelay,
+      Function<Transaction, DatastoreTransactionalStorage> transactionalStorageFactory) {
     this.datastore = Objects.requireNonNull(datastore);
     this.retryBaseDelay = Objects.requireNonNull(retryBaseDelay);
+    this.transactionalStorageFactory = Objects.requireNonNull(transactionalStorageFactory);
 
-    this.componentKeyFactory = datastore.newKeyFactory().setKind(KIND_COMPONENT);
     this.globalConfigKey = datastore.newKeyFactory().setKind(KIND_STYX_CONFIG).newKey(KEY_GLOBAL_CONFIG);
   }
 
@@ -181,7 +188,7 @@ class DatastoreStorage {
   }
 
   boolean enabled(WorkflowId workflowId) throws IOException {
-    final Key workflowKey = workflowKey(workflowId);
+    final Key workflowKey = workflowKey(datastore.newKeyFactory(), workflowId);
 
     return getOpt(datastore, workflowKey)
         .filter(w -> w.contains(PROPERTY_WORKFLOW_ENABLED))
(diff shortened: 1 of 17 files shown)


danielnorberg:
Can we have a method like `runInTransaction()` method that takes care of commit/rollback instead of doing manual commit/rollback in the scheduler code etc?

I.e. something like this: 

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreHelper.java#L92

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java#L71

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -704,4 +654,47 @@ private Entity backfillToEntity(Backfill backfill) {
...
+      tx.rollback();
+      throw new TransactionException(false, ex);
+    } finally {
+      if (tx.isActive()) {
+        tx.rollback();
+      }
+    }
+  }
+
+  public <E extends Exception> void runConsumerInTransaction(TransactionConsumer<E> f)
```
danielnorberg: Is it necessary to also have this variant?

styx-common/src/main/java/com/spotify/styx/storage/TransactionConsumer.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.styx.storage;
+
+import java.io.IOException;
+
+public interface TransactionConsumer<E extends Exception> {
```
danielnorberg: Is it necessary to also have this interface? Isn't `TransactionFunction` enough for all use cases?

styx-common/src/main/java/com/spotify/styx/storage/TransactionalStorage.java
```diff
@@ -0,0 +1,83 @@
...
+  WorkflowId store(Workflow workflow) throws IOException;
+
+  /**
+   * Updates the next natural trigger for a {@link Workflow}.
+   *
+   * @param workflowId  The {@link WorkflowId} to update the next natural trigger for.
+   * @param triggerSpec The next natural trigger spec describing when the {@link Workflow} should
+   *                    be instantiated.
+   */
+  void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec triggerSpec) throws IOException;
```
danielnorberg: Have these methods return something and avoid the need for `TransactionConsumer`?

danielnorberg:
Seems to me we can easily eliminate `TransactionConsumer` and all associated duplicate code.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -188,69 +195,24 @@ boolean enabled(WorkflowId workflowId) throws IOException {
...
-      final String json = OBJECT_MAPPER.writeValueAsString(workflow);
-      final Key workflowKey = workflowKey(workflow.id());
-      final Optional<Entity> workflowOpt = getOpt(transaction, workflowKey);
-      final Entity workflowEntity = asBuilderOrNew(workflowOpt, workflowKey)
-          .set(PROPERTY_WORKFLOW_JSON, StringValue.newBuilder(json).setExcludeFromIndexes(true).build())
-          .build();
-
-      return transaction.put(workflowEntity);
-    }));
+    runInTransaction(tx -> tx.store(workflow));
```
danielnorberg: How come we're not doing retries here anymore?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -188,69 +195,24 @@ boolean enabled(WorkflowId workflowId) throws IOException {
...
-        throw new ResourceNotFoundException(
-            String.format("%s:%s doesn't exist.", workflowId.componentId(), workflowId.id()));
-      }
-
-      final Entity.Builder builder = Entity
-          .newBuilder(workflowOpt.get())
-          .set(PROPERTY_NEXT_NATURAL_TRIGGER, instantToTimestamp(instant));
-      return transaction.put(builder.build());
-    }));
+    runInTransaction(tx -> tx.updateNextNaturalTrigger(workflowId, triggerSpec));
```
danielnorberg: retries

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -390,27 +352,12 @@ void deleteActiveState(WorkflowInstance workflowInstance) throws IOException {
...
-
-      final Entity.Builder builder = Entity.newBuilder(workflowOpt.get());
-      state.enabled().ifPresent(x -> builder.set(PROPERTY_WORKFLOW_ENABLED, x));
-      state.nextNaturalTrigger()
-          .ifPresent(x -> builder.set(PROPERTY_NEXT_NATURAL_TRIGGER, instantToTimestamp(x)));
-      state.nextNaturalOffsetTrigger()
-          .ifPresent(x -> builder.set(PROPERTY_NEXT_NATURAL_OFFSET_TRIGGER, instantToTimestamp(x)));
-      return transaction.put(builder.build());
-    }));
+    runInTransaction(tx -> tx.patchState(workflowId, state));
```
danielnorberg: retries

fabriziodemaria:
Squashed and rebased.

styx-common/src/test/java/com/spotify/styx/storage/DatastoreTransactionalStorageTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+@RunWith(MockitoJUnitRunner.class)
+public class DatastoreTransactionalStorageTest {
+
+  private static LocalDatastoreHelper helper;
+  private static Datastore datastore;
+
+  @BeforeClass
+  public static void setUpClass() throws Exception {
+    helper = LocalDatastoreHelper.create(1.0); // 100% global consistency
```
ulzha: Any reason to go to max consistency? I think in testing we might want something closer to the opposite, max chaos?

styx-common/src/test/java/com/spotify/styx/storage/StorageTest.java
```diff
@@ -0,0 +1,66 @@
...
+            return "foo";
+        }
+      });
+    } catch (IOException | FooException ignore) {
+      fail(); // Satisfy static analysis rule requiring assert or fail in junit tests
+
+      // The throws declaration of the above lambda should be inferred to FooException, not the
+      // generic Exception. This allows a try/catch wrapping the runInTransaction call to catch
+      // a user defined checked exception type to communicate a business logic exception without
+      // having to do a generic catch Exception.
```
ulzha: :star:

styx-common/src/test/java/com/spotify/styx/storage/DatastoreTransactionalStorageTest.java
fabriziodemaria: Perhaps 100% consistency should be the default testing behaviour and that value decreased for specific tests around strong/weak consistency.

ulzha: Hm, my point was indeed about the default testing behavior, by default we should try to catch things we didn't think of. A piece of code that we _thougt_ wasn't affected by weak consistency, might actually _be_ affected.

styx-common/src/main/java/com/spotify/styx/storage/TransactionalStorage.java
```diff
@@ -0,0 +1,82 @@
...
+package com.spotify.styx.storage;
+
+import com.spotify.styx.model.Workflow;
+import com.spotify.styx.model.WorkflowId;
+import com.spotify.styx.model.WorkflowState;
+import com.spotify.styx.util.TriggerInstantSpec;
+import java.io.IOException;
+
+/**
+ * The interface to the persistence layer where the same transaction can be used across storage
```
ulzha: The use of "can" here... I think the design we have is "the same transaction _is_ used across storage operations", should we document that clearly?

ulzha: The naming choice also made me initially think that a "TransactionalStorage" would manage several transactions. If not, then perhaps let's not hide that it _is a_ (wrapper of a) transaction? `com.spotify.styx.storage.Transaction`? `com.spotify.styx.storage.StorageTransaction`?

See how in `TransactionFunction.apply` the parameter is called `tx`, short for "transaction", yet the type name contradicts that reading, it doesn't read as "FooBarTransaction" but rather `TransactionalStorage`...

Furthermore in `DatastoreStorage` we have `transactionalStorageFactory`, suggesting that we produce different storages out of given storage. At this point I as a reader am confused what such stacking of Storage abstractions means, and it's probably even harder to grasp for new readers... Actually, in common usage terms, the objects that the factory produces are wrappers of _single transactions_ against the same `DatastoreStorage`, so, huh...

ulzha:
The meat of it LGTM.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -793,4 +743,31 @@ private StringValue jsonValue(Object o) throws JsonProcessingException {
...
+  public <T, E extends Exception> T runInTransaction(TransactionFunction<T, E> f)
+      throws IOException, E {
+    final TransactionalStorage tx = newTransaction();
+    try {
+      final T value = f.apply(tx);
+      tx.commit();
+      return value;
+    } catch (DatastoreException ex) {
+      tx.rollback();
+      throw new TransactionException(false, ex);
```
ulzha: Why `false`, I think it can be conflict here as well?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -793,4 +743,31 @@ private StringValue jsonValue(Object o) throws JsonProcessingException {
...
+      }
+    }
+  }
+
+  private TransactionalStorage newTransaction() throws TransactionException {
+    final Transaction transaction;
+    try {
+      transaction = datastore.newTransaction();
+    } catch (DatastoreException e) {
+      throw new TransactionException(false, e);
```
ulzha: Same here - even though paranoid (hard to imagine a `newTransaction` call would "conflict") - if we can easily consistently translate the conflict codes to `true` then we may be saving ourselves some troubleshooting headache later.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -793,4 +743,31 @@ private StringValue jsonValue(Object o) throws JsonProcessingException {
...
+
+  public <T, E extends Exception> T runInTransaction(TransactionFunction<T, E> f)
+      throws IOException, E {
+    final TransactionalStorage tx = newTransaction();
+    try {
+      final T value = f.apply(tx);
+      tx.commit();
+      return value;
+    } catch (DatastoreException ex) {
+      tx.rollback();
```
ulzha: Redundant? There's one in `finally`.

styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -173,10 +173,8 @@
 
   private static LocalDatastoreHelper helper;
   private DatastoreStorage storage;
+  private Datastore datastore;
 
-  @Mock Transaction transaction;
-  @Mock Datastore datastore;
-  @Mock DatastoreStorageTransaction storageTransaction;
```
ulzha: Really nice!

ulzha:
How many likes do you want? :D :+1: :zap:

PR #365, opened 2018-01-19
errorprone static analysis

danielnorberg:
http://errorprone.info/

### DIFF
--- a/styx-api-service/src/test/java/com/spotify/styx/api/ApiVersionTestUtils.java
+++ b/styx-api-service/src/test/java/com/spotify/styx/api/ApiVersionTestUtils.java
@@ -29,6 +29,10 @@ import org.hamcrest.TypeSafeMatcher;
  */
 class ApiVersionTestUtils {
 
+  private ApiVersionTestUtils() {
+    throw new UnsupportedOperationException();
+  }
+
   static Api.Version[] ALL_VERSIONS = Api.Version.values();
 
   /**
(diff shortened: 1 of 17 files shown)


danielnorberg:
@honnix PTAL

danielnorberg:
Not going to go through the effort to get line coverage on those private constructors

honnix:
So we give up codacy? It’s too slow because we don’t pay. :(

danielnorberg:
I was thinking we'd use both =)

danielnorberg:
errorprone is useful because it's part of our local build cycle. Catch issues before pushing and avoid  delayed codacy failures...

styx-api-service/src/test/java/com/spotify/styx/api/ApiVersionTestUtils.java
```diff
@@ -29,6 +29,10 @@
  */
 class ApiVersionTestUtils {
 
+  private ApiVersionTestUtils() {
+    throw new UnsupportedOperationException();
```
honnix: Is this the new best practice? Like making it really impossible to construct an instance. I wish Java had `object`.

danielnorberg: Yeah, from Effective Java: http://www.informit.com/articles/article.aspx?p=1216151&seqNum=4

honnix: Ok. I could only recall the private constructor part not the exception. :)

PR #367, opened 2018-01-22
limit messages - keep 5 latest

danielnorberg:


fabriziodemaria:
I am not sure why we keep a list of messages in the first place. It seems like we always only care of the latest message in the list (both in the CLI and in the logic for issuing `info` events. Perhaps change the field into an `Optional<Message> lastMessage?

danielnorberg:
@fabriziodemaria Sounds like a good idea, I wasn't sure why we're keeping a list of messages either.  But the `StateData` type is also used in the styx client so that might be a breaking change 🤔

danielnorberg:
@fabriziodemaria see https://github.com/spotify/styx/pull/372

PR #370, opened 2018-01-23
scheduler: deterministic rsrc limit msg contents

danielnorberg:


danielnorberg:
@fabriziodemaria PTAL

fabriziodemaria:
`toCollection(TreeSet::new)` could be closer to when the message list is constructed, for easier comments/readability.

PR #371, opened 2018-01-23
move submission -> running timing to k8s runner

danielnorberg:
Make this measurement less dependent on event transitions happening on the same scheduler instance in preparation for HA multiple instance scheduler setup.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -84,8 +84,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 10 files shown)


NatashaL:
LGTM

fabriziodemaria:
+1

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -133,34 +136,36 @@
   private final Debug debug;
   private final int pollPodsIntervalSeconds;
   private final int podDeletionDelaySeconds;
-  private final Clock clock;
+  private final Time time;
+  private final ConcurrentMap<String, Long> submissions = new ConcurrentHashMap<>();
```
ulzha: Hard to read, have to guess the semantics until I read into usage. `submissionTimeByExecutionId` or a comment saying as much would be superb.

ulzha:
:+1:

danielnorberg:
Moved the actual timing code to `MetricsStats`. The `KubernetesDockerRunner` now just calls `reportSubmission()` and `reportRunning()`.

Also replaced the map holding the submission timestamps with a bounded guava cache to ensure that it doesn't grow without bound in case we don't observe pods running.

PR #372, opened 2018-01-23
single (last) message in RunState

danielnorberg:
We're always only using the last message anyway.

As opposed to https://github.com/spotify/styx/pull/369 this should be backwards compatible.


danielnorberg:
@fabriziodemaria PTAL

styx-common/src/test/java/com/spotify/styx/state/RunStateTest.java
```diff
@@ -510,9 +507,7 @@ public void testAccumulatesMessagesWithInfo() throws Exception {
 
     assertThat(
         transitioner.get(WORKFLOW_INSTANCE).data().messages(),
-        contains(
-            Message.info("info message"),
-            Message.warning("warning message")));
+        contains(Message.warning("warning message")));
```
fabriziodemaria: These tests might need renaming, and we should also test the new `.data().message()`

ulzha:
:+1:

PR #373, opened 2018-01-24
log pod UID

danielnorberg:
Allow looking up WFIs for pods spotted in the container engine visualizer (which only shows pod UIDs).


danielnorberg:
lol that this PR increases coverage by breaking a line

fabriziodemaria:
+1

PR #375, opened 2018-01-25
Use fixed timestamp for tests

fabriziodemaria:
This PR also fixes test failures on JDK9 due to the differences applied to `java.time.Clock.systemUTC()` (ref. https://bugs.openjdk.java.net/browse/JDK-8068730)


honnix:
+1

PR #377, opened 2018-01-26
fix codacy complaints

honnix:
Won't bother covering private constructors.


honnix:
@danielnorberg PTAL

honnix:
rebased and squashed. @danielnorberg

PR #378, opened 2018-01-26
Transactional sharded counters

NatashaL:


styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+  private static final long CACHE_EXPIRY_MILLIS = 1000;
+
+  private final Datastore datastore;
+  public static final String KIND_COUNTER_LIMIT = "CounterLimit";
+  public static final String PROPERTY_LIMIT = "limit";
+  public static final String KIND_COUNTER_SHARD = "CounterShard";
+  public static final String PROPERTY_SHARD = "shard";
+
+  private static class CounterSnapshot {
+    Instant updatedAt;
```
danielnorberg: Prefer final fields and initialization in the constructor.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+        snapshot.limit = Long.MAX_VALUE;
+        // Or IllegalStateException("No limit found in Datastore for " + counterId);?
+      } else {
+        snapshot.limit = limit.getLong(PROPERTY_LIMIT);
+      }
+
+      final EntityQuery queryShards = EntityQuery.newEntityQueryBuilder()
+          .setKind(KIND_COUNTER_SHARD)
+          .setFilter(CompositeFilter.and(
+              PropertyFilter.ge("__key__", counterId + "-0"),
```
danielnorberg: Could the counter id be a field on the `CounterShard` entity kind? Then this filter could just be `PropertyFilter.eq("counterId", counterId)`.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+          .setLimit(NUM_SHARDS)
+          .build();
+      final QueryResults<Entity> shards = datastore.run(queryShards);
+      snapshot.shards = new ArrayList<Long>(NUM_SHARDS);
+      int shardCount;
+      for (shardCount = 0; shards.hasNext(); shardCount++) {
+        Entity shard = shards.next();
+        String shardKey = shard.getKey().toString();
+        int shardIndex = Integer.valueOf(shardKey.substring(shardKey.indexOf('-') + 1));
+        snapshot.shards.set(shardIndex, shard.getLong(PROPERTY_SHARD));
```
danielnorberg: As the `shards` list is empty (even though it has `NUM_SHARDS` capacity), calling `set()` will throw `IndexOutOfBoundsException`. 

Call `add()` instead.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * <p>Note that the ShardedCounter stores state in the same Datastore database that the caller of this
+ * class can access, too; guarantees don't apply if something else than ShardedCounter updates the
+ * stored state.
+ */
+public class ShardedCounter {
+
+  // Ought to be enough (parallelism) for everyone. We could make it dynamic with extra effort.
+  private static final int NUM_SHARDS = 128;
+  private static final long CACHE_EXPIRY_MILLIS = 1000;
```
danielnorberg: Make this a strongly typed `Duration` instead of a loosely typed `long` with millis.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+      snapshot.updatedAt = Instant.now();
+      return snapshot;
+    }
+
+    private boolean isRecent() {
+      return updatedAt.plus(CACHE_EXPIRY_MILLIS, ChronoUnit.MILLIS).isAfter(Instant.now());
+    }
+
+    private long shardCapacity(int shardIndex) {
+      return limit / NUM_SHARDS + (shardIndex < limit % NUM_SHARDS ? 1 : 0);
```
danielnorberg: Would be great to have a comment explaining the math here.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+
+    /**
+     * Returns shard index which _likely_ could be successfully updated by delta, according to our
+     * cached view of the state in Datastore.
+     */
+    private int pickShardWithSpareCapacity(long delta) {
+      List<Integer> candidates = new ArrayList<>();
+
+      for (int i = 0; i < shards.size(); i++) {
+        if (shards.get(i) + delta >= 0
```
danielnorberg: Consider doing `Range.closed(0, shardCapacity(i)).contains(shards.get(i) + delta)`, or `shardHasCapacity(i delta)` or at least only computing `shards.get(i) + delta` once and storing it in a variable.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+
+    private long shardCapacity(int shardIndex) {
+      return limit / NUM_SHARDS + (shardIndex < limit % NUM_SHARDS ? 1 : 0);
+    }
+
+    /**
+     * Returns shard index which _likely_ could be successfully updated by delta, according to our
+     * cached view of the state in Datastore.
+     */
+    private int pickShardWithSpareCapacity(long delta) {
```
danielnorberg: Is delta ever not just the number `1` ?

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+
+      for (int i = 0; i < shards.size(); i++) {
+        if (shards.get(i) + delta >= 0
+            && shards.get(i) + delta <= shardCapacity(i)) {
+          candidates.add(i);
+        }
+      }
+
+      if (candidates.isEmpty()) {
+        return new Random().nextInt(NUM_SHARDS);
```
danielnorberg: At this point it looks like the counter does not have the capacity asked for and we might want to give that feedback to the user. If we just return random shard here we lose that information. Do something like throwing a `CounterFullException` to propagate this information to upper layers that can log it?

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  /**
+   * A weakly consistent view of the state in Datastore, refreshed by ShardedCounter on demand.
+   */
+  private Map<String, CounterSnapshot> counterCache;
+
+  public ShardedCounter(Datastore datastore) {
+    this.datastore = Objects.requireNonNull(datastore);
+    this.counterCache = new ConcurrentHashMap<>();
```
danielnorberg: Use a [guava cache](https://github.com/google/guava/wiki/CachesExplained) with some max size to avoid the risk of unbounded growth and memory leak?

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  /**
+   * Returns a recent snapshot.
+   */
+  private CounterSnapshot getCounterSnapshot(String counterId) {
+    if (counterCache.containsKey(counterId) && counterCache.get(counterId).isRecent()) {
```
danielnorberg: This is racy and does unnecessary duplicate lookups. Prefer:

```java
final CounterSnapshot snapshot = counterCache.get(counterId);
if (snapshot != null && snapshot.isRecent()) {
  return snapshot;
}
...
```

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,241 @@
...
+
+  // Ought to be enough (parallelism) for everyone. We could make it dynamic with extra effort.
+  private static final int NUM_SHARDS = 128;
+  private static final long CACHE_EXPIRY_MILLIS = 1000;
+
+  private final Datastore datastore;
+  public static final String KIND_COUNTER_LIMIT = "CounterLimit";
+  public static final String PROPERTY_LIMIT = "limit";
+  public static final String KIND_COUNTER_SHARD = "CounterShard";
+  public static final String PROPERTY_SHARD = "shard";
```
danielnorberg: As this is the *value of the shard*,  should this field be called `value` instead of `shard`?

danielnorberg:
Looking forward to see `ShardedCounterTest` =)

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
ulzha: +1 or -1, in our current usages.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
ulzha: We return a random choice here to later get the "Chosen shard has no spare capacity anymoar" exception (or succeed, with low likelihood), so upper layers do get that information.

(It looks like we'll just adopt the TransactionException from https://github.com/spotify/styx/pull/363 instead of ConcurrentModificationException.)

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
ulzha: +1 to using Guava, but where do you mean there would be unbounded growth? It's growing with the number of distinct resources, and we don't have a deleteCounter yet - did you mean that?

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
ulzha: Yeah I like that too, will update the doc as well to match.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
ulzha: Might throw that exception already here, but opted to do in just one place, to a slight perceived maintainability benefit.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,364 @@
...
+     */
+    private static CounterSnapshot fromDatastore(Datastore datastore, String counterId, Time time) {
+      return new CounterSnapshot(datastore, counterId, time);
+    }
+
+    private static Map<Integer, Long> fetchShards(Datastore datastore, String counterId) {
+      final EntityQuery queryShards = EntityQuery.newEntityQueryBuilder()
+          .setKind(KIND_COUNTER_SHARD)
+          .setFilter(PropertyFilter.eq(PROPERTY_COUNTER_ID, counterId))
+          .setOrderBy(OrderBy.asc("__key__"))
```
ulzha: We can remove this - as you pointed out the lexicographical ordering was of no use at all.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,364 @@
...
+          }
+          return null;
+        });
+      }
+    }
+
+    /**
+     * Gets the counter state. Creates it in Datastore if it doesn't exist yet.
+     */
+    private static CounterSnapshot fromDatastore(Datastore datastore, String counterId, Time time) {
```
ulzha: Dunno if we need this method if we moved the functionality to the constructor.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,364 @@
...
+
+    /**
+     * The shard's capacity is calculated as:
+     * 1/NUM_SHARDS part of the total count capacity
+     * plus an extra 1 unit of the remainder (whenever limit % NUM_SHARDS > 0) depending on the shardIndex
+     * The first (limit % NUM_SHARDS) shards get 1 unit extra, and the rest get 0.
+     *
+     * </p>ex. If limit=5 for a given counter and NUM_SHARDS=3,
+     * then the distribution of capacity between the 3 shards will be [2, 2, 1]
+     */
```
ulzha: :star:

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,364 @@
...
+    private long shardCapacity(int shardIndex) {
+      // The shard's capacity is calculated as:
+      // 1/NUM_SHARDS part of the total count capacity
+      // plus an extra 1 unit of the remainder (whenever limit % NUM_SHARDS > 0) depending on the shardIndex
```
ulzha: ...but probably no need to duplicate the doc in a comment inside...

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,364 @@
...
+          candidates.add(i);
+        } else if (!shards.containsKey(i)) {
+          LOG.warn("Shard {} for counter {} is not present in local cache", i, counterId);
+        }
+      }
+
+      if (candidates.isEmpty()) {
+        if (shards.size() == 0) {
+          throw new InvalidShardStateException(
+              "Trying to operate with a potentially uninitialized counter. Cache needs to be updated first.");
```
ulzha: This is a permitted state in normal operation. We may have updated the cache but still gotten zero shards. Let's not throw... In the earlier review I was still in favor of `return new Random().nextInt(NUM_SHARDS);` because that leads to fewer code paths (harder to make mistakes when maintaining) and only one place to throw CounterCapacityException, in updateCounter.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,364 @@
...
+   * <p>Delta should be +/-1 for graceful behavior, due to how sharding is currently implemented.
+   * Updates with a larger delta are prone to spuriously fail even when the counter is not near to
+   * exceeding its limit. Failures are certain when delta >= limit / NUM_SHARDS + 1.
+   */
+  public void updateCounter(DatastoreReaderWriter transaction, String counterId, long delta) {
+    CounterSnapshot snapshot = getCounterSnapshot(counterId);
+    int shardIndex;
+    try {
+      shardIndex = snapshot.pickShardWithSpareCapacity(delta);
+    } catch (InvalidShardStateException e) {
```
ulzha: As commented on the throw - it's really not invalid, and catching the condition ought to be unneeded unneeded, it's a rabbit hole to more complexity... The repeated `fetchShards` in the constructor ought to be sufficient for reducing the spurious failures at startup that we didn't like in the stress test, but even that might be unnecessary in continued operation. The algorithm as such is really meant to survive all eventually consistent query behavior. Let's clarify that, if needed? And also test some scenarios where it does so?

Writing code to "handle" the eventual consistency makes that fact ever more hidden - be it two refreshes or a loop of refreshes, we still aren't guaranteed to overcome it, but it's extra code to test (hard to test, too).

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,364 @@
...
+            "Invalid shard state for counter {} but cache has already been reloaded for counter. {}",
+            counterId, e1);
+        return;
+      }
+    }
+    final Key shardKey = datastore.newKeyFactory().setKind(KIND_COUNTER_SHARD)
+        .newKey(counterId + "-" + shardIndex);
+    final Entity shard = transaction.get(shardKey);
+
+    final long newCounterValue = shard.getLong(PROPERTY_SHARD_VALUE) + delta;
```
ulzha: newShardValue?

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,364 @@
...
+    if (shard != null && Range.closed(0L, snapshot.shardCapacity(shardIndex))
+        .contains(newCounterValue) && newCounterValue <= getLimit(datastore, counterId)) {
```
ulzha: getLimit comparison seems superfluous - if it was in the range for shardCapacity, then it was less than limit by design.
Or if this was meant to catch a newly decreased limit, then the correct calculation would be a bit like that in shardCapacity... But since we aren't killing jobs when we decrease limit, I'd say this extra logic won't noticeably improve the behavior when we decrease limit anyway.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,364 @@
...
+    }
+    return result;
+  }
+
+  /**
+   * Delete counter by counterId. Deletes both counter shards and counter limit if it exists.
+   *
+   * <p>There may be inconsistencies between instances, as each stores an in-memory cache, whose
+   * state may differ from what's stored in Datastore. Only after a cache expiry will other instances
+   * notice that a counter has been deleted.
```
ulzha: The other methods aren't in fact designed to notice such condition. They will just go and initialize the counter as if it was just coming into existence. Maybe be can say just that is that it's up to the user to ensure there's no instances expecting the counter to exist while deleteCounter is ongoing...

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
```diff
@@ -0,0 +1,299 @@
...
+    });
+
+    //expire cache so that the new limit value gets picked up
+    now = afterCacheExpiryDuration(now);
+
+    //increment counter by 1 until counter value gets to 10
+    while (shardedCounter.getCounter(COUNTER_ID1) < 10) {
+      try {
+        IntStream.range(0, 10).forEach(i -> updateCounterInTransaction(COUNTER_ID1, 1L));
+      } catch (DatastoreException ignored) {
```
ulzha: What were these exceptions about?

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
NatashaL: yes, indeed

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
NatashaL: You're right, it's completely wrong. Got thrown off by the already wrongly named `newCounterValue`

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
NatashaL: This is made to catch exceptions caused by collisions, or bad shard selection, or simply exceptions caused by no capacity left in a counter. The goal is to fill up all capacity of the counter (in this case capacity limit is set to 10). It runs in loops to make sure the we get up to the max capacity of the counter, and not fail early due to selecting a full shard.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
NatashaL: True true

NatashaL:
@ulzha PTAL

Note: there's still more to be done around the usage of `ShardedCounter`

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,339 @@
...
+ * class can access, too; guarantees don't apply if something else than ShardedCounter updates the
+ * stored state.
+ */
+public class ShardedCounter {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ShardedCounter.class);
+
+  // Ought to be enough (parallelism) for everyone. We could make it dynamic with extra effort.
+  public static final int NUM_SHARDS = 128;
+  public static final Instant CACHE_EXPIRY_DURATION = Instant.ofEpochMilli(1000);
```
ulzha: The suggestion about `Duration` was referring to https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html . `Instant` makes an ill suited type for CACHE_EXPIRY_DURATION

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
ulzha: Ah I misread the line above as doing _one_ increment.

If it did so then, then exceptions would not occur, I think? (We bump `now` in each iteration, and 100% consistency is used in `helper`, and no other tests run in parallel, so in that setup collisions and bad shard selection, and exceeded capacity are not expected to happen by design.)

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,339 @@
...
+      final long newShardValue = shard.getLong(PROPERTY_SHARD_VALUE) + delta;
+      if (Range.closed(0L, snapshot.shardCapacity(shardIndex))
+          .contains(newShardValue)) {
+        transaction.put(Entity.newBuilder(shard)
+                            .set(PROPERTY_SHARD_VALUE,
+                                 shard.getLong(PROPERTY_SHARD_VALUE) + delta)
+                            .build());
+      } else {
+        // TODO fail (rollback) the transaction
+        throw new ConcurrentModificationException(
```
ulzha: I mean, this here should be the place we throw `CounterCapacityException` (and make sure that it's handled by styx in a way that rolls back the entire transaction).

And if we avoid throwing it in multiple places, then we have less code and more consistent execution paths.

ulzha: I used ConcurrentModificationException just as a placeholder, the CounterCapacityException is a better class for it.

And maybe we'll want it to inherit from TransactionException (from https://github.com/spotify/styx/pull/363) when we try to integrate the PRs...

ulzha: Or if `StorageTransaction` has a better way to signal need to abort, then we could use that instead of throwing exceptions, but there is none yet.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,339 @@
...
+            "Chosen shard has no spare capacity anymoar, pls rollback kths");
+      }
+    } else {
+      throw new ShardNotFoundException(shardKey.getName());
```
ulzha: :star: 

Let's add a message and/or comment though, ~"unexpected Datastore behavior or our bug - the code should've run `initialize()` before reaching this point, and all shards should strongly consistently appear in transactions thereafter"

ulzha:
I think let's merge when the usage is just doing `updateCounter` parts (and the existing in-memory limit enforcement is still in place) - and completely migrate to ShardedCounter usage in a separate PR...

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,342 @@
...
+     */
+    private int pickShardWithSpareCapacity(long delta) {
+      List<Integer> candidates = new ArrayList<>();
+
+      for (int i : shards.keySet()) {
+        if (shards.containsKey(i) && Range
+            .closed(0L, shardCapacity(i)).contains(shards.get(i) + delta)) {
+          candidates.add(i);
+        } else if (!shards.containsKey(i)) {
+          LOG.warn("Shard {} for counter {} is not present in local cache", i, counterId);
```
ulzha: This can happen in normal operation due to eventual consistency. So I think we shouldn't warn about that, better avoid warning fatigue. If the counter doesn't get out of this state for a prolonged period, _then_ we or Datastore have a bug or incident, that would be worth warning about, but detecting that condition requires more code and investment, and as such should better be decoupled from the counter algorithm implementation...

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,342 @@
...
+
+  private final Time time;
+  private final Datastore datastore;
+
+  /**
+   * A weakly consistent view of the state in Datastore, refreshed by ShardedCounter on demand.
+   */
+  private final Cache<String, CounterSnapshot> cache = CacheBuilder.newBuilder()
+      .maximumSize(100_000)
+      .build();
```
ulzha: We can also get rid of our isRecent code, because this does apparently have support for the same TTL behavior, with [`expireAfterWrite`](https://google.github.io/guava/releases/16.0/api/docs/com/google/common/cache/CacheBuilder.html#expireAfterWrite(long, java.util.concurrent.TimeUnit)).

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -0,0 +1,342 @@
...
+  public static final String PROPERTY_COUNTER_ID = "counterId";
+
+  private final Time time;
+  private final Datastore datastore;
+
+  /**
+   * A weakly consistent view of the state in Datastore, refreshed by ShardedCounter on demand.
+   */
+  private final Cache<String, CounterSnapshot> cache = CacheBuilder.newBuilder()
+      .maximumSize(100_000)
```
ulzha: Going to remove this cap again, I think. The data is small after all. We won't suffer from having all shards in memory. We might though suffer from having an arbitrary cap on the cache size - as it is reached, we lose the fixed cap on the Datastore request rate :)

ulzha: We could call it something else than cache though, because I think people assume "cache" means "meant to hold a subset of all the things".

NatashaL: Agree about the cache size limit.

For the name, even with that assumption that people might have, it's not far from the truth. We won't *always* have *all* the things, only those that have recently been requested and not expired yet. Once expired, an entry will be refreshed only when it is requested again. So in those terms it's simply a most-recently-used type of cache.

NatashaL:
@ulzha ptal

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -229,8 +229,16 @@ private void limitAndDequeue(
    *
    * @return true if the state consumes resources, otherwise false.
    */
-  private static boolean isConsumingResources(State state) {
-    return state != State.NEW && state != State.QUEUED;
+  public static boolean isConsumingResources(State state) {
```
ulzha: Pretty nice :star:

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -229,8 +229,16 @@ private void limitAndDequeue(
...
+    return !javaslang.collection.List.of(State.NEW, State.QUEUED).contains(state);
+  }
+
+
+  private static boolean isReleasingResources(State state) {
```
ulzha: Pretty nice :star:

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -228,6 +229,36 @@ private void initialize(WorkflowInstance workflowInstance) {
...
+        && nextRunState.data().resourceIds().isPresent()) {
+      // decrement counters if transitioning from a state that consumes resources
+      // to a state that releases the same resources
+      for (String resource : nextRunState.data().resourceIds().get()) {
+        tx.updateCounter(shardedCounter, resource, -1);
+      }
+    }
+  }
+
+  private static boolean isReleasingResources(State state) {
```
ulzha: Whoops, duplicated?

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -228,6 +229,36 @@ private void initialize(WorkflowInstance workflowInstance) {
...
+    // increment counters if event is dequeue
+    if (isDequeue(event, nextRunState) && nextRunState.data().resourceIds().isPresent()) {
+      for (String resource : nextRunState.data().resourceIds().get()) {
+        tx.updateCounter(shardedCounter, resource, 1);
+      }
+    }
+
+    if (isConsumingResources(currentRunState.state())
+        && isReleasingResources(nextRunState.state())
+        && nextRunState.data().resourceIds().isPresent()) {
```
ulzha: Best practice would be to assert that resourceIds().isPresent. (Log an error if it isn't; it's a condition that should hold by design and if it doesn't we are in trouble. If we just express it as a benign if, the maintainer of code will think that it can be absent and will eventually go down odd paths when troubleshooting...)

ulzha: Or, hm, do we encounter cases when it's absent, during replay of legacy events or something?

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
NatashaL: yes

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
NatashaL: I couldn't come to a conclusion when following the flow, but to be safe, we can add a logline now. After deployment we can observe to see if we see those loglines on any cases where the resources might be empty, and use that to figure out what we have missed. If we don't see it happening, it should be safe to assume that resources are present everywhere.

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
```diff
@@ -616,4 +589,125 @@ public void shouldThrowRuntimeException() throws Exception {
...
+
+  @Test
+  public void shouldNotUpdateResourceCountersOnStarted() throws Exception {
+    givenState(INSTANCE, State.SUBMITTED);
+    receiveEvent(Event.started(INSTANCE));
+    verify(transaction, never()).updateCounter(eq(shardedCounter), anyString(), anyInt());
+  }
+
+  @Test
+  public void shouldNotUpdateResourceCountersOnTerminate() throws Exception {
```
ulzha: Mispaste. `shouldUpdate`

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
```diff
@@ -616,4 +589,125 @@ public void shouldThrowRuntimeException() throws Exception {
...
+  }
+
+
+  public void givenState(WorkflowInstance instance, State state) throws IOException {
+    when(transaction.activeState(instance))
+        .thenReturn(Optional.of(PersistentWorkflowInstanceState.builder()
+                                    .counter(17)
+                                    .timestamp(NOW.minusMillis(1))
+                                    .state(state)
+                                    .data(STATE_DATA_1)
```
ulzha: This will sometimes set resourceIds in a state where they're normally not set (QUEUED)... Not sure if it's a problem.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -244,6 +244,8 @@ private void updateResourceCounters(StorageTransaction tx, Event event,
       for (String resource : nextRunState.data().resourceIds().get()) {
         tx.updateCounter(shardedCounter, resource, -1);
       }
+    } else if (!nextRunState.data().resourceIds().isPresent()) {
+      LOG.error("Resource ids are missing for {}.", nextRunState.workflowInstance());
```
ulzha: Log also the currentRunState and nextRunState?

ulzha:
Yay, only those minor things and otherwise LGTM. Let's rock?

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
NatashaL: It shouldn't be a problem. The reason I've done it, is because if I set data to `StateData.zero()` that will make resourceids be empty, which will fail the `.isPresent()` check and won't make the `.updateCounter` call. This is important for the negative tests: I'm making sure the resource ids are there, and the only reason `.updateCounter` won't be called is due to the type of transition that is taking place (i.e. the condition boils down to `isConsuming` and `!isConsuming` check and the event type for the `dequeue` case) .

ulzha:
:+1:

PR #379, opened 2018-01-29
nicer --json ls keys

danielnorberg:
It's a bit weird to emit a json object with `WorkflowId.toString()` as keys.

Before: 
```
$ styx --json ls
{
  "WorkflowId{componentId=foo-component, id=foo-workflow}": [ ... ],
  "WorkflowId{componentId=bar-component, id=bar-workflow}": [ ... ],
  ...
}
```

After: 
```
$ styx --json ls
{
  "foo-component#foo-workflow": [ ... ],
  "bar-component#bar-workflow": [ ... ],
  ...
}
```

### DIFF
--- a/styx-cli/src/main/java/com/spotify/styx/cli/JsonCliOutput.java
+++ b/styx-cli/src/main/java/com/spotify/styx/cli/JsonCliOutput.java
@@ -20,6 +20,8 @@
 
 package com.spotify.styx.cli;
 
+import static java.util.stream.Collectors.toMap;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
@@ -32,6 +34,7 @@ import com.spotify.styx.model.WorkflowState;
 import com.spotify.styx.model.data.EventInfo;
 import com.spotify.styx.serialization.Json;
 import java.util.List;
+import java.util.Map.Entry;
 
 /**
  * Cli output printer that prints json output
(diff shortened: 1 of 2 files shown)


danielnorberg:
@Zatine PTAL

styx-cli/src/main/java/com/spotify/styx/cli/JsonCliOutput.java
```diff
@@ -48,7 +51,8 @@ private static void printJson(Object object) {
 
   @Override
   public void printStates(RunStateDataPayload runStateDataPayload) {
-    printJson(CliUtil.groupStates(runStateDataPayload.activeStates()));
+    printJson(CliUtil.groupStates(runStateDataPayload.activeStates()).entrySet().stream()
+        .collect(toMap(e -> e.getKey().toKey(), Entry::getValue)));
```
Zatine: Slightly confused as to what is going on here (but I think it's just my lack of Java knowledge). What is happening in `.collect`? Is that where we transform `"WorkflowId{componentId=foo-component, id=foo-workflow}"` to `"foo-component#foo-workflow"`?

Zatine:
Looks a lot better! :+1:
Added a small comment mostly to increase my own understanding. :)

styx-cli/src/main/java/com/spotify/styx/cli/JsonCliOutput.java
honnix: @Zatine You just happened to discover the dark side of Java lambda: it is soooo verbose. This essentially creates a map keyed on `e.getKey().toKey()`, mapped to value `e.getValue()`.

Zatine: Ah, thanks. My mistake was looking at the `getKey().toKey()` in general contexts online when it was actually methods on the WorkflowInstance. It's get it now!

PR #380, opened 2018-01-30
[WIP] Transactional state transition

fabriziodemaria:
#376 rebased on top of master.


styx-common/src/main/java/com/spotify/styx/state/RunState.java
```diff
@@ -91,61 +90,28 @@ public boolean isTerminal() {
 
   abstract Time time();
 
-  // TODO: remove the unused outputHandler field
-  abstract OutputHandler outputHandler();
```
danielnorberg: \o/

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -322,9 +329,12 @@ public void create(Environment environment) {
 
     warmUpCache(workflowCache, storage);
 
+    // TODO: hack to get around circular reference. Change OutputHandler.transitionInto() to
+    //       take StateManager as argument instead?
```
fabriziodemaria: Is this comment still valid? Moving the OutputHandler into the StateManager seems like a design choice in this PR rather than a workaround.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -339,20 +349,17 @@ public void create(Environment environment) {
 
     final RateLimiter dequeueRateLimiter = RateLimiter.create(DEFAULT_SUBMISSION_RATE_PER_SEC);
 
-    final OutputHandler[] outputHandlers = new OutputHandler[] {
+    outputHandlers.addAll(ImmutableList.of(
```
fabriziodemaria: Populating `outputHandlers` right after variable declaration would be a bit clearer.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -371,13 +378,15 @@ public void create(Environment environment) {
 
     final Cleaner cleaner = new Cleaner(dockerRunner);
 
-    restoreState(storage, outputHandlers, stateManager, dockerRunner);
+    // TODO: is this still needed?
+    dockerRunner.restore();
```
fabriziodemaria: This should still be needed for the reasons explained here https://github.com/spotify/styx/pull/380/files#diff-a3fabd4f5e1e00cf48f40dfdfa3356f1L463

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -584,10 +564,6 @@ private void setupMetrics(
       StateManager stateManager,
       BiConsumer<Optional<Workflow>, Optional<Workflow>> workflowConsumer) {
     return (workflow) -> {
-      stats.registerActiveStatesMetric(
```
fabriziodemaria: I think this metric can still be useful (also for user facing metrics), any specific reason it was deleted?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
```diff
@@ -131,4 +143,34 @@ public WorkflowId patchState(WorkflowId workflowId, WorkflowState state) throws
 
     return workflowId;
   }
+
+  @Override
+  public Optional<PersistentWorkflowInstanceState> activeState(WorkflowInstance instance) throws IOException {
```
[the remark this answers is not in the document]
danielnorberg: We could also refactor these methods into e.g. static methods acting on the `DatastoreReaderWriter` interface, which is implemented by both `Transaction` and `Datastore`.

As for always using transactions, it does involve  `beginTransaction` and `commit` RPCs, with associated overhead. Not sure what the performance implications will be beyond that.

The only real use of this method that I could find is by `KubernetesDockerRunner.lookupPodRunState()`, and I think it might be good to move that into a transaction. The current behavior of (outside a transaction) reading the current state from datastore and then emitting events based on that seems inherently racy.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: The circular dependency is still there. The code is passing in the list of output handlers into the `QueuedStateManager` constructor, and then passing in the `QueuedStateManager` instance into the constructors of the output handlers.

As the comment says, I would suggest breaking this cycle by not passing in the `QueuedStateManager` instance into the output handler constructors. Instead have the state manager pass itself as an argument to `OutputHandler.transitionInto()`.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: See above.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: Should restore that comment.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: Didn't see that we were using it anywhere. We can re-add it. Preferably as a cheap datastore query or something to avoid excessive per-active instance entity reads just for the sake of this metric.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -60,391 +58,227 @@
...
+      final long counter = storage.getLatestStoredCounter(workflowInstance).orElse(NO_EVENTS_PROCESSED);
+      nextCounter = counter + 1;
     } catch (IOException e) {
-      throw Throwables.propagate(e);
+      return CompletableFutures.exceptionallyCompletedFuture(e);
     }
 
-    states.computeIfAbsent(workflowInstance, (wfi) -> new InstanceState(wfi, runState, counter + 1));
-  }
+    // TODO: optional retry on transaction conflict
```
fabriziodemaria: I don't think we want to retry on conflict. Edge cases might cause the same event to be processed twice. For example:
- `dequeue` event transition fails with conflict (and succeeds on another instance);
- instance is slow before retrying. In the meantime the workflow instance runs and gets back to QUEUED state;
- the previous `dequeue` event is retried and instance gets dequeued a second time (skipping the retry delay, etc.); 

EDIT: I wonder if this can still happen without retries on conflict. Two instances might try to `dequeue` at the same time but one gets slow before even starting the transaction while the other succeeds and a whole execution completes and the state goes back to QUEUED. Afterwards, the slow instance catches up and `dequeue` again (skipping retry delay, etc.).

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesDockerRunnerTest.java
```diff
@@ -509,7 +511,7 @@ public void shouldEnsureAndMountServiceAccountSecret() throws IsClosedException,
         WORKFLOW_INSTANCE.workflowId().toString(), SERVICE_ACCOUNT)).thenReturn(SERVICE_ACCOUNT_SECRET);
 
     StateData stateData = StateData.newBuilder().executionId(POD_NAME).build();
-    stateManager.initialize(RunState.create(WORKFLOW_INSTANCE, RunState.State.SUBMITTED, stateData));
+//    stateManager.trigger(RunState.create(WORKFLOW_INSTANCE, RunState.State.SUBMITTED, stateData), trigger);
```
fabriziodemaria: Delete

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesDockerRunnerTest.java
```diff
@@ -587,41 +591,47 @@ public void shouldNotSendStatsOnOtherError() throws Exception {
...
 
   @Test
   public void shouldIgnoreDeletedEvents() throws Exception {
+    reset(stateManager);
+
+    StateData stateData = StateData.newBuilder().executionId(POD_NAME).build();
+    RunState runState = RunState.create(WORKFLOW_INSTANCE, State.RUNNING, stateData);
+
+    when(stateManager.activeStates()).thenReturn(ImmutableMap.of(WORKFLOW_INSTANCE, runState));
+    when(stateManager.get(WORKFLOW_INSTANCE)).thenReturn(runState);
```
fabriziodemaria: Reuse code in `setUp`

styx-scheduler-service/src/test/java/com/spotify/styx/docker/KubernetesDockerRunnerTest.java
fabriziodemaria: Delete

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
fabriziodemaria: This already is a datastore query and it is executed per workflow rather than workflow instance. Impact on storage should be acceptable. 
However, for this best-effort situations that have low requirements on states consistency we could implement a "slow" caching in memory. For now I am just going to add back the metric.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
fabriziodemaria: Gotcha. Unfortunately passing StateManager as argument to `OutputHandler.transitionInto()` is also not great since `OutputHandler` interface should have no relations with `StateManager`. Indeed, some `OutputHandlers` don't interact with any `StateManager`.

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
```diff
@@ -105,285 +114,232 @@ public void tearDown() throws Exception {
...
+
+    verify(transaction).insertActiveState(INSTANCE, PersistentWorkflowInstanceState.builder()
+        .counter(18)
+        .state(State.QUEUED)
+        .data(StateData.newBuilder().trigger(TRIGGER1).build())
+        .timestamp(NOW)
+        .build());
+
+    verify(storage).writeEvent(
+        SequenceEvent.create(Event.triggerExecution(INSTANCE, TRIGGER1), 18, NOW.toEpochMilli()));
```
fabriziodemaria: This should be counter 19

styx-common/src/main/java/com/spotify/styx/state/RunState.java
```diff
@@ -89,49 +88,28 @@ public boolean isTerminal() {
   public abstract long timestamp();
   public abstract StateData data();
 
-  abstract Time time();
-  abstract OutputHandler outputHandler();
```
ulzha: :star: great to get rid of these

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
```diff
@@ -92,6 +96,14 @@ public WorkflowId store(Workflow workflow) throws IOException {
     return workflow.id();
   }
 
+  @Override
+  public Optional<Workflow> workflow(WorkflowId workflowId) throws IOException {
+    return DatastoreStorage.getOpt(tx, DatastoreStorage.workflowKey(tx.getDatastore().newKeyFactory(), workflowId))
+        .filter(e -> e.contains(PROPERTY_WORKFLOW_JSON))
```
ulzha: Unneeded now?

styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
```diff
@@ -91,6 +91,12 @@ public StyxConfig config() throws IOException {
     return datastoreStorage.activeStates(componentId);
   }
 
+  @Override
+  public Optional<PersistentWorkflowInstanceState> readActiveWorkflowInstance(WorkflowInstance workflowInstance)
```
ulzha: Seems we should rename `WorkflowInstance` to `WorkflowInstanceId`. (Not for this PR.) It's confusing to see methods named `readFooBar` literally take `FooBar` as their argument.

styx-common/src/main/java/com/spotify/styx/storage/StorageTransaction.java
```diff
@@ -42,6 +45,14 @@
    */
   WorkflowId store(Workflow workflow) throws IOException;
 
+  /**
+   * Get a {@link Workflow} definition.
+   *
+   * @param workflowId  The workflow to get
+   * @return Optionally a workflow, if one was found for te given id
```
ulzha: Typo "te"

styx-common/src/main/java/com/spotify/styx/storage/TransactionException.java
```diff
@@ -34,7 +34,25 @@ public TransactionException(boolean conflict, Throwable cause) {
...
   }
 
+  // TODO: represent the failure cause using an enum instead
+
   public boolean isConflict() {
     return conflict;
   }
+
+  public boolean isAlreadyExists() {
+    // TODO: This is a horrible hack to paper over difference between datastore emulator and real service behavior
```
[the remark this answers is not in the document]
ulzha: Include a link to https://github.com/GoogleCloudPlatform/google-cloud-datastore/issues/191 as a more concrete description of the difference?

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -252,21 +253,31 @@ private StatusType eventInjectorHelper(Event event) {
       return Response.forStatus(BAD_REQUEST.withReasonPhrase(e.getMessage()));
     }
 
-    // Verifying active
```
ulzha: Why is this verification removed?

fabriziodemaria:
Continued by #385

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
fabriziodemaria: The `triggerListener.event()` returned future we wait for (a few lines after this) will throw if the instance is already active

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
fabriziodemaria: Probably. Or a missing `PROPERTY_WORKFLOW_JSON` should actually throw an IOException instead of returning an empty optional, since the field is always expected.

Issue #217, opened 2017-05-30
Fix log level according to #212


honnix:
According to #212. We agreed that user error shall be logged as `info`; events, state machine transition, etc. shall be logged as `debug`.

PR #382, opened 2018-02-01
according to issue #217

honnix:
those logs shall be debug instead of info.

@fabriziodemaria PTAL, we have enabled debug logging, so...

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
@@ -37,6 +37,7 @@ public final class ReplayEvents {
     throw new UnsupportedOperationException();
   }
 
+  // TODO: fix NPath complexity
   public static Optional<RunState> getBackfillRunState(
       WorkflowInstance workflowInstance,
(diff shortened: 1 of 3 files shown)


honnix:
Would this change break some of our metrics? Because debug logs will not be shipped to stackdriver.

fabriziodemaria:
I'd remove `transitionLogger` completely, since we log already from here:
https://github.com/spotify/styx/blob/1ebd23a9ff4ff17234eb97ef79f54efa6c968d28/styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java#L106

Hopefully this will also enable us to move `OutputHandler` class out from `styx-common` module.

honnix:
interesting `transitionLogger` is not only for replaying, but happen to be in `ReplayEvents`.

honnix:
Let's first decouple it from `ReplayEvents`. Moving `OutputHandler` requires quite some change.

honnix:
@fabriziodemaria Please take another look.

honnix:
@fabriziodemaria Would this break any Stackdriver metrics?

honnix:
`info` might still be needed. close for now, but factoring out `transitionLogger` is better to have.

PR #381, opened 2018-01-31
striped in-order event processing

danielnorberg:
Use a [`StripedExecutorService`](https://raw.githubusercontent.com/kabutz/javaspecialists/master/src/main/java/eu/javaspecialists/tjsn/concurrency/stripedexecutor/StripedExecutorService.java) to get in-order event processing in the `QueuedStateManager`.

PR #385, opened 2018-02-07
Transactional state transition

fabriziodemaria:
Continuation of #380. Rebased on top of master and rearranged/squashed commits.


styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -551,6 +537,7 @@ private void setupMetrics(
             .filter(workflow -> workflow.configuration().dockerTerminationLogging())
             .count());
 
+    final Map<WorkflowInstance, RunState> activeStates = stateManager.activeStates();
```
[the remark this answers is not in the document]
fabriziodemaria: Fixed

styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTransactionTest.java
```diff
@@ -123,6 +170,132 @@ public void shouldThrowIfTransactionFailed() throws TransactionException {
       fail("Expected exception!");
     } catch (TransactionException e) {
       assertTrue(e.isConflict());
+      transaction3.rollback();
+    }
+  }
+
+  @Test
+  public void insertActiveStateShouldFailIfAlreadyExists() throws Exception {
+    final DatastoreStorage storage = new DatastoreStorage(datastore, Duration.ZERO);
```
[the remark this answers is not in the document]
fabriziodemaria: Partially fixed to remove `storage.runInTransaction` calls, to explicitly call the methods from DatastoreStorageTransaction.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -338,82 +343,101 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
...
 
-        persistentState
-            .state(state)
-            .data(data)
-            .timestamp(Instant.ofEpochMilli(timestamp));
-      }
+    return mapBuilder.build();
+  }
+
+  Optional<PersistentWorkflowInstanceState> activeState(WorkflowInstance instance) throws IOException {
```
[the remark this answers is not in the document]
fabriziodemaria: Added

fabriziodemaria:
Tested this PR in a staging environment, observing sporadic`java.lang.IllegalStateException: ........ received terminate while in SUBMITTED` when execution is very fast (~1 second). 
Perhaps due to missing in-order processing of queued events (ref. #381)

fabriziodemaria:
Also observing `received dequeue while in PREPARE`, probably due to the `Scheduler` tick happening twice before the `dequeue` event is completely processed and Datastore updated. Overall this makes me think we should always wait for the future from `receive` method to complete when issuing a new event.

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -252,21 +253,31 @@ private StatusType eventInjectorHelper(Event event) {
...
+    final CompletionStage<Void> triggered = triggerListener.event(workflow, Trigger.adhoc(triggerId), instant);
+
+    // TODO: return future instead of blocking
+    try {
+      triggered.toCompletableFuture().get();
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new RuntimeException(e);
+    } catch (ExecutionException e) {
+      final Throwable cause = e.getCause();
```
ulzha: In the test, `shouldFailTriggerWFIfAlreadyActive`, we use `Throwables.getRootCause` - should that be the same here?

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -252,21 +253,31 @@ private StatusType eventInjectorHelper(Event event) {
...
+      if (cause instanceof IllegalStateException
+          || cause instanceof IllegalArgumentException) {
+        // TODO: propagate error information using a more specific exception type
+        return Response.forStatus(BAD_REQUEST.withReasonPhrase(cause.getMessage()));
```
ulzha: Should use `CONFLICT` for `IllegalStateException`? It was not a malformed request.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -60,132 +59,203 @@
...
     try {
       counter = storage.getLatestStoredCounter(workflowInstance).orElse(NO_EVENTS_PROCESSED);
     } catch (IOException e) {
-      throw Throwables.propagate(e);
+      return CompletableFutures.exceptionallyCompletedFuture(e);
     }
 
-    states.computeIfAbsent(workflowInstance, (wfi) -> new InstanceState(wfi, runState, counter + 1));
-  }
+    // TODO: optional retry on transaction conflict
```
ulzha: In what scenario would trigger lead to a conflict? I think this is a questionable TODO, the API caller only knows how long they'd like to retry.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -60,132 +59,203 @@
...
+            throw new IllegalArgumentException("Workflow not found: " + workflowInstance.workflowId().toKey());
+          }
+          return tx.insertActiveState(workflowInstance, PersistentWorkflowInstanceState.of(runState, counter));
+        });
+      } catch (TransactionException e) {
+        if (e.isAlreadyExists()) {
+          throw new IllegalStateException("Workflow instance is already triggered: " + workflowInstance.toKey());
+        } else if (e.isConflict()) {
+          LOG.debug("Transactional conflict, abort triggering Workflow instance: " + workflowInstance.toKey());
+          throw new RuntimeException(e);
```
ulzha: A bit worried what this will look like to the API caller. INTERNAL_SERVER_ERROR, that seems not really right. Something like CONFLICT would be kind of informative as to the root cause, as well as lead users to retrying instead of poking DataEx with bug reports.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -60,132 +59,203 @@
...
-      LOG.warn(message);
-      return exceptionallyCompletedFuture(new IllegalArgumentException(message));
-    }
+    // TODO: optional retry on transaction conflict
 
-    CompletionStage<Void> processed = state.transition(event);
-    signalDispatcher();
-    return processed;
-  }
+    // TODO: run on striped executor to get event execution in order.
```
ulzha: How important is this? Sounds quite important, like a FIXME not just TODO? Or maybe we're fine as is, but I don't know what safeguard lets us claim that?

honnix:
I think it is probably fine we ignore those `IllegalStateException` or we handle a few repeatable events (dequeue, timeout, and?) specially.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -60,132 +59,203 @@
...
+    // TODO: optional retry on transaction conflict
 
-  @Override
-  public void restore(RunState runState, long count) {
-    final WorkflowInstance workflowInstance = runState.workflowInstance();
-    if (states.containsKey(workflowInstance)) {
-      throw new RuntimeException("RunState initialization called on active instance "
-                                 + workflowInstance.toKey());
-    }
+    return CompletableFuture.supplyAsync(() -> {
```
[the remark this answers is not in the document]
honnix: Actually, why not just `CompletableFuture.runAsync(...).thenRunAsync(...)`? Then we don't need to return `null` and don't need to accept `nil`.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -60,132 +59,203 @@
...
+              final SequenceEvent sequenceEvent =
+                  SequenceEvent.create(event, nextCounter, nextRunState.timestamp());
+
+              return Tuple.of(sequenceEvent, nextRunState);
+            });
+          } catch (IOException e) {
+            throw new RuntimeException(e);
+          }
+        }, eventTransitionExecutor);
+    return processedTransition.thenComposeAsync((emitEventInfoTuple) -> {
```
honnix: We can do `thenAcceptAsync`.

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
fabriziodemaria: Turns out `shouldFailTriggerWFIfAlreadyActive` test was not exactly right. Fixed.

honnix:
Good to go for now. Let's squash and do a proper release. +1

ulzha:
A thought that bugs me: even with StripedExecutorService, I guess needless illegal states will be encountered if we submit one state transition several times (e.g. once in one tick and then again in the subsequent tick because it has been sitting in the executor's queue and hasn't actually transitioned). I haven't inspected carefully enough - do you know if/where do we prevent that from happening?

honnix:
It does happen and we noticed in staging test. None of them is really harmful because they are really just illegal transitions. One solution we talked about could be waiting for the future to complete, but that would cause much delay for most cases.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -60,391 +59,239 @@
...
-    }
-
-    states.computeIfAbsent(workflowInstance, (wfi) -> new InstanceState(wfi, runState, count + 1));
+    return CompletableFuture.runAsync(() -> initialize(workflowInstance)).thenCompose((ignore) -> {
+      final Event event = Event.triggerExecution(workflowInstance, trigger);
+      try {
+        return receive(event);
+      } catch (IsClosedException isClosedException) {
+        LOG.warn("Failed to send 'triggerExecution' event", isClosedException);
+        // Best effort attempt to rollback the creation of the NEW state
```
danielnorberg: Earlier this PR directly wrote the `QUEUED` state. What is the benefit of first writing `NEW` ?

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -98,10 +98,9 @@ public QueuedStateManager(
   public CompletionStage<Void> trigger(WorkflowInstance workflowInstance, Trigger trigger) throws IsClosedException {
     ensureRunning();
 
-    final long nextCounter;
+    final long counter;
     try {
-      final long counter = storage.getLatestStoredCounter(workflowInstance).orElse(NO_EVENTS_PROCESSED);
-      nextCounter = counter + 1;
+      counter = storage.getLatestStoredCounter(workflowInstance).orElse(NO_EVENTS_PROCESSED);
```
danielnorberg: Looks like we lost a counter increment here. So a retrigger of a WFI will cause the last counter value to be reused.

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
```diff
@@ -112,6 +117,36 @@ public void tearDown() throws Exception {
     }
   }
 
+  @Test
+  public void shouldInitializeAndTriggerWFInstance() throws Exception {
+    final PersistentWorkflowInstanceState instanceStateFresh =
+        INSTANCE_NEW_STATE.toBuilder().counter(-1).build();
```
danielnorberg: `-1` 😅

PR #386, opened 2018-02-07
do not block for 30 seconds

honnix:
not sure what would be a good value.

also make it a smaller batch.

@ulzha PTAL


ulzha:
Related to https://github.com/spotify/styx/pull/291#discussion_r161894099

PR #388, opened 2018-02-13
striped in-order event processing

fabriziodemaria:
#381 rebased


honnix:
https://stackoverflow.com/questions/36343149/mockito-difference-between-dothrow-and-thenthrow
`spy` is different.

fabriziodemaria:
+1

PR #389, opened 2018-02-14
guarded backfill triggering

honnix:
When triggering one backfill fails, continue triggering others. `InterruptedException` is not handled specially any more because when that happens executor thread is stopping.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
@@ -22,6 +22,7 @@ package com.spotify.styx;
 
 import static com.google.common.base.CaseFormat.LOWER_UNDERSCORE;
 import static com.google.common.base.CaseFormat.UPPER_CAMEL;
+import static com.spotify.styx.util.GuardedRunnable.guard;
 import static com.spotify.styx.util.TimeUtil.nextInstant;
 import static java.util.stream.Collectors.counting;
 import static java.util.stream.Collectors.groupingBy;
@@ -92,7 +93,7 @@ class BackfillTriggerManager {
     }
 
     final Map<String, Long> backfillStates = getBackfillStates();
-    backfills.forEach(backfill -> triggerBackfill(backfill, backfillStates));
+    backfills.forEach(backfill -> guard(() -> triggerBackfill(backfill, backfillStates)).run());
 
     final long durationMillis = t0.until(time.get(), ChronoUnit.MILLIS);
     stats.recordTickDuration(TICK_TYPE, durationMillis);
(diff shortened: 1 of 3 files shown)


honnix:
@fabriziodemaria PTAL

styx-scheduler-service/src/test/java/com/spotify/styx/BackfillTriggerManagerTest.java
```diff
@@ -282,28 +280,53 @@ public void shouldTriggerAndUpdateBackfillsSynchronously() throws Exception {
   }
 
   @Test
-  public void shouldNotContinueTriggeringBackfillsIfTriggerFails() throws Exception {
```
ulzha: Whaaa did we have a test to assert _that_?!

honnix: :) so it was by design.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -123,15 +124,13 @@ private void triggerBackfill(Backfill backfill, Map<String, Long> backfillStates
...
-        throw new RuntimeException(e);
+        LOG.warn("tried to trigger backfill for already active state [{}]: {}", partition, backfill);
       } catch (ExecutionException e) {
-        LOG.error("failed to trigger backfill for state [{}]: {}",
-                  partition, backfill);
-        throw new RuntimeException(e);
+        LOG.warn("failed to trigger backfill for state [{}]: {}", partition, backfill, e);
+        return;
+      } catch (Throwable e) {
+        LOG.warn("backfill triggering threw exception for state [{}]: {}", partition, backfill, e);
```
ulzha: Isn't it confusing to have distinct warnings for `ExecutionException` and `Throwable`? The rephrase of Throwable into "threw exception" isn't helping either...

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -123,15 +124,13 @@ private void triggerBackfill(Backfill backfill, Map<String, Long> backfillStates
         // Wait for the trigger execution to complete before proceeding to the next partition
         processed.get();
       } catch (AlreadyInitializedException e) {
-        LOG.warn("tried to trigger backfill for already active state [{}]: {}",
-                 partition, backfill);
-      } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
-        throw new RuntimeException(e);
+        LOG.warn("tried to trigger backfill for already active state [{}]: {}", partition, backfill);
```
ulzha: Should this even be a warning? I suggest info or lower according to https://github.com/spotify/styx/blob/master/CONTRIBUTING.md ... And in HA (redundant instances) setup I guess we'll have these situations constantly (conflicts when another instance already did something). Would be great to not fill logs with noise because of that.

honnix: Agree, let's remove this.

ulzha:
IRLed a bit about cancelling the `InterruptedException` - we think this act of nongracefulness will not immediately hurt. At least it's now consistent with other places where we use `guard`. :+1:

PR #390, opened 2018-02-14
add histograms for important tick methods

honnix:
`TriggerManager`, `BackfillTriggerManager` and `Scheduler`.

`Cleaner` is covered by instrumented `DockerRunner`.

@fabriziodemaria PTAL


styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -74,8 +89,13 @@ void tick() {
       return;
     }
 
+    final Instant t0 = time.get();
```
fabriziodemaria: Should this be the first line in the tick?

fabriziodemaria:
+1

PR #392, opened 2018-02-15
Hot-fix disable activeStatesMetric per Workflow

fabriziodemaria:
After #385, retrieving active states requires a call to persistent layer with considerable higher latencies. In case of high number of registered workflows, `registerActiveStatesMetric` per each workflow could cause the metric framework to be very slow.


ulzha:
+1 - I believe active states per workflow is probably not relied upon anywhere...

PR #393, opened 2018-02-15
trigger in parallel

honnix:
@fabriziodemaria just in case it works.


honnix:
Doesn't seem to be helpful in staging though.

fabriziodemaria:
The improvement should be more visible with the increase of volume of triggers to be processed.

+1

PR #394, opened 2018-02-16
Bump MAX_DOCKER_ARGS_TOTAL

ulzha:
We're bound by Datastore property size limit really, https://cloud.google.com/datastore/docs/concepts/limits . Command lines usually can be 2MB long. It just astonishes users if knocked down by orders of magnitude.


NatashaL:
👍

PR #395, opened 2018-02-16
Strict counter verification in event processing

fabriziodemaria:
This is a early prototype of an idea to improve slow event processing and spot stale events that were enqueued but should not be processed anymore.

We can either supply the expected counter explicitly according to custom logic (for example according to the current state of active states read at the beginning of a tick) or, as a default behaviour, just read the current counter from Datastore at the moment of enqueueing (no transactions involved).

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -236,7 +236,7 @@ public class Scheduler {
 
   private void limitAndDequeue(Map<String, Resource> resources,
       Map<WorkflowId, Set<String>> workflowResourceReferences,
-      Map<String, Long> currentResourceUsage, InstanceState instance,
+      Map<String, Long> currentResourceUsage, InstanceState instanceState,
       CompletionStage<Optional<ExecutionBlocker>> executionBlockerFuture) {
 
     // Check for execution blocker
@@ -248,24 +248,24 @@ public class Scheduler {
       throw new RuntimeException(e);
     } catch (ExecutionException | TimeoutException e) {
       LOG.warn("Failed to check execution blocker for {}, assuming there is no blocker",
-          instance.workflowInstance(), e);
+          instanceState.workflowInstance(), e);
     }
 
     if (blocker.isPresent()) {
-      stateManager.receiveIgnoreClosed(Event.retryAfter(instance.workflowInstance(),
+      stateManager.receiveIgnoreClosed(Event.retryAfter(instanceState.workflowInstance(),
                                                         blocker.get().delay().toMillis()),
-                                       instance.runState().counter());
-      LOG.debug("Dequeue rescheduled: {}: {}", instance.workflowInstance(), blocker.get());
+                                       instanceState.runState().counter());
+      LOG.debug("Dequeue rescheduled: {}: {}", instanceState.workflowInstance(), blocker.get());
       return;
     }
 
     final Set<String> workflowResourceRefs =
-        workflowResourceReferences.getOrDefault(instance.workflowInstance().workflowId(), emptySet());
+        workflowResourceReferences.getOrDefault(instanceState.workflowInstance().workflowId(), emptySet());
 
     // TODO move to ShardedCounter being authoritative for resource accounting
-    final Set<String> instanceResourceRefs = workflowCache.workflow(instance.workflowInstance().workflowId())
+    final Set<String> instanceResourceRefs = workflowCache.workflow(instanceState.workflowInstance().workflowId())
         .map(workflow -> resourceDecorator.decorateResources(
-            instance.runState(), workflow.configuration(), workflowResourceRefs))
+            instanceState.runState(), workflow.configuration(), workflowResourceRefs))
         .orElse(workflowResourceRefs);
 
     final Set<String> unknownResources = instanceResourceRefs.stream()
@@ -286,9 +286,9 @@ public class Scheduler {
 
     if (!unknownResources.isEmpty()) {
       stateManager.receiveIgnoreClosed(
-          Event.runError(instance.workflowInstance(),
+          Event.runError(instanceState.workflowInstance(),
                          String.format("Referenced resources not found: %s", unknownResources)),
-          instance.runState().counter());
+          instanceState.runState().counter());
     } else if (!depletedResources.isEmpty()) {
       final Message message = Message.info(
           String.format("Resource limit reached for: %s",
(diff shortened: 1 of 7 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -181,6 +194,9 @@ private void initialize(WorkflowInstance workflowInstance) {
           throw new IllegalArgumentException(message);
```
fabriziodemaria: Add test for this line

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -120,14 +121,26 @@ public QueuedStateManager(
 
   @Override
   public CompletableFuture<Void> receive(Event event) throws IsClosedException {
+    // Read state counter at enqueueing time
```
fabriziodemaria: Should the default to _not_ check the counter at all and avoid additional Datastore lookups?

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -121,14 +122,26 @@ public QueuedStateManager(
...
+    if (!currentRunState.isPresent()) {
+      return CompletableFutures.exceptionallyCompletedFuture(
+          new IllegalArgumentException("Workflow not found: "
+                                       + event.workflowInstance().workflowId().toKey()));
+    }
+    return receive(event, currentRunState.get().counter());
+  }
+
+  @Override
+  public CompletableFuture<Void> receive(Event event, long expectedCounter) throws IsClosedException {
```
honnix: This seems can stay as a private method instead of sitting in the interface.

styx-scheduler-service/src/main/java/com/spotify/styx/state/StateManager.java
```diff
@@ -71,6 +81,20 @@ default void receiveIgnoreClosed(Event event) {
     }
   }
 
+  /**
+   * Like {@link #receive(Event)} but ignoring the {@link IsClosedException} exception.
+   *
+   * @param event The event to receive
+   * @param counter The state counter upon which the event must act upon
+   */
+  default void receiveIgnoreClosed(Event event, long counter) {
```
honnix: Doesn't seem to be used.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -209,6 +225,22 @@ private void initialize(WorkflowInstance workflowInstance) {
...
   }
 
+  private void verifyCounter(Event event, long expectedCounter, RunState currentRunState) {
+    final long currentCounter = currentRunState.counter();
+    if (currentCounter > expectedCounter) {
+      final String message = "Stale event encountered. Expected counter is "
+                             + expectedCounter  + " but current counter is "
+                             + currentCounter + ". Discarding event " + event;
+      LOG.warn(message);
+      throw new StaleEventException(message);
```
honnix: I think we can either ignore this even completely or log it as `debug`. We can just throw `RuntimeException` instead of having a new exception type.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -209,6 +225,22 @@ private void initialize(WorkflowInstance workflowInstance) {
...
+    } else if (currentCounter < expectedCounter) {
```
honnix: Do we really expect this to happen?

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
fabriziodemaria: Preliminary stress tests don't show issues with this additional lookup step in the StateManager

styx-scheduler-service/src/main/java/com/spotify/styx/state/StateManager.java
fabriziodemaria: This will be gradually used when using explicit counters with StateManager

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
fabriziodemaria: As we talked irl, we can keep the `StaleEventException` but I changed the log level to `debug`

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
fabriziodemaria: This shouldn't happen, but I think it's good to cover all possibilities and log accordingly

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
honnix: Disregard this comment.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
honnix: Then let’s at least have a comment like “this should never happen” so that later on when we read the code we won’t spend too much brain cycle trying to understand why.

honnix:
we can merge #398 already to this branch, I think.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -347,16 +349,17 @@ private boolean shouldExecute(RunState runState) {
     return !deadline.isAfter(now);
   }
 
-  private void sendDequeue(InstanceState instanceState, Set<String> resourceIds) {
-    final WorkflowInstance workflowInstance = instanceState.workflowInstance();
-    final RunState state = instanceState.runState();
+  private void sendDequeue(InstanceState instance, Set<String> resourceIds) {
```
honnix: Now let's ask the question here. Why renaming? `instanceState` feels more precise.

honnix:
👍

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -127,14 +128,27 @@ public QueuedStateManager(
 
   @Override
   public CompletableFuture<Void> receive(Event event) throws IsClosedException {
+    ensureRunning();
+    // Read state counter at enqueueing time
+    final Optional<RunState> currentRunState = getActiveState(event.workflowInstance());
```
danielnorberg: This makes all callers of `receive` block on reading the active state from datastore. I worry that this causes problems like e.g. the k8s pod watcher not being able to keep up with the rate of events as it is single threaded.

honnix: This is partially addressed by https://github.com/spotify/styx/pull/431 where we send in counter explicitly.

PR #396, opened 2018-02-16
WIP: Handle limit updates gracefully

NatashaL:
A suggestion on handling updates of resource limits gracefully.


styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -217,11 +251,65 @@ private CounterSnapshot refreshCounterSnapshot(String counterId) {
...
+    final Entity limitEntity = transaction.get(limitKey);
+    final Long oldLimit =
+        limitEntity == null ? DEFAULT_MAX_COUNTER_LIMIT : limitEntity.getLong(PROPERTY_LIMIT);
+    transaction.put(Entity.newBuilder(limitKey).set(PROPERTY_LIMIT, newLimit).build());
+    redistributeShardsIfNeeded(counterId, newLimit, oldLimit);
+  }
+
+  /**
+   * Redistribution of shard values for the given counter and limits.
+   * Redistribution should be done before the limit update happens.
```
ulzha: "Should" or "must"? What are the requirements and guarantees this implementation is about to offer?

I'm seeing what seems to be not a 100% guarantee here - the redistribution is not in the same transaction as the limit update, so there is a race condition - any shard can become updated between those operations, by another instance, so the edge case is not fully averted.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -217,11 +251,65 @@ private CounterSnapshot refreshCounterSnapshot(String counterId) {
...
+   */
+  private void redistributeShardsIfNeeded(String counterId, long newLimit, long oldLimit) {
+    // get fresh snapshot that will include the new limit
+    final CounterSnapshot snapshot = refreshCounterSnapshot(counterId);
+
+    if (newLimit < oldLimit && newLimit >= getCounter(counterId)) {
+      // iterate through all shards and run single (1/-1) unit swaps between shards with excess usage
```
ulzha: +1/-1 could take a very long time...

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -250,20 +338,32 @@ public static long getLimit(Datastore datastore, String counterId) {
...
+      if (delta < 0) {
+        int shardIndex = snapshot.pickShardWithExcessUsage(delta);
+        updateCounterShard(transaction, snapshot, shardIndex, delta);
+        return;
+      }
+      throw new CounterCapacityException("Counter %s has no more capacity.", counterId);
+    }
+
     int shardIndex = snapshot.pickShardWithSpareCapacity(delta);
+    updateCounterShard(transaction, snapshot, shardIndex, delta);
```
ulzha: Could refactor this method, these lines are a repeat.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -250,20 +338,32 @@ public static long getLimit(Datastore datastore, String counterId) {
    */
   public void updateCounter(DatastoreReaderWriter transaction, String counterId, long delta) {
     CounterSnapshot snapshot = getCounterSnapshot(counterId);
+    if (getCounter(counterId) >= snapshot.limit) {
+      // when counter usage is at/over its limit, only allow negative deltas
+      if (delta < 0) {
+        int shardIndex = snapshot.pickShardWithExcessUsage(delta);
+        updateCounterShard(transaction, snapshot, shardIndex, delta);
```
ulzha: `shardIndex` can be -1, we'll throw the bug exception.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
NatashaL: Not sure how it should be phrased best. Ideally it should be called only when the `limit` is updated (which here is the case at L268 https://github.com/spotify/styx/pull/396/files#diff-33665282cd4186e101219fbbe4c1faddR268). A `limit` is a completely separate entity, so the transactions doing the redistribution of the counter shards should not conflict with the transaction provided for the `limit` update.

That being said, there *could* be issues if the counter is in use/being updated while the redistribution is ongoing, but only if the counter surpasses the new `limit` that's being set. I can see that case causing an ongoing loop until the usage decreases below the new `limit`, but that's a case that can be addressed and fixed.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
NatashaL: If you're referring to L351-352: only L352 repeats on L345, but the shardIndex is determined by two different methods, and I find this repetition to be more readable.

ulzha:
Closing the PR to free up transactional-sharded-counters branch for deletion.

PR #397, opened 2018-02-19
Log type of state that times out

NatashaL:
We've observed timeouts being applied potentially too late on a workflow state - when the state has already transitioned forward, so we see the timeout out of order in the events sequence. This PR logs which state the timeout has been applied to. We may in the future want to increase the timeout limit for some states.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -179,8 +179,7 @@ public class Scheduler {
             .collect(toCollection(Lists::newArrayList));
     Collections.shuffle(eligibleInstances);
 
-    timedOutInstances.stream().map(wfi -> InstanceState.create(wfi, activeStatesMap.get(wfi)))
-        .collect(toList()).forEach(this::sendTimeout);
+    timedOutInstances.forEach(wfi -> this.sendTimeout(wfi, activeStatesMap.get(wfi)));
 
     limitAndDequeue(config, resources, workflowResourceReferences, currentResourceUsage,
         eligibleInstances);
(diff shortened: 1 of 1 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -178,7 +179,8 @@ void tick() {
             .collect(toCollection(Lists::newArrayList));
     Collections.shuffle(eligibleInstances);
 
-    timedOutInstances.forEach(this::sendTimeout);
+    timedOutInstances.stream().map(wfi -> InstanceState.create(wfi, activeStatesMap.get(wfi)))
```
fabriziodemaria: How about 
`timedOutInstances.forEach(wfi -> sendTimeout(wfi, activeStatesMap.get(wfi))`

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -378,9 +380,10 @@ private boolean hasTimedOut(RunState runState) {
     return !deadline.isAfter(now);
   }
 
-  private void sendTimeout(WorkflowInstance workflowInstance) {
-    LOG.info("Found stale state, issuing timeout for {}", workflowInstance);
-    stateManager.receiveIgnoreClosed(Event.timeout(workflowInstance));
+  private void sendTimeout(InstanceState instanceState) {
+    LOG.info("Found stale {} state, issuing timeout for {}",
```
fabriziodemaria: Can we print something more here? Maybe not the whole `runState` but at least the registered `runState().timestamp()` could be useful for debugging

fabriziodemaria:
Looks good to be merged when tests are green

PR #402, opened 2018-02-21
backfill in transaction

honnix:
@fabriziodemaria we need to think about it.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -339,6 +339,16 @@ class DatastoreStorage {
     return queryActiveStates(query);
   }
 
+  public Map<WorkflowInstance, PersistentWorkflowInstanceState> activeStatesByTriggerId(
+      String triggerId) throws IOException {
+    final EntityQuery query =
+        Query.newEntityQueryBuilder().setKind(KIND_ACTIVE_WORKFLOW_INSTANCE)
+            .setFilter(PropertyFilter.eq(PROPERTY_STATE_TRIGGER_ID, triggerId))
+            .build();
+
+    return queryActiveStates(query);
+  }
+
   private Map<WorkflowInstance, PersistentWorkflowInstanceState> queryActiveStates(EntityQuery activeStatesQuery)
       throws IOException {
     final ImmutableMap.Builder<WorkflowInstance, PersistentWorkflowInstanceState> mapBuilder = ImmutableMap.builder();
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -724,27 +724,8 @@ class DatastoreStorage {
   }
 
   void storeBackfill(Backfill backfill) throws IOException {
-    storeWithRetries(() -> runInTransaction(tx -> tx.storeBackfill(backfill)));
+    storeWithRetries(() -> runInTransaction(tx -> tx.store(backfill)));
   }
-
-  static Entity backfillToEntity(Key key, Backfill backfill) {
-    Entity.Builder builder = Entity.newBuilder(key)
-        .set(PROPERTY_CONCURRENCY, backfill.concurrency())
-        .set(PROPERTY_START, instantToTimestamp(backfill.start()))
-        .set(PROPERTY_END, instantToTimestamp(backfill.end()))
-        .set(PROPERTY_COMPONENT, backfill.workflowId().componentId())
-        .set(PROPERTY_WORKFLOW, backfill.workflowId().id())
-        .set(PROPERTY_SCHEDULE, backfill.schedule().toString())
-        .set(PROPERTY_NEXT_TRIGGER, instantToTimestamp(backfill.nextTrigger()))
-        .set(PROPERTY_ALL_TRIGGERED, backfill.allTriggered())
-        .set(PROPERTY_HALTED, backfill.halted());
-
-    backfill.description().ifPresent(x -> builder.set(PROPERTY_DESCRIPTION, StringValue
-        .newBuilder(x).setExcludeFromIndexes(true).build()));
-
-    return builder.build();
-  }
-
 
   private <T> Stream<T> readStream(Entity entity, String property) {
     return read(entity, property, Collections.<Value<T>>emptyList()).stream()
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -43,7 +43,6 @@ import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
 import com.google.cloud.datastore.Transaction;
 import com.google.cloud.datastore.Value;
 import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
@@ -621,11 +620,11 @@ class DatastoreStorage {
   List<Resource> getResources() {
     final EntityQuery query = Query.newEntityQueryBuilder().setKind(KIND_RESOURCE).build();
     final QueryResults<Entity> results = datastore.run(query);
-    final ImmutableList.Builder<Resource> resources = ImmutableList.builder();
+    final List<Resource> resources = Lists.newArrayList();
     while (results.hasNext()) {
       resources.add(entityToResource(results.next()));
     }
-    return resources.build();
+    return resources;
   }
 
   private Resource entityToResource(Entity entity) {
@@ -676,9 +675,9 @@ class DatastoreStorage {
 
   private List<Backfill> backfillsForQuery(Query<Entity> query) {
     final QueryResults<Entity> results = datastore.run(query);
-    final ImmutableList.Builder<Backfill> resources = ImmutableList.builder();
-    results.forEachRemaining(entity -> resources.add(entityToBackfill(entity)));
-    return resources.build();
+    final List<Backfill> backfills = Lists.newArrayList();
+    results.forEachRemaining(entity -> backfills.add(entityToBackfill(entity)));
+    return backfills;
   }
 
   List<Backfill> getBackfills(boolean showAll) {
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -575,6 +575,10 @@ class DatastoreStorage {
   static Key componentKey(KeyFactory keyFactory, String componentId) {
     return keyFactory.setKind(KIND_COMPONENT).newKey(componentId);
   }
+  
+  static Key backfillKey(KeyFactory keyFactory, String backfillId) {
+    return keyFactory.setKind(KIND_BACKFILL).newKey(backfillId);
+  }
 
   static Key globalConfigKey(KeyFactory keyFactory) {
     return keyFactory.setKind(KIND_STYX_CONFIG).newKey(KEY_GLOBAL_CONFIG);
@@ -720,12 +724,10 @@ class DatastoreStorage {
   }
 
   void storeBackfill(Backfill backfill) throws IOException {
-    storeWithRetries(() -> datastore.put(backfillToEntity(backfill)));
+    storeWithRetries(() -> runInTransaction(tx -> tx.storeBackfill(backfill)));
   }
 
-  private Entity backfillToEntity(Backfill backfill) {
-    final Key key = datastore.newKeyFactory().setKind(KIND_BACKFILL).newKey(backfill.id());
-
+  static Entity backfillToEntity(Key key, Backfill backfill) {
     Entity.Builder builder = Entity.newBuilder(key)
         .set(PROPERTY_CONCURRENCY, backfill.concurrency())
         .set(PROPERTY_START, instantToTimestamp(backfill.start()))
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -711,7 +711,7 @@ class DatastoreStorage {
     return backfillsForQuery(query);
   }
 
-  private Backfill entityToBackfill(Entity entity) {
+  static Backfill entityToBackfill(Entity entity) {
     final WorkflowId workflowId = WorkflowId.create(entity.getString(PROPERTY_COMPONENT),
                                                     entity.getString(PROPERTY_WORKFLOW));
 
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
@@ -21,11 +21,22 @@
 package com.spotify.styx.storage;
 
 import static com.spotify.styx.serialization.Json.OBJECT_MAPPER;
+import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_ALL_TRIGGERED;
+import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_COMPONENT;
+import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_CONCURRENCY;
+import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_DESCRIPTION;
+import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_END;
+import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_HALTED;
 import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_NEXT_NATURAL_OFFSET_TRIGGER;
 import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_NEXT_NATURAL_TRIGGER;
+import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_NEXT_TRIGGER;
+import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_SCHEDULE;
+import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_START;
+import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_WORKFLOW;
 import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_WORKFLOW_ENABLED;
 import static com.spotify.styx.storage.DatastoreStorage.PROPERTY_WORKFLOW_JSON;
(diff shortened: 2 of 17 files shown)


honnix:
I guess this is reviewable now.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -110,53 +124,76 @@ private void triggerBackfill(Backfill backfill, Map<String, Long> backfillStates
...
+        LOG.debug("transaction failure when working on backfill {}", backfill, e);
+        break;
+      }
+    }
+  }
+
+  private Boolean tryTriggerAndProgress(StorageTransaction tx,
+                                        String id, Workflow workflow) {
+    final Backfill momentBackfill =
+        storage.backfill(id).orElseThrow(RuntimeException::new);
```
fabriziodemaria: One remote issue could be that instance A gets very slow right after this line, with read `nextTrigger` = 1, while a faster instance B proceeds with the backfill bringing `nextTrigger` to, for example, 5. WFI 1 in the meantime finishes with SUCCESS. Afterwards, instance A resumes and triggers WFI 1 _again_ and set `nextTrigger` to 2, causing the whole backfill to be re-triggered.

Perhaps `storage.backfill(id)` should be in the same transaction as `tx.store(<Backfill>)` that appears later?
If that is the case we would still trigger partition 1 twice, but at least we wouldn't regress `nextCounter` to 2.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -110,53 +124,76 @@ private void triggerBackfill(Backfill backfill, Map<String, Long> backfillStates
...
+
+    final Instant partition = momentBackfill.nextTrigger();
     final int remainingCapacity =
-        backfill.concurrency() - backfillStates.getOrDefault(backfill.id(), 0L).intValue();
+        momentBackfill.concurrency() - stateManager.activeStates(id).size();
 
-    Instant partition = backfill.nextTrigger();
+    if (remainingCapacity < 1) {
+      LOG.debug("no capacity left for backfill {}, stop triggering", momentBackfill);
```
fabriziodemaria: Should we store again the backfill with `.allTriggered(true)` just in case something went wrong when storing that after the last trigger in a previous call to `tryTriggerAndProgress`?

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
honnix: It may trigger twice in that extreme case but wouldn’t be able to progress next trigger because transaction would fail that.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -339,6 +339,16 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
     return queryActiveStates(query);
   }
 
+  public Map<WorkflowInstance, PersistentWorkflowInstanceState> activeStatesByTriggerId(
+      String triggerId) throws IOException {
+    final EntityQuery query =
+        Query.newEntityQueryBuilder().setKind(KIND_ACTIVE_WORKFLOW_INSTANCE)
+            .setFilter(PropertyFilter.eq(PROPERTY_STATE_TRIGGER_ID, triggerId))
```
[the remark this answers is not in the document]
honnix: I kind of hoped the query should filter it out.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
fabriziodemaria: Shouldn't this be `tx.backfill()` for the transaction to fail instead of `storage.backfill()`?

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
honnix: Maybe to be paranoid. :)

The last storing (next trigger and that flag) all happens within a transaction. Not sure how that would happen.

styx-scheduler-service/src/main/java/com/spotify/styx/state/StateManager.java
```diff
@@ -57,6 +57,11 @@
    */
   Map<WorkflowInstance, RunState> activeStates();
 
+  /**
+   * Get a map of all active {@link WorkflowInstance} states filtered by triggerId.
+   */
+  Map<WorkflowInstance, RunState> activeStates(String triggerId);
```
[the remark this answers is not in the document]
honnix: Yeah. Makes sense. Let’s change it.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -110,53 +124,76 @@ private void triggerBackfill(Backfill backfill, Map<String, Long> backfillStates
...
+
+  private Boolean tryTriggerAndProgress(StorageTransaction tx,
+                                        String id, Workflow workflow) {
+    final Backfill momentBackfill =
+        storage.backfill(id).orElseThrow(RuntimeException::new);
+
+    final Instant partition = momentBackfill.nextTrigger();
     final int remainingCapacity =
-        backfill.concurrency() - backfillStates.getOrDefault(backfill.id(), 0L).intValue();
+        momentBackfill.concurrency() - stateManager.activeStates(id).size();
```
fabriziodemaria: I am a bit concerned about this best effort solution.
1 - Could be confusing for users to set limit 2 and see 3 instances being triggered
2 - Sometimes concurrency 1 is used to ensure strict sequential advancement of the backfill, which would now not be guaranteed anymore.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -91,17 +104,18 @@ void tick() {
...
+    shuffler.accept(backfills);
+
+    backfills.forEach(backfill -> guard(() -> triggerBackfill(backfill)).run());
 
     final long durationMillis = t0.until(time.get(), ChronoUnit.MILLIS);
     stats.recordTickDuration(TICK_TYPE, durationMillis);
   }
 
-  private void triggerBackfill(Backfill backfill, Map<String, Long> backfillStates) {
+  private void triggerBackfill(Backfill backfill) {
```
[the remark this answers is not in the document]
honnix: triggerAndProgressBackfill?

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
honnix: The slow node wouldn’t be able to store a stale backfill if it has been updated by the other node.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
honnix: I thought about it and it is actually not possible, otherwise it means two nodes can work on the same backfill concurrently and they both succeed without any conflict.

fabriziodemaria: I think it can also happen if the query from `activeStates` uses old indexes and doesn't see all the triggered backfill.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
fabriziodemaria: > The last storing (next trigger and that flag) all happens within a transaction. Not sure how that would happen.

Good point, that's enough

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
honnix: That won’t matter I think. Even we calculate wron here, it won’t happen node A triggers partition 1 and node B triggers partition 2 concurrently. Or?

fabriziodemaria: From my tests on staging this seems to happen with even just one instance, i.e. one more WFI is triggered over the concurrency limit (it happens consistently). I assume that after the last expected trigger we call `tryTriggerAndProgress` one more time and we don't see the previously triggered WFI yet from `stateManager.activeStates(id)`

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
honnix: As we figured out we need to read the backfill in the beginning of the transaction.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
honnix: It turned out datastore query is eventually consistent, so this solution doesn’t work. It’s probably better just keeping the counter in memory. Unless one node is super slow after computing the remaining capacity, race will be very rare.

styx-common/src/main/java/com/spotify/styx/util/TimeUtil.java
```diff
@@ -103,6 +103,34 @@ public static Instant nextInstant(Instant instant, Schedule schedule) {
...
+  public static int numberOfInstants(Instant lastInstant, Instant firstInstant, Schedule schedule) {
+    if (lastInstant.isBefore(firstInstant)) {
+      throw new IllegalArgumentException();
+    }
+
+    final ExecutionTime executionTime = ExecutionTime.forCron(cron(schedule));
+
+    Instant currentInstant = lastInstant;
+    int number = 0;
+    while (!currentInstant.equals(firstInstant)) {
```
fabriziodemaria: A bit concerned about receiving instants that are not aligned with schedule (this might loop forever). Perhaps `currentInstant.isAfter(firstInstant)`?

honnix: This is just a lib, the caller should take the responsibility in this case I think.

fabriziodemaria: I agree, but even if we push the alignment responsibility to the caller I think we should at least catch misalignment and throw an exception instead of getting stuck in an infinite loop.

honnix: OK, if we really want to. To me it is over complicating things. I can make a fix.

honnix: They are now enforced to be aligned https://github.com/spotify/styx/pull/402/files#diff-83fdfa5c37381e562bd0705651c031a6R115

fabriziodemaria:
Went back to returning just a boolean from the inner nested layer to indicate if to break the loop, in the attempt to simplify the logic and readability a bit; the proper check of the `numberOfInstants` is inside the inner nested layer anyway.

To me this looks good to be merged.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
```diff
@@ -175,4 +188,36 @@ public WorkflowInstance deleteActiveState(WorkflowInstance instance) {
...
+  }
+
+  @Override
+  public Optional<Backfill> backfill(String id) {
+    final Key key = DatastoreStorage.backfillKey(tx.getDatastore().newKeyFactory(), id);
+    final Entity entity = tx.get(key);
+    if (entity == null) {
+      return Optional.empty();
+    }
+    return Optional.of(entityToBackfill(entity));
```
ulzha: Minor cleanup for later: why backfillToEntity was ported to here but entityToBackfill wasn't?

fabriziodemaria: Good point, we'll need a proper refactoring for these Datastore classes in future PR.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -91,72 +104,110 @@ void tick() {
...
-        LOG.warn("failed to trigger backfill for state [{}]: {}", partition, backfill, e);
-        return;
+        if (!storage.runInTransaction(tx ->
+            triggerNextPartitionAndProgress(tx, backfill.id(), workflow,
+                initialNextTrigger, remainingCapacity))) {
+          break;
+        }
+      } catch (IOException e) {
+        // transaction failure, yield
+        LOG.debug("Transaction failure when working on backfill {}", backfill, e);
```
ulzha: IOException can be something else than transaction failure; isn't this a potentially misleading message?

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -91,72 +104,110 @@ void tick() {
...
+                                                  String id,
+                                                  Workflow workflow,
+                                                  Instant initialNextTrigger,
+                                                  int remainingCapacity) {
+    final Backfill momentBackfill = tx.backfill(id).orElseThrow(RuntimeException::new);
+    final Instant momentNextTrigger = momentBackfill.nextTrigger();
+
+    if (numberOfInstants(momentNextTrigger, initialNextTrigger,
+        momentBackfill.schedule()) >= remainingCapacity) {
+      LOG.debug("Capacity reached for backfill {}", momentBackfill);
```
ulzha: The `lastInstant` and `firstInstant` were both inclusive, so the comparison doesn't exactly make sense. If `remainingCapacity` is 1 and `momentNextTrigger` equals `initialNextTrigger`, then I expect this WFI to be triggered, but if `numberOfInstants()` behaves as documented then we are comparing `1 >= 1` and saying no to it...?

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -91,72 +104,110 @@ void tick() {
...
-        .map(Optional::get)
-        .filter(TriggerUtil::isBackfill)
-        .collect(groupingBy(
-            TriggerUtil::triggerId,
-            HashMap::new,
-            counting()));
+    try {
+      // race conditions:
+      //   - the other scheduler instance has triggered partition, but hasn't committed the
+      //     transaction, and the workflow instance has finished very quickly
```
ulzha: Hm, how can this happen? The instance can't get submitted to GKE before transactional state transition goes through commit, can it?

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -91,72 +104,110 @@ void tick() {
...
+      //   - this scheduler reads initialNextTrigger before the other scheduler commits the
+      //     transaction, so it will move on to trigger the same partition again, but we still
+      //     don't violate concurrency limit in this case
+      final CompletableFuture<Void> processed = triggerListener
+          .event(workflow, Trigger.backfill(momentBackfill.id()), momentNextTrigger)
+          .toCompletableFuture();
+      // Wait for the trigger execution to complete before proceeding to the next partition
+      processed.get();
```
ulzha: It seems one could just `.event().wait()` instead of going `toCompletableFuture()` in between, but I guess this was preexisting code...

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -91,72 +104,110 @@ void tick() {
...
-      partition = nextInstant(partition, backfill.schedule());
-      storeBackfill(backfill.builder()
-                        .nextTrigger(partition)
-                        .build());
+  private boolean triggerNextPartitionAndProgress(StorageTransaction tx,
+                                                  String id,
+                                                  Workflow workflow,
+                                                  Instant initialNextTrigger,
+                                                  int remainingCapacity) {
+    final Backfill momentBackfill = tx.backfill(id).orElseThrow(RuntimeException::new);
```
ulzha: A bit odd to not include a message...

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
ulzha: Shouldn't we have a timeout? `get()` blocks indefinitely...

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
fabriziodemaria: The actual triggering (state transition) happens synchronously but on a separated transaction with respect to the one created in this class (the latter handling the backfill entity only). This implementation is indeed a bit different from the one we discussed irl.

ulzha: Or maybe you mean the other scheduler instance has triggered the partition and transitioned through many states, in fact up until finishing it and becoming inactive (each of those is a committed transaction) and then this instance triggers it again...

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
fabriziodemaria: I believe we'd have this problem in many other places in the code. I suggest to open a new PR about this in the future if we want to tackle this.

ulzha:
LGTM, let's rock.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
fabriziodemaria: > Or maybe you mean the other scheduler instance has triggered the partition and transitioned through many states, in fact up until finishing it and becoming inactive (each of those is a committed transaction) and then this instance triggers it again...

Precisely. Very remote case nevertheless, and we would still not pollute `nextTrigger` anyway because that part of the transaction would fail.

fabriziodemaria:
Rebased, ready to be merged

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -91,72 +104,111 @@ void tick() {
...
-      storeBackfill(backfill.builder()
-                        .nextTrigger(partition)
-                        .build());
+  private boolean triggerNextPartitionAndProgress(StorageTransaction tx,
+                                                  String id,
+                                                  Workflow workflow,
+                                                  Instant initialNextTrigger,
+                                                  int remainingCapacity) {
+    final Backfill momentBackfill = tx.backfill(id).orElseThrow(() ->
+        new RuntimeException("Error while fetching backfill " + id));
```
ulzha: If it comes back empty then it's probably a fact that the backfill was removed, not that there was an error fetching it.

PR #403, opened 2018-02-22
 Add resourceIds in StateData to propagate and persist them

ulzha:
Needed for upcoming use of ShardedCounters (https://github.com/spotify/styx/pull/378) - so that we can consistently decrement the right resources (that were incremented upon dequeue), when the execution finishes.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
@@ -38,7 +38,7 @@ public interface EventVisitor<R> {
 
   R triggerExecution(@Getter WorkflowInstance workflowInstance, Trigger trigger);
   R info(@Getter WorkflowInstance workflowInstance, Message message);
-  R dequeue(@Getter WorkflowInstance workflowInstance, @Nullable Set<String> resourceRefs);
+  R dequeue(@Getter WorkflowInstance workflowInstance, @Nullable Set<String> resourceIds);
   R submit(@Getter WorkflowInstance workflowInstance, ExecutionDescription executionDescription,
       @Nullable String executionId);
   R submitted(@Getter WorkflowInstance workflowInstance, @Nullable String executionId);
--- a/styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
@@ -38,7 +38,7 @@ public interface EventVisitor<R> {
 
   R triggerExecution(@Getter WorkflowInstance workflowInstance, Trigger trigger);
   R info(@Getter WorkflowInstance workflowInstance, Message message);
-  R dequeue(@Getter WorkflowInstance workflowInstance, @Nullable Set<String> resourceIds);
+  R dequeue(@Getter WorkflowInstance workflowInstance, Set<String> resourceIds);
   R submit(@Getter WorkflowInstance workflowInstance, ExecutionDescription executionDescription,
       @Nullable String executionId);
   R submitted(@Getter WorkflowInstance workflowInstance, @Nullable String executionId);
--- a/styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
@@ -26,6 +26,7 @@ import com.github.sviperll.adt4j.Visitor;
 import com.spotify.styx.state.Message;
 import com.spotify.styx.state.Trigger;
 import java.util.Optional;
+import java.util.Set;
 import javax.annotation.Nullable;
 
 /**
@@ -37,7 +38,7 @@ public interface EventVisitor<R> {
 
   R triggerExecution(@Getter WorkflowInstance workflowInstance, Trigger trigger);
   R info(@Getter WorkflowInstance workflowInstance, Message message);
-  R dequeue(@Getter WorkflowInstance workflowInstance);
+  R dequeue(@Getter WorkflowInstance workflowInstance, Set<String> resourceRefs);
   R submit(@Getter WorkflowInstance workflowInstance, ExecutionDescription executionDescription,
       @Nullable String executionId);
   R submitted(@Getter WorkflowInstance workflowInstance, @Nullable String executionId);
--- a/styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
@@ -38,7 +38,7 @@ public interface EventVisitor<R> {
 
   R triggerExecution(@Getter WorkflowInstance workflowInstance, Trigger trigger);
   R info(@Getter WorkflowInstance workflowInstance, Message message);
-  R dequeue(@Getter WorkflowInstance workflowInstance, Set<String> resourceRefs);
+  R dequeue(@Getter WorkflowInstance workflowInstance, @Nullable Set<String> resourceRefs);
   R submit(@Getter WorkflowInstance workflowInstance, ExecutionDescription executionDescription,
       @Nullable String executionId);
   R submitted(@Getter WorkflowInstance workflowInstance, @Nullable String executionId);
--- a/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
@@ -104,9 +104,9 @@ class WFIExecutionBuilder {
     }
 
     @Override
-    public Void dequeue(WorkflowInstance workflowInstance, Set<String> resourceRefs) {
+    public Void dequeue(WorkflowInstance workflowInstance, Set<String> resourceIds) {
       currWorkflowInstance = workflowInstance;
-      // TODO use resourceRefs somehow?
+      // TODO use resourceIds somehow?
       return null;
     }
 
--- a/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
@@ -31,6 +31,7 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import javax.annotation.Nullable;
 
 class WFIExecutionBuilder {
@@ -103,8 +104,9 @@ class WFIExecutionBuilder {
     }
 
     @Override
-    public Void dequeue(WorkflowInstance workflowInstance) {
+    public Void dequeue(WorkflowInstance workflowInstance, Set<String> resourceRefs) {
       currWorkflowInstance = workflowInstance;
+      // TODO use resourceRefs somehow?
       return null;
     }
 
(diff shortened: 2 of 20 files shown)


styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
```diff
@@ -37,7 +38,7 @@
 
   R triggerExecution(@Getter WorkflowInstance workflowInstance, Trigger trigger);
   R info(@Getter WorkflowInstance workflowInstance, Message message);
-  R dequeue(@Getter WorkflowInstance workflowInstance);
+  R dequeue(@Getter WorkflowInstance workflowInstance, @Nullable Set<String> resourceIds);
```
honnix: Is it possible not to mark it as `@Nullable`, empty set should be good enough to indicate nothing.

styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
```diff
@@ -103,8 +104,9 @@ public Void info(WorkflowInstance workflowInstance, Message message) {
     }
 
     @Override
-    public Void dequeue(WorkflowInstance workflowInstance) {
+    public Void dequeue(WorkflowInstance workflowInstance, Set<String> resourceIds) {
       currWorkflowInstance = workflowInstance;
+      // TODO use resourceIds somehow?
```
honnix: Raw event is not really used by UI, so fine I think.

styx-common/src/main/java/com/spotify/styx/serialization/PersistentEvent.java
```diff
@@ -89,8 +90,8 @@ public PersistentEvent created(WorkflowInstance workflowInstance, String executi
     }
 
     @Override
-    public PersistentEvent dequeue(WorkflowInstance workflowInstance) {
-      return new PersistentEvent("dequeue", workflowInstance.toKey());
+    public PersistentEvent dequeue(WorkflowInstance workflowInstance, Set<String> resourceIds) {
```
honnix: We might need to make `resourceIds` as `Optional<Set<String>>`. Backfill still replays events.

honnix: We did exactly the same thing before, so something we can refer to https://github.com/spotify/styx/commit/6c3ce31a269f6e0f555875776a05eebebeee3ada#diff-caa8e641522b9a838652711092851507R92

styx-common/src/test/java/com/spotify/styx/state/RunStateTest.java
```diff
@@ -95,7 +97,7 @@ public void testTimeTriggerAndRetry() throws Exception {
   public void testTimeTriggerAndRetry2() throws Exception {
     transitioner.initialize(RunState.fresh(WORKFLOW_INSTANCE));
     transitioner.receive(eventFactory.triggerExecution(UNKNOWN_TRIGGER));
-    transitioner.receive(eventFactory.dequeue());
+    transitioner.receive(eventFactory.dequeue(null));
```
honnix: empty set instead?

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
```diff
@@ -427,7 +428,7 @@ public void shouldWriteActiveStateOnEvent() throws Exception {
         .data(StateData.zero())
         .build()));
 
-    stateManager.receive(Event.dequeue(INSTANCE))
+    stateManager.receive(Event.dequeue(INSTANCE, null))
```
honnix: empty set instead?

styx-common/src/main/java/com/spotify/styx/state/StateData.java
```diff
@@ -46,6 +47,7 @@
   Optional<String> triggerId(); //for backwards compatibility
   Optional<String> executionId();
   Optional<ExecutionDescription> executionDescription();
+  Optional<Set<String>> resourceIds(); // resources referenced in the workflow configuration at the time of dequeue
```
honnix: Don't we need to persist this in `DatastoreStorage`?

NatashaL: Yes, because we need it for two actions `increment counters` for the given resources whenever a wfi is dequeued, and `decrement counters` whenever it is terminated or failed. Those actions may be done by two different scheduler instances, and we need the resourceIds for both.

NatashaL: Yes, addressed in additional commit

styx-common/src/main/java/com/spotify/styx/serialization/PersistentEvent.java
ulzha: In a bunch of places `null` was intentionally used to test backwards compatibility. (Deserialization when resource_ids is absent.)  I think the latest changes have made it actually different from the handling in other cases... Confused.

styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
```diff
@@ -37,7 +38,7 @@
 
   R triggerExecution(@Getter WorkflowInstance workflowInstance, Trigger trigger);
   R info(@Getter WorkflowInstance workflowInstance, Message message);
-  R dequeue(@Getter WorkflowInstance workflowInstance);
+  R dequeue(@Getter WorkflowInstance workflowInstance, Set<String> resourceIds);
```
ulzha: This was `@Nullable` to handle backwards compatibility consistently, like in the `submitted` event below, which has `@Nullable String executionId`.

honnix: I think it is usually a good practice using empty collection instead of null to indicate no value. But this case is different.

styx-common/src/main/java/com/spotify/styx/serialization/PersistentEvent.java
```diff
@@ -322,6 +322,25 @@ public Event toEvent() {
...
   }
 
+  public static class Dequeue extends PersistentEvent {
+
+    public final Set<String> resourceIds;
+
+    @JsonCreator
+    public Dequeue(
+        @JsonProperty("workflow_instance") String workflowInstance,
+        @JsonProperty("resource_ids") Optional<Set<String>> resourceIds) {
```
ulzha: This was also consistent with `Submit` handling below but b5777c0 made it inconsistent...

styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
ulzha: Yeah I'd like to be able to use empty collection (no `null`s and no `Optionals` too); it would be a great move to make but let's make it consistently, for all events that currently have backwards compatibility concerns and resolve them with @Nullable fields for distinguishing "no field even existed" from "no value".

styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
ulzha: Done this way to be consistent with `submit` below, that similarly evolved a new field `executionId`.

styx-common/src/test/java/com/spotify/styx/state/RunStateTest.java
ulzha: I intended to leave this test as close to unchanged as possible, so that it tests the same bits of functionality that it tested before (i.e. cover functionality of legacy events, which don't record the set at all). The new tests added below use sets, never `null`.

ulzha:
@fabriziodemaria PTAL

styx-common/src/main/java/com/spotify/styx/state/RunState.java
```diff
@@ -182,13 +183,14 @@ public RunState info(WorkflowInstance workflowInstance, Message message) {
...
     @Override
-    public RunState dequeue(WorkflowInstance workflowInstance) {
+    public RunState dequeue(WorkflowInstance workflowInstance, Set<String> resourceIds) {
       switch (state()) {
         case QUEUED:
           return state(
               PREPARE,
               data().builder()
                   .retryDelayMillis(empty())
+                  .resourceIds(resourceIds)
```
fabriziodemaria: `resourceIds` can be null here, but I believe the builder translates it into a Optional.empty. I'd still prefer to explicitly receive an Optional.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -387,6 +388,7 @@ static PersistentWorkflowInstanceState readPersistentWorkflowInstanceState(Entit
           .executionId(readOpt(entity, PROPERTY_STATE_EXECUTION_ID))
           .executionDescription(readOptJson(entity, PROPERTY_STATE_EXECUTION_DESCRIPTION,
               ExecutionDescription.class))
+          .resourceIds(readOpt(entity, PROPERTY_STATE_RESOURCE_IDS))
```
fabriziodemaria: This is for reading from Datastore but writing to Datastore is missing (i.e. `activeStateToEntity`)
EDIT: `testFullPersistentStatePersistence` would have caught the issue, also update `FULL_PERSISTENT_STATE` in tests.

styx-common/src/test/java/com/spotify/styx/state/RunStateTest.java
fabriziodemaria: I think we should keep this test class up to date with transition logic, and limit backwards compatible tests to `ReplayEventsTest` class. Current transition logic passes empty sets from `Scheduler.java`, not null. Note that by using Optional to begin with in `EventVisitor` we would get rid of `null` everywhere.

styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
```diff
@@ -581,7 +582,7 @@ public void shouldLimitOnDecoratedWorkflowInstanceResourcesIfNotAvailable() thro
     verify(stateManager).receiveIgnoreClosed(Event.info(INSTANCE_1,
         Message.info("Resource limit reached for: [Resource{id=baz, concurrency=0}]")));
 
-    verify(stateManager, never()).receiveIgnoreClosed(Event.dequeue(INSTANCE_1));
+    verify(stateManager, never()).receiveIgnoreClosed(Event.dequeue(INSTANCE_1, ImmutableSet.of("foo", "bar", "baz", "GLOBAL_STYX_CLUSTER")));
```
fabriziodemaria: Long line :)

styx-scheduler-service/src/test/java/com/spotify/styx/state/QueuedStateManagerTest.java
fabriziodemaria: I agree, `Scheduler.java` doesn't use `null` but empty set.

styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
ulzha: Oh, yeah, I didn't notice there was already two different prior approaches :/

Intellij actually highlights the `Optional<Integer>` used as type for parameter: "Optional was designed to provide a limited mechanism for library method _return types_"...

styx-common/src/main/java/com/spotify/styx/state/RunState.java
ulzha: Indeed translated by builder same way as in `submit`; will leave it consistent with that for this PR.

styx-common/src/main/java/com/spotify/styx/model/EventVisitor.java
ulzha: Apologies for going back and forth; actually I agree with making it non-null already (equating "no resources were used" with "we don't know what resources were used") because it's ok that the scheduler behaves the same, for all purposes AFAICT...

PR #404, opened 2018-02-23
Shardedcounter refactoring

NatashaL:
Refactor `ShardedCounter` to depend on `Storage` and `StorageTransaction` instead of `Datastore`.

As per the current setup of how Styx communicates with Datastore, there are two places where read/write logic happens, in `DatastoreStorage` used by `AggregateStorage implements Storage` and `DatastoreStorageTransaction implements StorageTransaction`, so it was a bit confusing when deciding where to put the logic needed for `ShardedCounter`.  My approach was to put as much as I could in `DatastoreStorageTransaction` (whatever was required to be done in a transaction) and the rest in `DatastoreStorage`.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
@@ -20,11 +20,6 @@
 
 package com.spotify.styx.util;
 
-import com.google.cloud.datastore.Datastore;
-import com.google.cloud.datastore.Entity;
-import com.google.cloud.datastore.EntityQuery;
-import com.google.cloud.datastore.QueryResults;
-import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
@@ -39,7 +34,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Random;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.IntStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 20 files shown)


styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -296,35 +258,18 @@ public long getCounter(String counterId) {
...
-          if (results.hasNext()) {
-            transaction.delete(results.next().getKey());
-          }
-        });
-        return null;
-      });
+  public void deleteCounter(Storage storage, String counterId) {
+    storage.deleteShardsForCounter(counterId);
+    try {
+      storage.deleteLimitForCounter(counterId);
```
ulzha: Curious why the deletions are differently handled, one has a catch and the other doesn't?

NatashaL: Hm, yeah, it does look strange a bit. I think my reasoning for it was that if we fail when deleting some shards we should short-circuit, forward any exception to the caller and keep the limit if it exists, so that it still applies on the remaining shards. And if deleting the shards is successful then try to delete the limit as well.

However, reading the comment now, we do explicitly say that the whole operation is best effort, so maybe we should catch all exceptions after all. What do you think?

ulzha: Discrepancy also in terms of storeWithRetries, and in terms of where the code is located (deleteShardsForCounter is implemented in DatastoreStorage, yet deleteLimitForCounter forwards to an implementation in DatastoreStorageTransaction).

NatashaL: Yes, the reason for that is that deleting of a limit can be done in a single transaction, whereas deleting the shards usually takes more than one given the constraint that Datastore transactions have of modifying max 25 entries at once.

Additionally, at the time being, there's no clear contract of what belongs where when it comes to operations against datastore and the two locations of implementations that we have which you mentioned (DatastoreStorage and DatastoreStorageTransaction)

ulzha:
:+1: and I guess we should make all the methods in ShardedCounter API have consistent exception signatures... Later...

PR #406, opened 2018-02-27
Mark completed backfills as allTriggered

fabriziodemaria:
It is possible (maybe it shouldn't) to create backfills that have start=end, for which `allTriggered` will never be reached. Complements #402 

@honnix @ulzha PTAL


honnix:
Isn’t end exclusive?

honnix:
Anyway it doesn’t hurt.

fabriziodemaria:
> Isn’t end exclusive?

Yes, but at backfill's creation time we still allow start=end. We could prevent that possibility to begin with, but as you said it doesn't hurt to be a bit more robust here as well.

honnix:
+1

PR #408, opened 2018-03-12
Revert "Transactional sharded counters"

NatashaL:
Reverts spotify/styx#378

We've experienced issues after merging this PR. State transitions are much slower, which causes events to queue faster than they can be processed.


honnix:
👍

PR #409, opened 2018-03-12
best effort when hitting rate limit

honnix:
Also change the log level to info to ease metrics collection.

@fabriziodemaria PTAL


fabriziodemaria:
LGTM

PR #411, opened 2018-03-12
Sort by time when scheduling

honnix:
@fabriziodemaria PTAL

This prevents starvation when dequeuing workflow instances.


fabriziodemaria:
Looks good to me

PR #412, opened 2018-03-13
handling cases when trigger execution event is slow

honnix:
and timeout is issued

@fabriziodemaria We will end up with weird executions though.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
@@ -42,7 +42,7 @@ class WFIExecutionBuilder {
 
   @Nullable private WorkflowInstance currWorkflowInstance;
   @Nullable private String currExecutionId;
-  @Nullable private String currTriggerId;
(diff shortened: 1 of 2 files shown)


styx-common/src/main/java/com/spotify/styx/model/data/WFIExecutionBuilder.java
```diff
@@ -221,6 +220,12 @@ public Void timeout(WorkflowInstance workflowInstance) {
 
       executionStatusList.add(ExecStatus.create(eventTs, "TIMEOUT", Optional.empty()));
 
+      // we might get timeout before triggerExecution, and in that case we take best effort to
+      // set trigger timestamp
+      if (triggerTs == null) {
```
fabriziodemaria: How about moving this check at the beginning of the method and also return if this if condition is true (after setting `triggerTS`)? So we avoid showing weird executions

honnix: Good point! Let's try.

fabriziodemaria:
LGTM!

PR #414, opened 2018-03-14
Optimize shard initialization

NatashaL:


styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -74,48 +76,79 @@
...
+    private static final int TRANSACTION_GROUP_SIZE = 15;
+    private final String counterId;
+    private final Long limit;
+    private final Map<Integer, Long> shards;
+
+    public CounterSnapshot(Storage storage, String counterId,
+                           Map<Integer, Long> shards) {
+      limit = getLimit(storage, counterId);
+
+      this.shards = storage.shardsForCounter(counterId);
```
fabriziodemaria: This is now already fetched from the Factory code, right?

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
```diff
@@ -91,8 +94,10 @@ public void tearDown() throws IOException, InterruptedException {
 
   @Test
   public void shouldCreateCounterEmpty() {
+    long before = System.currentTimeMillis();
     assertEquals(shardedCounter.getCounter(COUNTER_ID1), 0L);
-
+    long after = System.currentTimeMillis();
+    System.out.println(after - before);
```
fabriziodemaria: To be removed I guess

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -74,48 +76,79 @@
...
+        });
+      } catch (IOException e) {
+        LOG.warn("Error when trying to create a group of shards in Datastore: ", e);
       }
     }
+  }
+
+  private static class CounterSnapshot {
+
+    private static final int TRANSACTION_GROUP_SIZE = 15;
```
fabriziodemaria: Unused here. Where does 15 come from by the way?

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -74,48 +76,79 @@
...
      */
-    private static void initialize(Storage storage, String counterId) {
-      for (int i = 0; i < NUM_SHARDS; i++) {
-        final int shardIndex = i;
-        try {
-          storage.runInTransaction(tx -> {
-            final Optional<Shard> shard = tx.shard(counterId, shardIndex);
+    private static void initialize(Storage storage, String counterId,
+                                   ExecutorService executorService) {
+      for (int i = 0; i < NUM_SHARDS; i += TRANSACTION_GROUP_SIZE) {
```
fabriziodemaria: This double `for` loop seems buggy, with 128 shards and group size of 15 we end up creating 135 shards. This should be tested as well.

NatashaL: You're right, there was one more condition to avoid this, which I've removed by mistake

NatashaL:
@fabriziodemaria that is the idea implemented here - parallelized batch writes. `initShard` (has a typo, should be `initShards`) does multiple writes in the same transaction. But maybe I've missed some nuances when it comes to the performance of multiple writes within a transaction. I'll look into it

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
NatashaL: Ran a few tests and came to the conclusion that grouping 15 writes per transaction balances the overhead between thread creation and transaction creation/size

NatashaL: This may change if we change how batch is done

NatashaL:
@fabriziodemaria it seems that the current implementation achieves what you mentioned. Calling `.put(entity)` multiple times within a transaction is the same as calling `.put(entities...)` once for a set of entities. The confusing double loops achieve running parallel threads while batching shards in fewer transactions.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
NatashaL: I've revised this, and see that I've done something wrong in the initial tests. the group/batch size doesn't seem to make a difference, and neither does parallelizing the transactions (due to the thread overhead and awaiting for thread shutdown). 

Simple batching has the highest impact.
That said, obviously, we do need to parallelize when we initialize shards on startup.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -126,23 +97,28 @@ private static void initialize(Storage storage, String counterId) {
...
      */
-    private int pickShardWithSpareCapacity(long delta) {
+    public int pickShardWithSpareCapacity(long delta) {
       List<Integer> candidates = new ArrayList<>();
 
-      for (int i : shards.keySet()) {
-        if (shards.containsKey(i) && Range
-            .closed(0L, shardCapacity(i)).contains(shards.get(i) + delta)) {
-          candidates.add(i);
+      shards.keySet().forEach(index -> {
```
honnix: This seems to be achievable using `filter`.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -246,8 +223,8 @@ public long getCounter(String counterId) {
     final CounterSnapshot snapshot = getCounterSnapshot(counterId);
     long result = 0;
 
-    for (long shard : snapshot.shards.values()) {
-      result += shard;
+    for (long shardValue : snapshot.getShards().values()) {
```
honnix: Let's do it like `snapshot.getShards().values().stream().mapToLong(i -> i).sum()`

styx-common/src/main/java/com/spotify/styx/util/ShardedCounterSnapshotFactory.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+public class ShardedCounterSnapshotFactory implements CounterSnapshotFactory {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ShardedCounterSnapshotFactory.class);
+
+  public static final int TRANSACTION_GROUP_SIZE = 25;
+  private final Storage storage;
+
+  public ShardedCounterSnapshotFactory(Storage storage) {
+    this.storage = storage;
```
honnix: `Null` check.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -74,47 +74,18 @@
...
-        // The counter probably has not been initialized (so we have empty QueryResults). Also
-        // possible that a prior initialize() crashed halfway, or we got a partial list of shards in
-        // QueryResults due to eventual consistency. In any case, repeated initialization eventually
-        // creates all NUM_SHARDS shards.
-        initialize(storage, counterId);
-        fetchedShards = storage.shardsForCounter(counterId);
-      }
+    public Snapshot(Storage storage, String counterId, Map<Integer, Long> shards) {
+      this.limit = getLimit(storage, counterId);
+      this.shards = shards;
```
honnix: `Null` check

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -161,8 +137,9 @@ private int pickShardWithSpareCapacity(long delta) {
     }
   }
 
-  public ShardedCounter(Storage storage) {
+  public ShardedCounter(Storage storage, CounterSnapshotFactory counterSnapshotFactory) {
     this.storage = storage;
+    this.counterSnapshotFactory = counterSnapshotFactory;
```
honnix: `Null` check, blame Java. :D

NatashaL:
@honnix @fabriziodemaria PTAL. Note that this will be merged into the `transactional-sharded-counters2` branch and NOT into master. We can, of course, make further changes on that branch that concern the sharded counter. This branch was for isolating the changes around shard initialization, and has grown a bit.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -126,23 +103,24 @@ private static void initialize(Storage storage, String counterId) {
...
-
-      for (int i : shards.keySet()) {
-        if (shards.containsKey(i) && Range
-            .closed(0L, shardCapacity(i)).contains(shards.get(i) + delta)) {
-          candidates.add(i);
-        }
-      }
+    public int pickShardWithSpareCapacity(long delta) {
+      List<Integer> candidates =
+          shards.keySet().stream().filter(index -> shards.containsKey(index) && Range
```
honnix: The code is indented in a strange way now. Try to break the line from `.filter` already? Also I wouldn't mind having two filters to make it easier to read instead of using `&&``.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -438,6 +444,34 @@ private void warmUpCache(WorkflowCache cache, Storage storage) {
...
+      storage.resources().parallelStream().forEach(
+          r -> initializeResourceShards_(counterSnapshotFactory, storage, shardedCounter, r));
+      LOG.info("Finished initializing resources");
+    } catch (IOException e) {
+      LOG.error("Error reading resources from datastore: {}", e);
+      // TODO: re-throw exception when moving to entirely depend on counter shards
+    }
+  }
+
+  private void initializeResourceShards_(CounterSnapshotFactory counterSnapshotFactory,
```
honnix: I don't think you need `_` here. It doesn't feel like very Java.

honnix:
Some minor comments.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -74,47 +74,24 @@
       .expireAfterWrite(CACHE_EXPIRY_DURATION.toMillis(), TimeUnit.MILLISECONDS)
       .build();
 
-  private static class CounterSnapshot {
+  public ShardedCounter(Storage storage, CounterSnapshotFactory counterSnapshotFactory) {
+    this.storage = Objects.requireNonNull(storage);
```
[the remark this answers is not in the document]
NatashaL: It is with the last commit e65b832

PR #415, opened 2018-03-14
flow the execution

honnix:
Reading active states consumes a big part of scheduler tick, so little
time is used for actual execution. By keeping the loop flowing we can
get better throughput without having a big burst.

@fabriziodemaria PTAL


fabriziodemaria:
LGTM

PR #416, opened 2018-03-16
[Hold] shuffle workflow instances again

honnix:
This is more or less a revert of #411. Because of #415, starvation
wouldn't happen anymore. Shuffle the workflow instances to make
multi-scheduler setup happier.

@fabriziodemaria PTAL 

According to https://github.com/spotify/styx/pull/411#issuecomment-373725143


honnix:
On-hold until we move on with multi-scheduler.

honnix:
@danielnorberg PTAL

danielnorberg:
👍

ulzha:
What do we expect to be the effects and how to measure if we were correct?

The total number of failed transactions is perhaps unchanged regardless of randomization, because sendDequeue is called unconditionally. I guess decrease in tick duration (duration halved given two schedulers) could be a somewhat intuitive improvement to expect, but will really be?

honnix:
Good point. It's hard to tell whether this optimisation actually works.

honnix:
I would close this for now until we have evidence to show it is absolutely needed.

PR #418, opened 2018-03-21
Revert "Transactional sharded counters"

NatashaL:
Reverts spotify/styx#413.

We've observed issues of failing to transition between states due to being unable to decrement sharded counters beyond 0 (zero). 

Either,
1) All decrements should be allowed, before deploying a version that syncs current resource usage with the shards' state
2) Deploy a version that syncs current resource usage with the shards' state and keep the current decrement logic


honnix:
+1

Let’s open it later and one comment would be log more in ShardCounter.

honnix:
We can work on master instead of reverting, but that would leave master undeployable now. Maybe OK this time.

PR #419, opened 2018-03-22
clean up and swallow counter update exception

honnix:
@NatashaL @ulzha @fabriziodemaria Start from this and close #408 ?


NatashaL:
👍

PR #422, opened 2018-03-25
datastore operation to delete a component

honnix:
~~preparing for workflow house cleaning.~~
Get ready to move away from keyed component.


honnix:
This is not important if we go down this direction: https://github.com/spotify/styx/pull/485

So close for now.

PR #423, opened 2018-03-25
styx housecleaner

honnix:
Prepare to inject Styx housecleaner.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -56,6 +56,8 @@ import com.spotify.apollo.route.Route;
 import com.spotify.metrics.core.SemanticMetricRegistry;
 import com.spotify.styx.api.Api;
 import com.spotify.styx.api.SchedulerResource;
+import com.spotify.styx.cleaner.Cleaner;
+import com.spotify.styx.cleaner.CleanerOperation;
 import com.spotify.styx.docker.DockerRunner;
 import com.spotify.styx.model.Event;
 import com.spotify.styx.model.Resource;
@@ -163,6 +165,7 @@ public class StyxScheduler implements AppInit {
   private final EventConsumerFactory eventConsumerFactory;
   private final WorkflowConsumerFactory workflowConsumerFactory;
   private final WorkflowExecutionGateFactory executionGateFactory;
+  private final HousecleaningOperationFactory housecleaningOperationFactory;
 
   private StateManager stateManager;
   private Scheduler scheduler;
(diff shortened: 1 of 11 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/cleaner/CleanerOperation.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.styx.cleaner;
+
+public interface CleanerOperation {
```
ulzha: Is this conceptually different from the pattern of using Closeable and Closer?

ulzha: Or, I guess this is meant to get invoked repeatedly... Then why is this not just a Runnable? If there's some design thinking about what is going to be  specific in CleanerOperation's used and evolution, then documenting it seems like a good idea.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -157,6 +159,7 @@
   private final EventConsumerFactory eventConsumerFactory;
   private final WorkflowConsumerFactory workflowConsumerFactory;
   private final WorkflowExecutionGateFactory executionGateFactory;
+  private final HouseCleaningOperationFactory housecleaningOperationFactory;
```
ulzha: Nitpick: if it's HouseCleaningFoo then the lower camel case is more consistently houseCleaningFoo...

styx-scheduler-service/src/main/java/com/spotify/styx/cleaner/CleanerOperation.java
honnix: Sure we can just use `RunnableWithException` in `styx-common`.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
honnix: Oh yeah, my bad, should just call it `HousecleaningOperationFacotry`.

ulzha:
All in all looks good as an extension point. Though the contract is a bit unclear, feels like implementing a generic scheduler inside scheduler.

styx-scheduler-service/src/main/java/com/spotify/styx/cleaner/CleanerOperation.java
honnix: It might still be a good idea to have this specialised interface than the very generic `RunnableWithException`.

honnix:
@ulzha PTAL again

honnix:
In favour of simpler solution.

PR #420, opened 2018-03-22
Sync resources at startup

fabriziodemaria:
TODO: Add tests for `StateUtil.class`

PR #425, opened 2018-03-26
Better logging for resources sync

fabriziodemaria:
On top of #420


honnix:
👍  let's ignore `StyxScheduler` coverage.

PR #426, opened 2018-03-26
align consuming resource check

honnix:


ulzha:
Also kill that comment which says "We'll keep counting terminal states as if they consume resources"...

ulzha:
Also kill that comment which says "We'll keep counting terminal states as if they consume resources"...

ulzha:
...I think we might want to put back the old conservative behavior later: especially when we transition into FAILED because of a timeout; we have a race condition where we can kick off new instances before killing the one that's timing out...

styx-scheduler-service/src/test/java/com/spotify/styx/state/StateUtilTest.java
```diff
@@ -113,4 +115,22 @@ public void shouldGetUsedResourcesStorageException() throws IOException {
...
+  @Test
+  public void shouldConsumeResource() {
+    assertTrue(StateUtil.isConsumingResources(RunState.State.PREPARE));
+    assertTrue(StateUtil.isConsumingResources(RunState.State.SUBMITTING));
+    assertTrue(StateUtil.isConsumingResources(RunState.State.SUBMITTED));
+    assertTrue(StateUtil.isConsumingResources(RunState.State.RUNNING));
+    assertTrue(StateUtil.isConsumingResources(RunState.State.TERMINATED));
+    assertTrue(StateUtil.isConsumingResources(RunState.State.FAILED));
+    assertTrue(StateUtil.isConsumingResources(RunState.State.ERROR));
+    assertTrue(StateUtil.isConsumingResources(RunState.State.DONE));
```
ulzha: Wait, I think exclude ERROR and DONE... If we never transition out of them, the resources will never decrement... In the old code we decremented resources implicitly by deleting the ActiveWorkflowInstance.

We should not literally be "counting terminal states" but "counting states where the container is not yet cleaned up"...

ulzha:
:+1:

ulzha:
We IRLed a bit about the effects of not counting resources in FAILED state and we think we're not making things much worse. Since `retryAfter` induces some wait in QUEUED state, the pod cleanup will have decent time to occur, before we kick off new instances.

I think a proper fix would be to perform cleanup of the pods strictly before returning to QUEUED state.

PR #427, opened 2018-03-26
send info event about resource limit

honnix:
Send info event from `QueuedStateManager` where resource validation will be done in the future.


fabriziodemaria:
LGTM

PR #428, opened 2018-03-27
(WIP) Remove the in-memory logic for limiting resources

ulzha:
Ended up being a long PR, containing
* the switch to transactional counters (removal of in-memory logic)
* a fix to correctly reset counters when syncing
* partial fixes for flaky tests

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ * 
  *      http://www.apache.org/licenses/LICENSE-2.0
- *
+ * 
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -19,6 +19,8 @@
  */
 
 package com.spotify.styx.util;
+
+import static java.util.stream.Collectors.toList;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
@@ -34,7 +36,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Random;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -81,10 +82,9 @@ public class ShardedCounter {
     private final Long limit;
     private final Map<Integer, Long> shards;
 
-    Snapshot(Storage storage, String counterId, Map<Integer, Long> shards) {
-      Objects.requireNonNull(storage);
+    Snapshot(String counterId, long limit, Map<Integer, Long> shards) {
       this.counterId = Objects.requireNonNull(counterId);
-      this.limit = getLimit(storage, counterId);
+      this.limit = limit;
       this.shards = Objects.requireNonNull(shards);
     }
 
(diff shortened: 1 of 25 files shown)


styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerServiceFixture.java
```diff
@@ -99,8 +101,8 @@
   private List<Tuple2<Optional<Workflow>, Optional<Workflow>>> workflowChanges = Lists.newArrayList();
 
   // captured fields from fakes
-  List<Tuple2<WorkflowInstance, DockerRunner.RunSpec>> dockerRuns = Lists.newArrayList();
-  List<String> dockerCleans = Lists.newArrayList();
+  Queue<Tuple2<WorkflowInstance, DockerRunner.RunSpec>> dockerRuns = new ConcurrentLinkedQueue();
```
honnix: Awesome!!

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -245,44 +244,19 @@ private void limitAndDequeue(Map<String, Resource> resources,
...
-      }
     } else {
       double sleepingTime = dequeueRateLimiter.acquire();
       if (sleepingTime > 0.0001) {
         LOG.debug("Dequeue rate limited and slept for {} ms", sleepingTime * 1000);
       }
 
-      instanceResourceRefs.forEach(id -> currentResourceUsage.computeIfAbsent(id, id_ -> 0L));
-      instanceResourceRefs.forEach(id -> currentResourceUsage.compute(id, (id_, l) -> l + 1));
+      // Racy: some resources may have been removed (become unknown) by now; in that case the
```
fabriziodemaria: Is this really a race _issue_? Resources that are removed should after all not be used to limit executions anymore...

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -486,6 +490,23 @@ private void initializeResources(Storage storage, ShardedCounter shardedCounter,
     }
   }
 
+  @VisibleForTesting
+  void resetShards(final Storage storage, final Resource resource) {
```
fabriziodemaria: Maybe 'updateShards' and 'resetShards' should be part of 'Storage' interface. No need to refactor in this PR.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -251,17 +248,18 @@ private void updateResourceCounters(StorageTransaction tx, Event event,
...
-      // to a state that doesn't consume any resources
-      for (String resource : nextRunState.data().resourceIds().get()) {
-        tx.updateCounter(shardedCounter, resource, -1);
+        && !isConsumingResources(nextRunState.state())) {
+      if (nextRunState.data().resourceIds().isPresent()) {
+        for (String resource : nextRunState.data().resourceIds().get()) {
+          tx.updateCounter(shardedCounter, resource, -1);
+        }
+      } else {
+        LOG.error("Resource ids are missing for {} when transitioning from {} to {}.",
```
fabriziodemaria: I think it is acceptable to have RunStates with no resource id. Even if practically we should always have at least global resource and some decorators, those are not a mandatory configuration in OSS.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
ulzha: I think it's not dangerous, but UX wise it's an issue anyway I think because resources that are removed normally cause workflow to error out (it references a deleted resource in its configuration - bad), instead of silently running without limiting.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
ulzha: The logic of persisting the set is meant to be consistent, even in OSS. We aren't checking if it has elements in it, just checking the optional that it `isPresent`.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
fabriziodemaria: Oh I get what you mean now. Deleting resources doesn't happen very often, we can live with this for now.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -273,19 +271,20 @@ private void tryUpdatingCounter(RunState runState,
...
         .collect(toSet());
-    final Set<String> depletedResources = failedTries.stream()
+    final Set<String> depletedResourceIds = failedTries.stream()
         .filter(x -> x._2.getCause() instanceof CounterCapacityException)
         .map(x -> x._1)
         .sorted()
         .collect(toSet());
 
-    if (!depletedResources.isEmpty()) {
+    if (!depletedResourceIds.isEmpty()) {
```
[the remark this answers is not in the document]
honnix: We do throw the exception in this case, just a few lines down. We discussed it.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -181,7 +181,7 @@ public ShardedCounter(Storage storage, CounterSnapshotFactory counterSnapshotFac
   /**
    * Returns a recent snapshot, possibly read from inMemSnapshot.
    */
-  private CounterSnapshot getCounterSnapshot(String counterId) {
+  public CounterSnapshot getCounterSnapshot(String counterId) {
```
ulzha: Why was this made public? It seems to be used below for `getLimit` but that is already available as a public method on `ShardedCounter`.

ulzha: My bad, not public yet. Let's fix that one to be public, instead of publicizing CounterSnapshot, which is a relatively deeper implementation detail.

honnix: I remember it wasn’t that straightforward to just make getLimit public. Maybe I’m wrong. It’s been a while.

PR #429, opened 2018-03-27
Handle cases of excess usage in shards

fabriziodemaria:


styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java
```diff
@@ -101,6 +101,38 @@ public long shardCapacity(int shardIndex) {
...
+     * Returns shard index which _likely_ has excess usage, as per our cached view of the state in Datastore.
+     */
+    @Override
+    public Optional<Integer> pickShardWithExcessUsage(long delta) {
+      final List<Integer> candidates = shards.keySet().stream()
+          .filter(index -> shards.get(index) > shardCapacity(index))
+          .filter(index -> shards.get(index) + delta >= 0)
+          .collect(toList());
+
+      if (candidates.isEmpty()) {
```
honnix: we can just do a `.map()`

honnix: never mind, disregard.

honnix:
👍

PR #430, opened 2018-03-28
[hold] Job no pod

honnix:
Launch a `Job` instead of a `Pod`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -225,7 +225,7 @@
       <dependency>
         <groupId>com.squareup.okhttp3</groupId>
         <artifactId>okhttp</artifactId>
-        <version>3.9.0</version>
+        <version>3.9.1</version>
       </dependency>
 
       <!-- external clients libs -->
@@ -237,7 +237,7 @@
       <dependency>
         <groupId>io.fabric8</groupId>
         <artifactId>kubernetes-client</artifactId>
-        <version>2.2.13</version>
+        <version>3.1.8</version>
       </dependency>
       <dependency>
         <groupId>com.google.apis</groupId>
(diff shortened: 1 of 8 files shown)


honnix:
Found cases that Pods are still deleted by k8s master if they belong to Jobs. Also found cases that deleting a Job doesn't delete it's Pod. :/

honnix:
It is an expected behaviour that delete a `Job` via API will NOT delete the `Pod`. However k8s can still decide to delete a `Pod` on its own (due to high load?)

honnix:
btw, this PR is ready for review.

honnix:
According to Google it should be reliable that Pod will be kept around until explicitly delete it. 
```
When a Job completes, no more Pods are created, but the Pods are not deleted either. Since they are terminated, they don’t show up with kubectl get pods, but they will show up with kubectl get pods -a. Keeping them around allows you to still view the logs of completed pods to check for errors, warnings, or other diagnostic output. The job object also remains after it is completed so that you can view its status. It is up to the user to delete old jobs after noting their status.
```
We talked to Google about our use case and using a Job was what they recommended. 

Regarding immediate benefit, Pod garbage collection has become very aggressive in recent releases and that has caused us issues like rerunning pipelines, losing logs.

danielnorberg:
But the terminated pods _are_ aggresively deleted by the k8s gc, that's what we're seeing right? I.e., they will _not_ show up using `kubectl get pods -a` either. Afaik using jobs will not mitigate that.

honnix:
Not that massive if we use Job. Verified that in our staging environment. In approximately a week time we only observed a couple of cases.

honnix:
We pushed the staging to very high load when testing it.

danielnorberg:
Ok, when I tried using a Job on our production k8s the terminated pod quickly disappeared.

danielnorberg:
Reverified just now by running a job in prod and the pod disappeared within 2 minutes. A manually created pod also disappeared within 2 minutes. I interpret that as raw pods and job pods being gc'd the same way. Job pods are not kept around longer than raw pods.

danielnorberg:
As to why google claims differently ¯\\_(ツ)_/¯. I guess we can report it as a GKE bug.

honnix:
That’s quite different result than staging. Yeah I think we should report it as a bug or get a more official and confirmed answer.

honnix:
Pod didn't survive more than a few minutes in production. Talking to Google to confirm.

honnix:
In favour of #465

honnix:
No good reason to keep this open.

PR #432, opened 2018-04-05
get rid of workflow cache

honnix:
how to deal with workflow cache

Let's start from here. If concurrent lookups are not very heavy, we seem to be in a very good position.


Zatine:
@danielnorberg PTAL

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -679,20 +680,18 @@ private void setupMetrics(
   }
 
   private static Consumer<Workflow> workflowRemoved(
-      WorkflowCache cache,
       Storage storage,
       BiConsumer<Optional<Workflow>, Optional<Workflow>> workflowConsumer) {
-    return workflow -> cache.workflow(workflow.id()).ifPresent(existingWorkflow -> {
+    return workflow -> {
```
honnix: `SchedulerResource` already reads from Datastore before calling this consumer.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -330,7 +329,9 @@ public void create(Environment environment) {
     final Config staleStateTtlConfig = config.getConfig(STYX_STALE_STATE_TTL_CONFIG);
     final TimeoutConfig timeoutConfig = TimeoutConfig.createFromConfig(staleStateTtlConfig);
 
-    warmUpCache(workflowCache, storage);
+    final WorkflowCache workflowCache = new InMemWorkflowCache(
+        new CachedSupplier<>(storage::workflows, time));
+
     initializeResources(storage, shardedCounter, counterSnapshotFactory, timeoutConfig, workflowCache);
```
danielnorberg: If this workflow cache is only used during initialization, how about just reading the all the workflows from storage here and directly passing that in to `initializeResources` etc instead of having the `WorkflowCache` layer of indirection?

honnix: It is used as a `Gauge` for metrics and that is why we chose `CachedSupplier`.

danielnorberg: Ah, the cache os also passed into `setupMetrics` and used there in gauge lambas so just reading the workflows once doesn't work.

I guess the `WorkflowCache` could still be replaced with a `Supplier<Map<WorkflowId, Workflow>>`, but maybe it's convenient to keep the `WorkflowCache` interface?

honnix: Yeah, right. Mostly our laziness. Didn't want to change too many things where that interface is expected. :D

danielnorberg: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/state/StateUtil.java
```diff
@@ -108,11 +109,11 @@ private StateUtil() {
...
-                                                                          InstanceState instanceState,
-                                                                          Set<Workflow> workflows,
-                                                                          WorkflowResourceDecorator resourceDecorator) {
-    final Optional<Workflow> workflowOpt = workflows.stream().filter(
-        wf -> wf.id().equals(instanceState.workflowInstance().workflowId())).findFirst();
+                                                                InstanceState instanceState,
+                                                                Map<WorkflowId, Workflow> workflows,
+                                                                WorkflowResourceDecorator resourceDecorator) {
+    final Optional<Workflow> workflowOpt =
+        Optional.ofNullable(workflows.get(instanceState.workflowInstance().workflowId()));
```
danielnorberg: You beat me to it! 😂

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -113,7 +111,14 @@ void tick() {
   }
 
   private void triggerAndProgress(Backfill backfill) {
-    final Optional<Workflow> workflowOpt = workflowCache.workflow(backfill.workflowId());
+    final Optional<Workflow> workflowOpt;
+    try {
+      workflowOpt = storage.workflow(backfill.workflowId());
+    } catch (IOException e) {
+      LOG.warn("Failed to read workflow {}", backfill.workflowId(), e);
```
danielnorberg: Before this change, the workflow cache would throw a RTE that would bubble up to the caller of `triggerAndProgress` but now the caller will not get an exception. Does that matter?

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -648,27 +645,35 @@ private void setupMetrics(
...
+      }
+
       stats.registerActiveStatesMetric(
           workflow.id(), workflowActiveStates(stateManager, workflow));
 
-      final Optional<Workflow> oldWorkflowOptional = cache.workflow(workflow.id());
-
-      workflowInitializer.inspectChange(workflow);
-      cache.store(workflow);
+      workflowInitializer.inspectChange(oldWorkflowOptional, workflow);
```
danielnorberg: Looking up the old workflow and then mutating it seems racy to me. Use a transaction?

styx-scheduler-service/src/main/java/com/spotify/styx/workflow/WorkflowInitializer.java
```diff
@@ -45,15 +45,7 @@ public WorkflowInitializer(Storage storage, Time time) {
...
-  public void inspectChange(Workflow workflow) {
-    final Optional<Workflow> previous;
-    try {
-      previous = storage.workflow(workflow.id());
-    } catch (IOException e) {
-      LOG.warn("failed to read workflow {} from storage", workflow.id(), e);
-      throw new RuntimeException(e);
-    }
-
+  public void inspectChange(Optional<Workflow> previous, Workflow workflow) {
```
danielnorberg: The lookup and mutation in this method should be in a transaction.

styx-scheduler-service/src/test/java/com/spotify/styx/InMemWorkflowCacheTest.java
```diff
@@ -0,0 +1,71 @@
...
+  public void setUp() {
+    when(supplier.get()).thenReturn(ImmutableMap.of(WORKFLOW_ID, WORKFLOW_WITH_RESOURCES,
+        WORKFLOW_ID_2, WORKFLOW_WITH_RESOURCES_2));
+    workflowCache = new InMemWorkflowCache(supplier);
+  }
+
+  @Test
+  public void shouldGetAllWorkflows() {
+    final Map<WorkflowId, Workflow> workflows = workflowCache.all();
+    assertThat(workflowCache.all().size(), is(2));
```
danielnorberg: Prefer an equality check against the expected map.

``` 
assertThat(workflowCache.all(), is(ImmutableMap.of(
  WORKFLOW_ID, WORKFLOW_WITH_RESOURCES,
   WORKFLOW_ID_2, WORKFLOW_WITH_RESOURCES_2)));
```

Even better, refactor the map in `setUp` into a constant that can be used in this method.

styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowCache.java
```diff
@@ -57,5 +43,5 @@
    *
    * @return an unordered set containing all Workflows
    */
-  ImmutableSet<Workflow> all();
```
danielnorberg: Consider removing the above `workflow(WorkflowId)` method as lookup can now be done using `all().get()`.

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerTest.java
```diff
@@ -172,6 +190,79 @@ public void shouldFailToResetShardsOfResource() throws IOException {
...
+  @Test
+  public void shouldHandleWorkflowChanged() throws IOException {
+    final Workflow workflow = Workflow.create(WORKFLOW_ID.componentId(),
+        HOURLY_WORKFLOW_CONFIGURATION_WITH_VALID_OFFSET);
+    when(storage.workflow(workflow.id())).thenReturn(Optional.of(workflow));
+    final Consumer<Workflow> consumer = StyxScheduler.workflowChanged(workflowInitializer, stats,
+        stateManager, storage, workflowConsumer);
+    consumer.accept(WORKFLOW_WITH_RESOURCES);
+    //noinspection unchecked
+    verify(stats).registerActiveStatesMetric(eq(WORKFLOW_WITH_RESOURCES.id()), any(Gauge.class));
```
danielnorberg: Use `any()` instead of `any(Gauge.class)` and remove the `//noinspection unchecked`

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerTest.java
```diff
@@ -172,6 +190,79 @@ public void shouldFailToResetShardsOfResource() throws IOException {
...
   }
 
+  @Test
+  public void shouldHandleWorkflowUnchanged() throws IOException {
+    when(storage.workflow(WORKFLOW_WITH_RESOURCES.id()))
+        .thenReturn(Optional.of(WORKFLOW_WITH_RESOURCES));
+    final Consumer<Workflow> consumer = StyxScheduler.workflowChanged(workflowInitializer, stats,
+        stateManager, storage, workflowConsumer);
+    consumer.accept(WORKFLOW_WITH_RESOURCES);
+    //noinspection unchecked
```
danielnorberg: Same as below

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerTest.java
```diff
@@ -172,6 +190,79 @@ public void shouldFailToResetShardsOfResource() throws IOException {
...
+    when(storage.workflow(workflow.id())).thenThrow(exception);
+    final Consumer<Workflow> consumer = StyxScheduler.workflowChanged(workflowInitializer, stats,
+        stateManager, storage, workflowConsumer);
+    try {
+      consumer.accept(WORKFLOW_WITH_RESOURCES);
+      fail();
+    } catch (Exception e) {
+      assertThat(e.getCause(), is(exception));
+    }
+    //noinspection unchecked
```
danielnorberg: Same as above

styx-scheduler-service/src/test/java/com/spotify/styx/api/SchedulerResourceTest.java
```diff
@@ -305,6 +307,18 @@ public void testCreateWorkflowInvalidConfiguration() throws Exception {
...
+  @Test
+  public void testCreateWorkflowFailedToInitialize() throws Exception {
+    doThrow(new WorkflowInitializationException(new Exception()))
+        .when(workflowChangeListener).accept(FULL_DAILY_WORKFLOW);
+    final Response<ByteString> r = serviceHelper.request(
+        "POST", BASE + "/workflows/styx",
+        serialize(FULL_DAILY_WORKFLOW.configuration())).toCompletableFuture().get();
+    verify(workflowValidator).validateWorkflowConfiguration(FULL_DAILY_WORKFLOW.configuration());
+    verify(workflowChangeListener).accept(FULL_DAILY_WORKFLOW);
+    assertThat(r, hasStatus(withCode(Status.BAD_REQUEST)));
```
danielnorberg: Is `WorkflowInitializationException` always a bad request?

honnix: I would say so. It happens when we cannot parse the `offset`, `schedule`.

styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowCache.java
honnix: Removed.

styx-scheduler-service/src/main/java/com/spotify/styx/workflow/WorkflowInitializer.java
honnix: I restored the code to make room for adding transaction which I prefer to do it in a separated PR.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -182,12 +184,30 @@ private void updateResourceStats(Map<String, Resource> resources,
         .forEach(r -> stats.recordResourceUsed(r, 0));
   }
 
+  private Map<WorkflowId, Workflow> getWorkflows(final List<InstanceState> activeStates) {
+    return activeStates.stream()
+        .map(activeState -> activeState.workflowInstance().workflowId())
+        .collect(toSet())
+        .parallelStream()
```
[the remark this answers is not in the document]
honnix: Doing it sequentially for multiple batches. Seems fine performance. Maybe we would like to parallelise that as well?

styx-scheduler-service/src/main/java/com/spotify/styx/WorkflowCache.java
honnix: Gone gone, like everything.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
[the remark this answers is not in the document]
honnix: Turned it into parallel execution using customized FJP to avoid degrading overall JVM performance potentially.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -154,20 +162,24 @@
...
   }
 
+  @VisibleForTesting
   DatastoreStorage(Datastore datastore, Duration retryBaseDelay,
-      Function<Transaction, DatastoreStorageTransaction> storageTransactionFactory) {
+                   Function<Transaction, DatastoreStorageTransaction> storageTransactionFactory) {
     this.datastore = Objects.requireNonNull(datastore);
     this.retryBaseDelay = Objects.requireNonNull(retryBaseDelay);
     this.storageTransactionFactory = Objects.requireNonNull(storageTransactionFactory);
+    this.forkJoinPool = new ForkJoinPool();
```
honnix: Didn't bother injecting which doesn't bring much value in this case. FJP is quite deterministic.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -308,6 +320,39 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
     return map;
   }
 
+  public Map<WorkflowId, Workflow> workflows(Set<WorkflowId> workflowIds) {
+    final Callable<Map<WorkflowId, Workflow>> callable =
```
danielnorberg: I'm a little uneasy with this as FJP and java streams are designed for parallel _computation_, not blocking IO. 

https://stackoverflow.com/questions/27584271/stream-processing-thread-pool-exhaustion#comment43608638_27584271

http://gee.cs.oswego.edu/dl/html/StreamParallelGuidance.html

I'm not sure how FJP will behave wrt blocked worker threads unless the `ManagedBlocker` is used.

https://zeroturnaround.com/rebellabs/fixedthreadpool-cachedthreadpool-or-forkjoinpool-picking-correct-java-executors-for-background-tasks/

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: It seems undesirable to have the datastore IO parallelism be derived from the number of cpu cores on the host.

Specify an explicit level of desired parallelism.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -308,6 +320,39 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
...
+            .partition(ImmutableList.copyOf(workflowIds), MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH)
+            .parallelStream()
+            .map(this::getBatchOfWorkflows)
+            .flatMap(Collection::stream)
+            .collect(toMap(Map.Entry::getKey, Map.Entry::getValue));
+    return forkJoinPool.submit(callable).join();
+  }
+
+  private Set<Map.Entry<WorkflowId, Workflow>> getBatchOfWorkflows(final List<WorkflowId> batch) {
+    final Set<Key> keys = batch.stream()
```
danielnorberg: There's a lot of seemingly unnecessary work here in creating maps / sets and transforming them to sequential streams and then repackaging them as maps again.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
honnix: Yeah I read those articles. It seems the concern is about using the common pool with blocking IO. And that for sure will impact overall JVM performance. ManagedBlocker gives a hint to create more workers to make enlarge common pool if I understand it correctly.

danielnorberg: Did some simple experiments (on my 4 core laptop) with managed vs unmanaged blocking of dedicated FJP workers and the results surprised me.

When using parallel stream, the number of threads used would never go above 16, even when configuring the FJP concurrency to a high value like 1000. This remained true even when using `managedBlock`. However, doing many raw `submit()` calls produced the expected level of concurrency.

Setting the FJP concurrency to a number lower than 16 capped the effective number of worker threads used to that number.

Turns out that java parallel streams limit the concurrency, independently of the underlying FJP. 

https://stackoverflow.com/a/37518272

danielnorberg: https://stackoverflow.com/a/29272776

danielnorberg: I've verified that the parallel stream concurrency behavior changes in JDK 10. In JDK 10 the full dedicated FJP concurrency is utilized.

honnix: I have started to question the actual benefits getting batches in parallel vs. complexity using a thread pool, future, wait and then merge. We only do 3-4 batches normally. Maybe something we can use in CompletableFurtures?

honnix: Btw even with the buggy FJP, the performance reading workflows was improved around 30%. 4s vs. 6s

danielnorberg:
@honnix attempted to address my above concerns wrt FJP etc in a PR for this PR =) #440

honnix:
@danielnorberg Is this good enough for a +1?

honnix:
I think next step could be completely removing the proxy that will not make any sense after this.

danielnorberg:
👍

PR #431, opened 2018-03-28
increase the counter

honnix:
@NatashaL PTAL

PR #433, opened 2018-04-08
by default do not check counter

honnix:
I tend to make default behaviour not to check counter. Those places where we require counter check should call the other method explicitly.
@fabriziodemaria PTAL and let's discuss it.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
@@ -26,7 +26,6 @@ import static java.util.stream.Collectors.toSet;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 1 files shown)


honnix:
Or we fix it like #431 in commit https://github.com/spotify/styx/pull/431/commits/e95341a84095446190651e2fbb0629193d0426b6

danielnorberg:
I think this change still makes sense. Ignoring e.g. an incoming `halt` due to concurrent event processing seems undesirable, which seems like it might happen as things are right now.

honnix:
There could be cases that it is so slow that a termination event for previous execution gets handled for current execution. But this is rather remote case.

honnix:
@danielnorberg I tend to get this merged and shipped.

danielnorberg:
I think we should handle case of termination events from the previous execution by filtering on the current execution id _in the state machine transition transaction_. E.g. instead of just injecting an event into the QSM, inject a lambda that gets executed in the transaction and that can abort if the pod event is stale.

danielnorberg:
👍

PR #434, opened 2018-04-09
test jdk 10

danielnorberg:


honnix:
It seems jdk10 is still not LTS version. ;)

danielnorberg:
Bumped jacoco to a version that supports jdk 10.

Now it seems to fail because our bigtable deps use a version of hbase that parses the `java.version` property and assumes that it contains a `.`. But for the GA JDK 10 release `java.version` is just `10`.

danielnorberg:
@honnix PTAL

danielnorberg:
This seems to also resolve https://github.com/spotify/styx/pull/457

honnix:
Another comment, do we want to keep jdk9? To me it is very unlike we would go 9 instead of 10, or later 11. Or the idea is to make sure we are aways backward+forward compatible. The only concern is it might take longer for CI to finish.

danielnorberg:
Java 9, the version nobody ever used... The idea was to ensure that styx works on all supported jdk versions, but I'd be fine with dropping 9.

PR #436, opened 2018-04-10
restore counter increment on re-trigger

danielnorberg:


danielnorberg:
https://github.com/spotify/styx/pull/385/commits/ed96463cdfe2d4a1e955aba955bf766d5db3bf90#r180320854

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -161,13 +161,14 @@ public QueuedStateManager(
   private void initialize(WorkflowInstance workflowInstance) {
     // Write active state to datastore
 
-    final long counter;
+    final long nextCounter;
     try {
-      counter = storage.getLatestStoredCounter(workflowInstance).orElse(NO_EVENTS_PROCESSED);
+      final long counter = storage.getLatestStoredCounter(workflowInstance).orElse(NO_EVENTS_PROCESSED);
+      nextCounter = counter + 1;
```
fabriziodemaria: This also means that the NEW RunState will have counter 0 instead of -1 in case of the first trigger, right? I don't see any problem with that (it probably makes more sense even)

bergman: yeah, don't block on my nitpick :)

danielnorberg:
I'm no longer sure that reverting to the old behavior makes any sense. 

Seems that the old behavior was for the QSM to keep the _next_ counter value in memory and increment _after_ processing and storing a new event on receive.

(O_o)

danielnorberg:
I now realize that increment the counter value on trigger/initialize would result in a gap in the event counter values on re-trigger. That seems undesirable.

danielnorberg:
As for the `NEW` being stored with a `-1` counter instead of `0` on initial trigger, I'm not sure I care. It only ever exists for a few milliseconds before transitioning to `QUEUED`. Afaict actually observing an active state in `NEW` means that something is broken. Additionally, `NEW` is the only state that was reached without a corresponding event. As such I wonder if `NEW` should exist at all. 

I propose eliminating the `-1` counter oddity and the possibility of the `NEW` -> `QUEUED` transition breaking by simply always writing `QUEUED` directly.

PR #438, opened 2018-04-10
Read(lookup) shards by key

honnix:
To get strongly consistent shard usage.

### DIFF
--- a/styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
+++ b/styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
@@ -738,12 +738,13 @@ public class DatastoreStorageTest {
   @Test
   public void shouldReturnShardsForCounter() throws Exception {
     storage.runInTransaction(tx -> {
+      tx.store(Shard.create(COUNTER_ID1, 0, 0));
       tx.store(Shard.create(COUNTER_ID1, 1, 3));
       return null;
     });
     final Map<Integer, Long> map = storage.shardsForCounter(COUNTER_ID1);
-    assertEquals(1, map.size());
-    assertTrue(map.containsKey(1));
+    assertEquals(2, map.size());
+    assertEquals(0, map.get(0).longValue());
     assertEquals(3, map.get(1).longValue());
   }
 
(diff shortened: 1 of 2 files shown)


styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -732,6 +735,18 @@ public void runInTransactionShouldThrowIfDatastoreNewTransactionFails() throws E
     verify(transactionFunction, never()).apply(any());
   }
 
+  @Test
+  public void shouldReturnShardsForCounter() throws Exception {
+    storage.runInTransaction(tx -> {
+      tx.store(Shard.create(COUNTER_ID1, 1, 3));
```
honnix: Let's at least create two? ;)

honnix:
👍

ulzha:
Note it's not guaranteed to be a strongly consistent snapshot of the entire list of shards: AFAIU each lookup by key in the batch `get` is still an independent lookup unless they're placed in a transaction.

Anyway this is necessary for avoiding the unpredictably delayed consistency of global queries...

styx-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -734,6 +735,19 @@ public void runInTransactionShouldThrowIfDatastoreNewTransactionFails() throws E
...
 
+  @Test
+  public void shouldReturnShardsForCounter() throws Exception {
+    storage.runInTransaction(tx -> {
+      tx.store(Shard.create(COUNTER_ID1, 0, 0));
+      tx.store(Shard.create(COUNTER_ID1, 1, 3));
+      return null;
+    });
+    final Map<Integer, Long> map = storage.shardsForCounter(COUNTER_ID1);
+    assertEquals(2, map.size());
```
ulzha: Minor nitpick: all the other tests use `assertThat` and `is`, for consistent and good readability of messages.

honnix: That has been a question mark in my head for long time. What is the benefit of using assertThat?

PR #439, opened 2018-04-10
to report invalid execution exception

honnix:
This kind of exception we would prefer not to get alert. With a specific type we can then filter out.

@danielnorberg PTAL


honnix:
`InvalidExecutionException` is caused by user fault, such as max keys, missing permission, etc.

PR #440, opened 2018-04-11
use FJP directly instead of parallel stream

danielnorberg:
parallel stream pre-jdk 10 does not respect the concurrency of the underlying FJP

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -30,7 +30,6 @@ import static com.spotify.styx.util.ShardedCounter.PROPERTY_SHARD_INDEX;
 import static com.spotify.styx.util.ShardedCounter.PROPERTY_SHARD_VALUE;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
-import static java.util.stream.Collectors.toSet;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
@@ -53,8 +52,8 @@ import com.google.cloud.datastore.Transaction;
 import com.google.cloud.datastore.Value;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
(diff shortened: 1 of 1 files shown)


styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -321,36 +321,28 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
...
-
-  private Set<Map.Entry<WorkflowId, Workflow>> getBatchOfWorkflows(final List<WorkflowId> batch) {
-    final Set<Key> keys = batch.stream()
-        .map(workflowId -> workflowKey(datastore.newKeyFactory(), workflowId))
-        .collect(toSet());
-    final Iterator<Entity> result = datastore.get(keys);
+    final Iterable<List<WorkflowId>> batches = Iterables.partition(workflowIds, MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH);
+    return StreamSupport.stream(batches.spliterator(), false)
+        .map(batch -> forkJoinPool.submit(() -> this.getBatchOfWorkflows(batch)))
+        .collect(toList())
```
honnix: Is this needed?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -321,36 +321,28 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
...
-            .collect(toMap(Map.Entry::getKey, Map.Entry::getValue));
-    return forkJoinPool.submit(callable).join();
-  }
-
-  private Set<Map.Entry<WorkflowId, Workflow>> getBatchOfWorkflows(final List<WorkflowId> batch) {
-    final Set<Key> keys = batch.stream()
-        .map(workflowId -> workflowKey(datastore.newKeyFactory(), workflowId))
-        .collect(toSet());
-    final Iterator<Entity> result = datastore.get(keys);
+    final Iterable<List<WorkflowId>> batches = Iterables.partition(workflowIds, MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH);
```
honnix: This method is what I tried to find. :)

honnix:
Merged it.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
honnix: Doesn't seem to be needed. At least to make it compile.

honnix: So much difference in performance! Without `collect` and `stream` it is much slower.

honnix: OK I get it. Java `collect` is magic.

danielnorberg: Without `collect` it will run just one task at a time =)

This behavior is certainly not obvious, should add a comment about the need for collect + stream.

honnix: I will add it.

PR #441, opened 2018-04-11
Remove active-states-per-workflow-count metric

NatashaL:
The metric is quite expensive but unused, and not that useful.


honnix:
Looks good but let's wait for https://github.com/spotify/styx/pull/432

honnix:
Rebased from master. Ready for final review.

honnix:
@danielnorberg PTAL

danielnorberg:
👍

PR #442, opened 2018-04-11
disable known flaky tests

honnix:
As we discussed, we will temporarily identify and disable them, and work on a branch (https://github.com/spotify/styx/tree/deflaky-tests) to get them fixed.


honnix:
Not sure of anything more though. Basically running CircleCI multiple times.

honnix:
I'm kind of relieved that disabling system test doesn't change our coverage.

danielnorberg:
Maybe we shouldn't use the datastore emulator. It behaves quite differently from the real thing.

👍

honnix:
I guess the idea was to get as real as possible for system test. However it has turned out to be quite opposite. ☹️

PR #443, opened 2018-04-11
(WIP) remove proxy for workflow operation

honnix:
@danielnorberg PTAL

This is on top of https://github.com/spotify/styx/pull/432

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
@@ -20,27 +20,33 @@
 
 package com.spotify.styx.api;
 
+import static com.spotify.apollo.Status.BAD_REQUEST;
+import static com.spotify.apollo.Status.INTERNAL_SERVER_ERROR;
+import static com.spotify.apollo.Status.NOT_FOUND;
+import static com.spotify.apollo.Status.NO_CONTENT;
 import static com.spotify.styx.api.Api.Version.V3;
 import static com.spotify.styx.api.Middlewares.json;
 import static com.spotify.styx.serialization.Json.OBJECT_MAPPER;
-import static com.spotify.styx.util.StreamUtil.cat;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
-import com.spotify.apollo.Client;
 import com.spotify.apollo.Request;
 import com.spotify.apollo.RequestContext;
 import com.spotify.apollo.Response;
-import com.spotify.apollo.Status;
+import com.spotify.apollo.entity.EntityMiddleware;
+import com.spotify.apollo.entity.JacksonEntityCodec;
 import com.spotify.apollo.route.AsyncHandler;
 import com.spotify.apollo.route.Route;
+import com.spotify.styx.api.workflow.WorkflowInitializationException;
+import com.spotify.styx.api.workflow.WorkflowInitializer;
 import com.spotify.styx.model.Workflow;
 import com.spotify.styx.model.WorkflowConfiguration;
 import com.spotify.styx.model.WorkflowId;
 import com.spotify.styx.model.WorkflowInstance;
 import com.spotify.styx.model.WorkflowState;
 import com.spotify.styx.model.data.WorkflowInstanceExecutionData;
+import com.spotify.styx.serialization.Json;
 import com.spotify.styx.storage.Storage;
 import com.spotify.styx.util.ResourceNotFoundException;
 import com.spotify.styx.util.WorkflowValidator;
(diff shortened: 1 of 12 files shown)


honnix:
not good. we have internal code relying on `WorkflowConsumerFactory`. :(

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -102,64 +109,67 @@ public WorkflowResource(Storage storage, String schedulerServiceBaseUrl,
...
-  private CompletionStage<Response<ByteString>> deleteWorkflow(String cid, String wfid,
-                                                               RequestContext rc) {
-    // TODO: handle workflow crud directly in api service instead of proxying to scheduler
-    return forwardingClient.send(rc.request().withUri(schedulerApiUrl("workflows", cid, wfid)));
+  private Response<ByteString> deleteWorkflow(String cid, String wfid) {
+    final WorkflowId workflowId = WorkflowId.create(cid, wfid);
+    try {
+      storage.delete(workflowId);
+    } catch (IOException e) {
+      final String message = String.format("Couldn't remove workflow %s. ", workflowId);
```
danielnorberg: Ideally this would return `404` if the workflow didn't exist but datastore delete doesn't tell...

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -102,64 +109,67 @@ public WorkflowResource(Storage storage, String schedulerServiceBaseUrl,
...
+
+    return Response.forPayload(workflow);
   }
 
   private Response<Collection<Workflow>> workflows() {
     try {
       return Response.forPayload(storage.workflows().values());
     } catch (IOException e) {
-      return Response.forStatus(Status.INTERNAL_SERVER_ERROR.withReasonPhrase(
+      return Response.forStatus(INTERNAL_SERVER_ERROR.withReasonPhrase(
```
danielnorberg: Would've been great to have this noise due to static import change in a different PR

styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
```diff
@@ -126,11 +128,11 @@ public WorkflowResourceTest(Api.Version version) {
 
   @Override
   protected void init(Environment environment) {
-    storage = new AggregateStorage(bigtable, datastore, Duration.ZERO);
+    storage = spy(new AggregateStorage(bigtable, datastore, Duration.ZERO));
```
danielnorberg: storage should be mocked instead but that's for another PR

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
danielnorberg: The `SchedulerResource` checked if the workflow existed before attempting to delete it and returned `404` if it didn't exist. Let's do that here as well.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -102,64 +109,67 @@ public WorkflowResource(Storage storage, String schedulerServiceBaseUrl,
...
+    final WorkflowId workflowId = WorkflowId.create(cid, wfid);
+    try {
+      storage.delete(workflowId);
+    } catch (IOException e) {
+      final String message = String.format("Couldn't remove workflow %s. ", workflowId);
+      LOG.warn(message, e);
+      throw new RuntimeException(message, e);
+    }
+    LOG.info("Workflow removed: {}", workflowId);
+    return Response.forStatus(NO_CONTENT);
```
danielnorberg: Also call `workflowConsumer.accept(Optional.of(workflow), Optional.empty());`

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -102,64 +109,67 @@ public WorkflowResource(Storage storage, String schedulerServiceBaseUrl,
...
-                                     .withPayload(payload.get())
-                                     .withUri(schedulerApiUrl("workflows", componentId)));
+    if (oldWorkflowOptional.isPresent()) {
+      LOG.info("Workflow modified, old config: {}, new config: {}", oldWorkflowOptional.get(),
+          workflow);
+    } else {
+      LOG.info("Workflow added: {}", workflow);
+    }
+
+    return Response.forPayload(workflow);
```
danielnorberg: Also call `workflowConsumer.accept(oldWorkflowOptional, Optional.of(workflow));`

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
honnix: That was actually a strange design to me. But yes I guess we need to keep backward compatibility.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
honnix: Sure will get them back.

styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
```diff
@@ -500,28 +518,37 @@ public void shouldReturnErrorMessageWhenSchedulerFailsWorkflow() throws Exceptio
...
+  public void shouldDeleteWorkflow() throws Exception {
+    sinceVersion(Api.Version.V3);
+
+    Response<ByteString> response =
+        awaitResponse(
+            serviceHelper.request("DELETE", path("/foo/bar")));
+
+    assertThat(response, hasStatus(withCode(Status.NO_CONTENT)));
     assertThat(response, hasNoPayload());
+    assertThat(storage.workflow(WORKFLOW.id()), is(Optional.empty()));
```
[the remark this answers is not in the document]
honnix: Yeah right it is a spy now.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -102,64 +109,67 @@ public WorkflowResource(Storage storage, String schedulerServiceBaseUrl,
...
-      return CompletableFuture.completedFuture(
-          Response.forStatus(Status.BAD_REQUEST.withReasonPhrase("Invalid workflow configuration: " + errors)));
+      return Response.forStatus(
+          BAD_REQUEST.withReasonPhrase("Invalid workflow configuration: " + errors));
+    }
+
+    final Workflow workflow = Workflow.create(componentId, workflowConfig);
+    final Optional<Workflow> oldWorkflowOptional;
+    try {
+      oldWorkflowOptional = workflowInitializer.inspectChange(workflow);
```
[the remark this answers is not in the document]
honnix: inspectChangeAndStore?

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
honnix: This needs a bit a rework. This consumer is hook on Scheduler not API, and it requires stats to send metric data, but stats is not available on API. I think we can create an instance of stats for API as well, but not sure what would happen for standalone setup.

honnix:
I would not prioritize this for now and change base branch to master after we merge the other PR.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
honnix: It is already one service for standalone setup. That’s why I’m not sure having two stats instances. Would that make Apollo sad? Maybe inject a stats factory? For standalone that would return a singleton.

danielnorberg: Standalone still treats scheduler and api as separate apps. I think we might as well merge them into a single app to make things simpler.

As a smaller step, yes we could just inject the same stats instance into both the scheduler and api.

honnix:
Continued by #447

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
danielnorberg: Returning `404` if something that was expected to exists didn't exist makes sense to me.

PR #445, opened 2018-04-12
ignore events from previous executions

danielnorberg:
Ugly but does the job. Might want to consider passing in a lambda into `receive` to inspect the run state and decide whether the event should be processed or not inside the transaction.


danielnorberg:
@honnix PTAL

honnix:
I will rebase that branch from master because of change in kube runner. Then let's iterate.

honnix:
This would do the job. Is there any other benefit of this solution over explicit counter?

danielnorberg:
One benefit I guess is that it would eliminate the need for the pod watcher to read the wfi from datastore before injecting the event and starting the state machine transition datastore transaction. Less blocking of the watcher thread by datastore IO seems good. That's not part of this PR though, maybe it should be.

honnix:
After that sidecar PR merged, we can remove watcher completely. I would prefer we keep the interface of QSM smaller without introducing another `receive`. What do you think?

PR #446, opened 2018-04-12
fix dequeue event type check

danielnorberg:
Including resource ID's in type equality check seems fragile.


danielnorberg:
@ulzha PTAL

honnix:
👍

ulzha:
:+1:

PR #447, opened 2018-04-12
remove proxy for workflow operation

honnix:
Close #443


honnix:
@danielnorberg PTAL

Unfortunate for the decrease of coverage, due to deleted code?

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -102,57 +109,79 @@ public WorkflowResource(Storage storage, String schedulerServiceBaseUrl,
...
+      return Response.forStatus(Status.INTERNAL_SERVER_ERROR
+          .withReasonPhrase("Error in internal storage"));
+    }
+    
+    if (!workflow.isPresent()) {
+      return Response.forStatus(Status.NOT_FOUND.withReasonPhrase("Workflow does not exist"));
+    }
+
+    try {
+      storage.delete(workflowId);
```
danielnorberg: Minor detail and not necessary to do in this PR but deletion should probably be atomic. I.e. do the lookup and delete in a transaction. Semantically it seems strange for two concurrent deletes to both be able to succeed due to racy existence checking and removal.

styx-api-service/src/main/java/com/spotify/styx/api/workflow/WorkflowInitializer.java
```diff
@@ -45,7 +45,8 @@ public WorkflowInitializer(Storage storage, Time time) {
     this.time = Objects.requireNonNull(time);
   }
 
-  public Optional<Workflow> inspectChange(Workflow workflow) {
+  public Optional<Workflow> inspectChangeAndStore(Workflow workflow)
```
danielnorberg: Nitpick but the naming here is confusing me. Isn't the primary purpose of this method to handle storing a (new) workflow? Calculating the new next natural trigger is done as well, as a side effect part of storing the workflow. I wouldn't call that "inspecting" and I can't see any other inspection going on. How about just calling this method `storeWorkflow` ?

styx-api-service/src/main/java/com/spotify/styx/api/workflow/WorkflowInitializer.java
```diff
@@ -45,7 +45,8 @@ public WorkflowInitializer(Storage storage, Time time) {
...
+      throws WorkflowInitializationException {
     final Optional<Workflow> previous;
     try {
       previous = storage.workflow(workflow.id());
```
danielnorberg: Looking up the previous workflow, storing the new workflow and updating the trigger should be atomic, i.e. done in a transaction. Otherwise concurrent store requests might end up corrupting the workflow state.

styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
```diff
@@ -467,61 +472,117 @@ public void shouldReturnBadRequestWhenMalformedStatePayloadIsSentWorkflow() thro
...
 
-    assertThat(response, hasStatus(withCode(Status.SERVICE_UNAVAILABLE)));
-    assertThat(response, hasNoPayload());
+    assertThat(response, hasStatus(withCode(Status.OK)));
+    assertThat(deserialize(response.payload().get(), Workflow.class), equalTo(WORKFLOW));
   }
 
   @Test
-  public void shouldForwardInternalResponseForDeleteWorkflow() throws Exception {
+  public void shouldReturnErrorMessageWhenFailedToInspectChange() throws Exception {
```
danielnorberg: `s/Inspect/StoreWorkflow`

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
honnix: I will work on a separate PR to solve atomic issue for both deleting and creating.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -322,7 +308,8 @@ public void create(Environment environment) {
     closer.register(executorCloser("event-consumer", eventConsumerExecutor));
 
     final Stats stats = statsFactory.apply(environment);
-    final Storage storage = instrument(Storage.class, storageFactory.apply(environment), stats, time);
+    final Storage storage = instrument(Storage.class,
```
danielnorberg: Looks like storage is getting instrumented / wrapped in a proxy twice here

styx-api-service/src/main/java/com/spotify/styx/api/workflow/WorkflowInitializer.java
honnix: That makes sense. I don't know why it was named as `inspect`. :)

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
honnix: Why?

styx-standalone-service/src/main/java/com/spotify/styx/StyxService.java
```diff
@@ -20,18 +20,31 @@
...
+  private static Stats STATS;
+
   private StyxService() {
+    throw new UnsupportedOperationException();
   }
 
   public static void main(String[] args) throws LoadingException {
-    final StyxScheduler scheduler = StyxScheduler.createDefault();
-    final StyxApi api = StyxApi.createDefault();
+    final StyxScheduler scheduler = StyxScheduler.newBuilder()
```
danielnorberg: We can avoid the syncronized static stats holder by moving this into the AppInit lambda.

Create the `MetricsStats` once there and pass it into the api and scheduler builders using lambdas.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: Both `instrument` and `new MeteredStorageProxy`

styx-standalone-service/src/main/java/com/spotify/styx/StyxService.java
honnix: Yeah that would look nicer. The builder stuff shouldn't have any side effect to make `AppInit` sad. I will fix.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
danielnorberg: Ah, `instrument` is a static method on the abstract class `MeteredProxy` which takes the concrete subclass as a parameter. That's a bit confusing to me.

honnix: I agree. I have an idea how to fix. :D

danielnorberg: I took the liberty of pushing a commit to fix this =)

danielnorberg:
👍

styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
```diff
@@ -462,61 +467,117 @@ public void shouldReturnBadRequestWhenMalformedStatePayloadIsSentWorkflow() thro
   }
 
   @Test
-  public void shouldReturnOkWhenSchedulerReturnsSuccessWorkflow() throws Exception {
+  public void shoulCreateWorkflow() throws Exception {
```
NatashaL: typo in `should`

styx-common/src/main/java/com/spotify/styx/monitoring/MeteredStorageProxy.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.spotify.styx.storage.Storage;
+import com.spotify.styx.util.Time;
+import java.lang.reflect.Proxy;
+import javaslang.control.Try;
+
+/**
+ * A proxy for instrumenting an instance of {@link Storage} {@link Proxy#newProxyInstance}.
+ */
+public class MeteredStorageProxy extends MeteredProxy<Storage> {
```
NatashaL: 🌟

NatashaL:
👍

PR #448, opened 2018-04-13
[HOLD] strongly consistent wfi listing using shards

danielnorberg:
Active workflow instances are indexed using 128 entity groups (index shards). 

The id's of the instances are stored as the names of child entities in the index shards. The entities themselves are empty.

This avoids the 1MB max entity size limit that might be hit if instance id's are stored as list properties.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -188,10 +188,6 @@ public class DatastoreStorage {
     this.retryBaseDelay = Objects.requireNonNull(retryBaseDelay);
     this.storageTransactionFactory = Objects.requireNonNull(storageTransactionFactory);
     this.forkJoinPool = new ForkJoinPool(REQUEST_CONCURRENCY);
-
-    // TODO: remove after bootstrapping the indexes once
-    // XXX: bootstrap indexes as an offline operation instead of here in the styx scheduler process?
-    indexActiveWorkflowInstances();
   }
 
   void indexActiveWorkflowInstances() {
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -190,6 +190,10 @@ public class DatastoreStorage {
     this.forkJoinPool = new ForkJoinPool(REQUEST_CONCURRENCY);
   }
 
+  /**
+   * Bootstrap active worflow instance index. This cannot run concurrently with the scheduler and must only
+   * be called once before the scheduler starts.
+   */
   void indexActiveWorkflowInstances() {
     // XXX: this listing is not strongly consistent
     final List<Entity> indexEntries = StreamUtil.stream(
@@ -201,6 +205,56 @@ public class DatastoreStorage {
 
     for (List<Entity> batch : Lists.partition(indexEntries, MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH)) {
       datastore.put(batch.toArray(new Entity[0]));
+    }
+  }
+
+  /**
+   * This indexing can run concurrently with the scheduler operations as it is using transactions to ensure it
+   * consistently updates the index, but that also makes it too slow to do the initial bootstrap.
+   *
+   */
+  void concurrentlyIndexActiveWorkflowInstances() {
+    // Weak listing, but we'll eventually read _everything_, in due time...
+    final List<Entity> expectedIndexEntries = StreamUtil.stream(
+        datastore.run(Query.newKeyQueryBuilder().setKind(KIND_ACTIVE_WORKFLOW_INSTANCE).build()))
+        .map(Key::getName)
+        .map(wfiKey -> activeWorkflowInstanceIndexShardEntryKey(datastore.newKeyFactory(), wfiKey))
+        .map(indexEntryKey -> Entity.newBuilder(indexEntryKey).build())
+        .collect(toList());
+
+    // Fetch index entries in batches
+    for (List<Entity> expectedEntryBatch : Lists.partition(expectedIndexEntries, MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH)) {
+
+      final List<Entity> actualEntryBatch = datastore.fetch(expectedEntryBatch.stream().map(Entity::getKey).toArray(Key[]::new));
+      assert actualEntryBatch.size() == expectedEntryBatch.size();
+
+      for (int i = 0; i < expectedEntryBatch.size(); i++) {
+
+        // If an entry is null, it might be missing
+        if (actualEntryBatch.get(i) == null) {
+
+          final Entity expectedEntry = expectedEntryBatch.get(i);
+
+          // Check again in a transaction and add the entry if it is really missing
+          datastore.runInTransaction(tx -> {
+            final String wfiName = expectedEntry.getKey().getName();
+
+            // Is the wfi really there?
+            if (tx.get(activeWorkflowInstanceKey(datastore.newKeyFactory(), wfiName)) == null) {
+              return null;
+            }
+
+            // Is the index entry really missing?
+            if (tx.get(expectedEntry.getKey()) != null) {
+              return null;
+            }
+
+            // Add the missing entry
+            tx.add(expectedEntry);
+            return null;
+          });
+        }
+      }
     }
   }
 
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -172,7 +172,7 @@ public class DatastoreStorage implements Closeable {
   public static final int MAX_RETRIES = 100;
   public static final int MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH = 1000;
 
-  private static final int REQUEST_CONCURRENCY = 2;
+  private static final int REQUEST_CONCURRENCY = 32;
 
   private final Datastore datastore;
   private final Duration retryBaseDelay;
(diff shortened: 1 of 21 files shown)


styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -377,14 +398,18 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
...
-   * states, but the values of all the states returned should be fresh.
+   * Strongly consistently read all active states
    */
   Map<WorkflowInstance, RunState> readActiveStates() throws IOException {
-    // Eventually consistently read active state keys
-    final List<Key> keys = readActiveInstanceKeys();
+    // Strongly read active state keys from index shards
+    final List<Key> keys = activeWorkflowInstanceIndexShardKeys(datastore.newKeyFactory()).stream()
+        .map(key -> forkJoinPool.submit(() ->
+            datastore.run(Query.newEntityQueryBuilder().setFilter(PropertyFilter.hasAncestor(key)).build())))
```
honnix: Does it make sense to put kind as filter as well? Or filtering by ancestor does it implicitly?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -377,14 +398,18 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
   }
 
   /**
-   * Eventually consistently list all active states and strongly consistently fetch their values.
-   *
-   * <p>This method will return a map of active states that might be missing some recently created
-   * states, but the values of all the states returned should be fresh.
+   * Strongly consistently read all active states
```
ulzha: Maybe nitpick, but "strongly consistently" might be misleading. I too lack a short term for the actual consistency mode :/ Thought to go for something like this: https://github.com/spotify/styx/compare/uldis/strong-listing#diff-1ec84406994237ab0330232416a42f1eR477

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -183,6 +190,20 @@
     this.forkJoinPool = new ForkJoinPool(REQUEST_CONCURRENCY);
   }
 
+  void indexActiveWorkflowInstances() {
+    // XXX: this listing is not strongly consistent
+    final List<Entity> indexEntries = StreamUtil.stream(
+        datastore.run(Query.newKeyQueryBuilder().setKind(KIND_ACTIVE_WORKFLOW_INSTANCE).build()))
+        .map(Key::getName)
+        .map(wfiKey -> activeWorkflowInstanceIndexShardEntryKey(datastore.newKeyFactory(), wfiKey))
```
honnix: This will create the ancestor entry if it doesn't exist already, right?

honnix: actually a comment of the line below

styx-common/src/main/java/com/spotify/styx/util/StreamUtil.java
```diff
@@ -40,4 +43,8 @@ private StreamUtil() {
   public static <T> Stream<T> cat(Stream<T>... streams) {
     return Arrays.stream(streams).reduce(Stream.empty(), Stream::concat);
   }
+
+  public static <T> Stream<T> stream(Iterator<T> iterator) {
+    return StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterator, 0), false);
```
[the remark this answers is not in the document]
danielnorberg: sure in a separate PR =)

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: no, the ancestor isn't needed, see comment and docs

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: 👍

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: "strongly consistently" matches the datastore documentation terminology and (I believe) the terminology in the distributed system community as well

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
honnix: OK. As long as it is safe to concurrently store two children referring to the same non-exist parent. Didn't find that in doc.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -183,6 +190,75 @@
...
+        .map(indexEntryKey -> Entity.newBuilder(indexEntryKey).build())
+        .collect(toList());
+
+    // Fetch index entries in batches
+    for (List<Entity> expectedEntryBatch : Lists.partition(expectedIndexEntries, MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH)) {
+
+      final List<Entity> actualEntryBatch = datastore.fetch(expectedEntryBatch.stream()
```
honnix: Is the eager loading intended? It feels easier to understand if we do `datastore.get` and then diff the two lists. Or performance is the concern here because List subtraction is slow? If that is the case, it is fine.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -183,6 +190,75 @@
...
+          .map(Entity::getKey).toArray(Key[]::new));
+      assert actualEntryBatch.size() == expectedEntryBatch.size();
```
honnix: Is this intended? Isn't assertion disabled by default at runtime?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -486,10 +558,42 @@ static RunState entityToRunState(Entity entity, WorkflowInstance instance)
...
+  }
+
+  static List<Key> activeWorkflowInstanceIndexShardKeys(KeyFactory keyFactory) {
+    return IntStream.range(0, ACTIVE_WORKFLOW_INSTANCE_INDEX_SHARDS)
+        .mapToObj(DatastoreStorage::activeWorkflowInstanceIndexShardName)
+        .map(name -> keyFactory.setKind(KIND_ACTIVE_WORKFLOW_INSTANCE_INDEX_SHARD).newKey(name))
+        .collect(toList());
+  }
+
+  static String activeWorkflowInstanceIndexShardName(WorkflowInstance workflowInstance) {
```
honnix: This method is not used.

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerServiceFixture.java
```diff
@@ -114,26 +108,20 @@
...
-  @AfterClass
-  public static void tearDownClass() throws Exception {
-    if (localDatastore != null) {
-      try {
-        localDatastore.stop(org.threeten.bp.Duration.ofSeconds(30));
-      } catch (Throwable e) {
-        e.printStackTrace();
-      }
+    try {
+      localDatastore.start();
```
honnix: This could be very slow stop/start emulator for each test case. If that is to make it more reliable, the code in constructor I think better to to `setUp`.

honnix: This makes circleci very sad. :D

honnix:
One more thing, we might want to use a flag to disable/enable the bootstrap, like: https://github.com/spotify/styx/blob/master/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java#L466

In case for multi-scheduler setup, if both of them try to bootstrap. We can of course have one scheduler and keep it running for a few days until everything converges and then spin up another scheduler.

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerServiceFixture.java
danielnorberg: Yeah, I'm actively experimenting here a bit with how to make the tests work

danielnorberg:
Debugging this PR, trying to get it to work on circleci.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: Yeah, the eager loading was intended. The diffing is happening below.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: Can change this to `if/throw` check instead.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: Yeah, that should be fine. The parent is not created and the docs specify that an entity ancestor does not have to exist. So there's no collision on implicitly creating a parent.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
honnix: `assert` is probably good enough. I'm expecting this thing to fail.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
ulzha: Yes if you're talking about each individual lookup. My point is that this situation is not a lookup but a series of lookups used to obtain one list. Calling the list "strongly consistent" in such a situation is not something that I've seen out there.

ulzha:
We're rebasing this and separating logic changes from CI fixes, in https://github.com/spotify/styx/pull/453 .

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: If you're expecting it to fail then we should probably make it a real `if` check =)

danielnorberg:
Moved system test fixes to #456

PR #449, opened 2018-04-13
Do not use common FJP

honnix:
This is on critical path and trigger is usually very slow due to
datastore operation.

@danielnorberg Let's talk about this.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
@@ -25,6 +25,7 @@ import static com.google.common.base.CaseFormat.UPPER_CAMEL;
 import static com.spotify.styx.util.GuardedRunnable.guard;
 import static com.spotify.styx.util.TimeUtil.nextInstant;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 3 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -68,6 +73,7 @@
     this.time = requireNonNull(time);
     this.storage = requireNonNull(storage);
     this.stats = requireNonNull(stats);
+    this.forkJoinPool = new ForkJoinPool(TRIGGER_CONCURRENCY);
```
danielnorberg: Make the `TriggerManager` `Closeable` and shut down the fjp on scheduler shutdown.

honnix: Good catch, will fix!

honnix: fixed by https://github.com/spotify/styx/pull/449/commits/f4a76013669a7e0123bd1c51698bd1956314f3df

danielnorberg:
👍

PR #450, opened 2018-04-16
Separate process per system test

danielnorberg:


ulzha:
Interesting, why does this decrease coverage? Because coverage is weird?

PR #451, opened 2018-04-16
we were catching wrong exception

honnix:
This causes problem that we are not progressing next natural trigger for
`AlreadyInitializedException`.

@ulzha PTAL

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
@@ -22,6 +22,7 @@ package com.spotify.styx;
 
 import static com.google.common.base.CaseFormat.LOWER_UNDERSCORE;
 import static com.google.common.base.CaseFormat.UPPER_CAMEL;
+import static com.spotify.styx.util.ExceptionUtil.findCause;
 import static com.spotify.styx.util.GuardedRunnable.guard;
 import static com.spotify.styx.util.ParameterUtil.toParameter;
 import static com.spotify.styx.util.TimeUtil.nextInstant;
@@ -44,7 +45,6 @@ import java.time.temporal.ChronoUnit;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletionStage;
-import java.util.concurrent.ExecutionException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -114,11 +117,22 @@ private void tryTriggering(Workflow workflow,
...
-          // nop
-        } catch (Throwable e) {
-          LOG.warn("Triggering {} threw exception", workflow.id(), e);
-          return; // so we don't update the trigger time
+        } catch (ExecutionException e) {
+          final WorkflowInstance workflowInstance = WorkflowInstance.create(workflow.id(),
+              toParameter(workflow.configuration().schedule(), instantSpec.instant()));
+
+          if (e.getCause() instanceof AlreadyInitializedException) {
+            LOG.debug("{} already triggered for {}", workflowInstance.toKey(), e);
```
ulzha: Lacking another argument for `{}`?

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -114,11 +117,22 @@ private void tryTriggering(Workflow workflow,
               instantSpec.instant());
           // Wait for the event to be processed before proceeding to the next trigger
           processed.toCompletableFuture().get();
-        } catch (AlreadyInitializedException ignored) {
-          // nop
-        } catch (Throwable e) {
-          LOG.warn("Triggering {} threw exception", workflow.id(), e);
-          return; // so we don't update the trigger time
+        } catch (ExecutionException e) {
```
ulzha: Are we sure this will never be further wrapped in another exception? I think to more properly mitigate fragility we could just `catch (Exception e)`?

honnix: It shouldn't or we will need to do the `findCause` stuff.

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
honnix: Good catch. Will fix.

ulzha:
Good catch, this entire issue!

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
ulzha: Yeah I mean we just do the `findCause` stuff for `Exception`, instead of having two branches, one for `Exception` that doesn't do `findCause` stuff, and one for `ExecutionException` that does `findCause` stuff and duplicates `Exception` stuff.

ulzha: "It shouldn't" implies relying on humans that they won't wrap it further, that's what I'm worried about.

honnix: Exactly so we are on it.

honnix:
Not refactored `MeteredProxy` usage of `findCause` because of https://github.com/spotify/styx/pull/447 and I will fix it there.

ulzha:
:+1: Nice!

PR #453, opened 2018-04-16
strongly consistent wfi listing using shards

ulzha:
Active workflow instances are indexed using 128 entity groups (index shards).

The id's of the instances are stored as the names of child entities in the index shards. The entities themselves are empty.

This avoids the 1MB max entity size limit that might be hit if instance id's are stored as list properties.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -188,10 +188,6 @@ public class DatastoreStorage {
     this.retryBaseDelay = Objects.requireNonNull(retryBaseDelay);
     this.storageTransactionFactory = Objects.requireNonNull(storageTransactionFactory);
     this.forkJoinPool = new ForkJoinPool(REQUEST_CONCURRENCY);
-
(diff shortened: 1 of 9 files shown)


styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -394,17 +493,6 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
         .collect(toMap(RunState::workflowInstance, Function.identity()));
   }
```
ulzha: This is just one of the methods where we used to do a weakly consistent query. E.g. `readActiveStates(String componentId)` also is used by `styx ls` and will return an inconsistent view... Maybe cool to fix in another pr.

danielnorberg: Caching `readActiveStates` (or similar state) might reintroduce the problem of not being able to read writes.

E.g. a user triggers a wfi and then does a read/listing that returns data from the cached `readActiveStates`. The user might then be surprised that their newly triggered wfi is not included in the response.

PR #455, opened 2018-04-16
trigger backfill in reversed order

honnix:
minimum effort to make this feature. everything being the same, but to
trigger backfill in reversed order, user only needs to put `start` after
`end`, where `start` is inclusive and `end` is exclusive as it is today.

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
@@ -23,9 +23,10 @@ package com.spotify.styx.api;
 import static com.spotify.apollo.StatusType.Family.SUCCESSFUL;
 import static com.spotify.styx.api.Api.Version.V3;
 import static com.spotify.styx.serialization.Json.serialize;
-import static com.spotify.styx.util.ParameterUtil.rangeOfInstants;
 import static com.spotify.styx.util.ParameterUtil.toParameter;
 import static com.spotify.styx.util.StreamUtil.cat;
+import static com.spotify.styx.util.TimeUtil.instantsInRange;
+import static com.spotify.styx.util.TimeUtil.instantsInReversedRange;
 import static java.util.stream.Collectors.toList;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
@@ -281,8 +282,10 @@ public final class BackfillResource {
           Status.BAD_REQUEST.withReasonPhrase("end parameter not aligned with schedule"));
     }
 
+    final boolean reversed = input.start().isAfter(input.end());
+
     final List<WorkflowInstance> alreadyActive =
-        rangeOfInstants(input.start(), input.end(), schedule).stream()
+        instants(input.start(), input.end(), schedule, reversed).stream()
             .map(instant -> WorkflowInstance.create(workflowId, toParameter(schedule, instant)))
             .filter(activeWorkflowInstances::contains)
             .collect(toList());
@@ -347,8 +350,10 @@ public final class BackfillResource {
       throw new RuntimeException(e);
     }
 
-    final List<Instant> processedInstants = rangeOfInstants(
-        backfill.start(), backfill.nextTrigger(), backfill.schedule());
+    final boolean reversed = backfill.start().isAfter(backfill.end());
+
+    final List<Instant> processedInstants = instants(
+        backfill.start(), backfill.nextTrigger(), backfill.schedule(), reversed);
     processedStates = processedInstants.parallelStream()
         .map(instant -> {
           final WorkflowInstance wfi = WorkflowInstance
@@ -367,8 +372,8 @@ public final class BackfillResource {
         })
         .collect(toList());
 
-    final List<Instant> waitingInstants = rangeOfInstants(
-        backfill.nextTrigger(), backfill.end(), backfill.schedule());
+    final List<Instant> waitingInstants = instants(
+        backfill.nextTrigger(), backfill.end(), backfill.schedule(), reversed);
     waitingStates = waitingInstants.stream()
         .map(instant -> {
           final WorkflowInstance wfi = WorkflowInstance.create(
@@ -378,5 +383,11 @@ public final class BackfillResource {
         .collect(toList());
 
     return Stream.concat(processedStates.stream(), waitingStates.stream()).collect(toList());
+  }
+
+  private static List<Instant> instants(Instant start, Instant end, Schedule schedule,
+                                        boolean reversed) {
+    return reversed ? instantsInReversedRange(start, end, schedule) :
+           instantsInRange(start, end, schedule);
   }
 }
--- a/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
@@ -282,10 +282,8 @@ public final class BackfillResource {
           Status.BAD_REQUEST.withReasonPhrase("end parameter not aligned with schedule"));
     }
 
-    final boolean reversed = input.start().isAfter(input.end());
-
     final List<WorkflowInstance> alreadyActive =
-        instants(input.start(), input.end(), schedule, reversed).stream()
+        instants(input.start(), input.end(), schedule).stream()
             .map(instant -> WorkflowInstance.create(workflowId, toParameter(schedule, instant)))
             .filter(activeWorkflowInstances::contains)
             .collect(toList());
@@ -350,10 +348,8 @@ public final class BackfillResource {
       throw new RuntimeException(e);
     }
 
-    final boolean reversed = backfill.start().isAfter(backfill.end());
-
     final List<Instant> processedInstants = instants(
-        backfill.start(), backfill.nextTrigger(), backfill.schedule(), reversed);
+        backfill.start(), backfill.nextTrigger(), backfill.schedule());
     processedStates = processedInstants.parallelStream()
         .map(instant -> {
           final WorkflowInstance wfi = WorkflowInstance
@@ -373,7 +369,7 @@ public final class BackfillResource {
         .collect(toList());
 
     final List<Instant> waitingInstants = instants(
-        backfill.nextTrigger(), backfill.end(), backfill.schedule(), reversed);
+        backfill.nextTrigger(), backfill.end(), backfill.schedule());
     waitingStates = waitingInstants.stream()
         .map(instant -> {
           final WorkflowInstance wfi = WorkflowInstance.create(
@@ -385,9 +381,8 @@ public final class BackfillResource {
     return Stream.concat(processedStates.stream(), waitingStates.stream()).collect(toList());
   }
 
-  private static List<Instant> instants(Instant start, Instant end, Schedule schedule,
-                                        boolean reversed) {
-    return reversed ? instantsInReversedRange(start, end, schedule) :
+  private static List<Instant> instants(Instant start, Instant end, Schedule schedule) {
+    return start.isAfter(end) ? instantsInReversedRange(start, end, schedule) :
            instantsInRange(start, end, schedule);
   }
 }
--- a/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
@@ -26,10 +26,11 @@ import static com.spotify.styx.serialization.Json.serialize;
 import static com.spotify.styx.util.ParameterUtil.toParameter;
 import static com.spotify.styx.util.StreamUtil.cat;
 import static com.spotify.styx.util.TimeUtil.instantsInRange;
-import static com.spotify.styx.util.TimeUtil.instantsInReversedRange;
+import static com.spotify.styx.util.TimeUtil.nextInstant;
 import static java.util.stream.Collectors.toList;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.collect.Iterables;
 import com.spotify.apollo.Client;
 import com.spotify.apollo.Request;
 import com.spotify.apollo.RequestContext;
@@ -272,6 +273,11 @@ public final class BackfillResource {
           + String.join(", ", errors)));
     }
 
+    if (!input.start().isBefore(input.end())) {
+      return Response.forStatus(
+          Status.BAD_REQUEST.withReasonPhrase("start must be before end"));
+    }
+
     if (!TimeUtil.isAligned(input.start(), schedule)) {
       return Response.forStatus(
           Status.BAD_REQUEST.withReasonPhrase("start parameter not aligned with schedule"));
@@ -282,8 +288,9 @@ public final class BackfillResource {
           Status.BAD_REQUEST.withReasonPhrase("end parameter not aligned with schedule"));
     }
 
+    final List<Instant> instants = instantsInRange(input.start(), input.end(), schedule);
     final List<WorkflowInstance> alreadyActive =
-        instants(input.start(), input.end(), schedule).stream()
+        instants.stream()
             .map(instant -> WorkflowInstance.create(workflowId, toParameter(schedule, instant)))
             .filter(activeWorkflowInstances::contains)
             .collect(toList());
@@ -305,8 +312,11 @@ public final class BackfillResource {
         .start(input.start())
         .end(input.end())
         .schedule(schedule)
-        .nextTrigger(input.start())
+        .nextTrigger(input.reverse()
+            ? Iterables.getLast(instants)
+            : input.start())
         .description(input.description())
+        .reverse(input.reverse())
         .halted(false);
 
     final Backfill backfill = builder.build();
@@ -348,8 +358,13 @@ public final class BackfillResource {
       throw new RuntimeException(e);
     }
 
-    final List<Instant> processedInstants = instants(
-        backfill.start(), backfill.nextTrigger(), backfill.schedule());
+    final List<Instant> processedInstants;
+    if (backfill.reverse()) {
+      final Instant firstInstant = nextInstant(backfill.nextTrigger(), backfill.schedule());
+      processedInstants = instantsInRange(firstInstant, backfill.end(), backfill.schedule());
+    } else {
+      processedInstants = instantsInRange(backfill.start(), backfill.nextTrigger(), backfill.schedule());
+    }
     processedStates = processedInstants.parallelStream()
         .map(instant -> {
           final WorkflowInstance wfi = WorkflowInstance
@@ -368,8 +383,13 @@ public final class BackfillResource {
         })
         .collect(toList());
 
-    final List<Instant> waitingInstants = instants(
-        backfill.nextTrigger(), backfill.end(), backfill.schedule());
+    final List<Instant> waitingInstants;
+    if (backfill.reverse()) {
+      final Instant lastInstant = nextInstant(backfill.nextTrigger(), backfill.schedule());
+      waitingInstants = instantsInRange(backfill.start(), lastInstant, backfill.schedule());
+    } else {
+      waitingInstants = instantsInRange(backfill.nextTrigger(), backfill.end(), backfill.schedule());
+    }
     waitingStates = waitingInstants.stream()
         .map(instant -> {
           final WorkflowInstance wfi = WorkflowInstance.create(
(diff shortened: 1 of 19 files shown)


honnix:
@ulzha wdyt of this minimum effort? The user facing side is, if `start` is after `end`, it is reversed.

styx-common/src/main/java/com/spotify/styx/util/TimeUtil.java
```diff
@@ -122,17 +125,53 @@ public static int instancesInRange(Instant firstInstant, Instant lastInstant, Sc
...
     }
-    return number;
+
+    return listOfInstants;
+  }
+
+  /**
+   * Gets the number of instants between firstInstant (inclusive) and lastInstant (exclusive)
+   * according to the {@link Schedule}. This works in a reversed order, meaning firstInstant should
+   * be after lastExecution.
```
ulzha: Typo, `lastInstant`?

ulzha:
A bit worried that the appearance of these backfills in UI will be confusing, people will be expecting the first chronological instance in START column in `styx backfill ls`...

styx-common/src/main/java/com/spotify/styx/util/TimeUtil.java
honnix: Seems like yes.

honnix:
That's hard to say. I would expect the reversed order as I exactly asked. :)

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -281,8 +282,10 @@ private static boolean isActiveState(RunStateData runStateData) {
           Status.BAD_REQUEST.withReasonPhrase("end parameter not aligned with schedule"));
     }
 
+    final boolean reversed = input.start().isAfter(input.end());
```
danielnorberg: Can this check move into `instants()` ?

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -347,8 +350,10 @@ private static boolean isActiveState(RunStateData runStateData) {
       throw new RuntimeException(e);
     }
 
-    final List<Instant> processedInstants = rangeOfInstants(
-        backfill.start(), backfill.nextTrigger(), backfill.schedule());
+    final boolean reversed = backfill.start().isAfter(backfill.end());
```
danielnorberg: same as above

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -342,6 +352,32 @@ public void shouldGetBackfillStatusWithStatusByDefault() throws Exception {
     assertJson(response, "statuses.active_states", hasSize(24));
   }
 
+  @Test
+  public void shouldGetBackfillStatusWithStatusByDefaultReversed() throws Exception {
+    sinceVersion(Api.Version.V3);
+
+    WorkflowInstance wfi = WorkflowInstance.create(BACKFILL_2.workflowId(), "2017-01-01T23");
```
danielnorberg: storage should be mocked out in these tests

danielnorberg: Maybe there's a bit too much event replay logic in `BackfillResource` itself. Factoring out that should ease testing. I.e., it should be possible to mock `getBackfillRunState()` in `BackfillResourceTest` and test `getBackfillRunState()` in a separate `ReplayEventsTest`.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
honnix: It wouldn't change the logic, but would change the semantics a bit though. Instead of checking only `start` and `end`, that would check `start`, `nextTrigger`, and `end`.

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
danielnorberg: can be another PR I guess

styx-common/src/main/java/com/spotify/styx/util/TimeUtil.java
```diff
@@ -104,15 +106,16 @@ public static Instant nextInstant(Instant instant, Schedule schedule) {
   }
 
   /**
-   * Gets the number of workflow instances between firstInstant (inclusive) and lastInstant (exclusive)
-   * according to the workflow's schedule.
+   * Gets the number of instants between firstInstant (inclusive) and lastInstant (exclusive)
```
danielnorberg: Nitpick: It actually gets an ordered list of the instants, not the number.

styx-common/src/main/java/com/spotify/styx/util/TimeUtil.java
```diff
@@ -122,17 +125,53 @@ public static int instancesInRange(Instant firstInstant, Instant lastInstant, Sc
     }
 
     final ExecutionTime executionTime = ExecutionTime.forCron(cron(schedule));
+    final List<Instant> listOfInstants = new ArrayList<>();
```
danielnorberg: Nitpick: hungarian notation? I would just call this `instants` =)

styx-common/src/test/java/com/spotify/styx/util/TimeUtilTest.java
```diff
@@ -21,12 +21,16 @@
 package com.spotify.styx.util;
 
 import static com.spotify.styx.util.TimeUtil.addOffset;
-import static com.spotify.styx.util.TimeUtil.instancesInRange;
+import static com.spotify.styx.util.TimeUtil.instantsInRange;
+import static com.spotify.styx.util.TimeUtil.instantsInReversedRange;
 import static com.spotify.styx.util.TimeUtil.isAligned;
 import static com.spotify.styx.util.TimeUtil.lastInstant;
 import static com.spotify.styx.util.TimeUtil.nextInstant;
+import static java.time.Instant.parse;
```
danielnorberg: This static import produces a lot of noise in this file, makes it harder to check what actually changed.

danielnorberg: Although the github diff highlights the actual changes, so no big deal

styx-common/src/test/java/com/spotify/styx/util/TimeUtilTest.java
```diff
@@ -187,53 +191,149 @@ public void shouldAddOffsetWithWeek() {
...
-    final Instant firstTimeWeeks = Instant.parse("2016-01-11T00:00:00.00Z");
-    final Instant lastTimeWeeks = Instant.parse("2016-01-18T00:00:00.00Z");
-    final Instant firstTimeMonths = Instant.parse("2010-01-01T00:00:00.00Z");
-    final Instant lastTimeMonths = Instant.parse("2016-01-01T00:00:00.00Z");
-
-    assertThat(instancesInRange(firstTimeHours, lastTimeHours, Schedule.HOURS), is(9));
-    assertThat(instancesInRange(firstTimeDays, lastTimeDays, Schedule.DAYS), is(9));
-    assertThat(instancesInRange(firstTimeWeeks, lastTimeWeeks, Schedule.WEEKS), is(1));
-    assertThat(instancesInRange(firstTimeMonths, lastTimeMonths, Schedule.YEARS), is(6));
+    final Instant firstTimeHours = parse("2016-01-19T00:00:00.00Z");
```
danielnorberg: Nitpick: `NumberOf` in the names of all these tests should now be removed? E.g. `shouldGetCorrectNumberOfInstants` becomes `shouldGetCorrectInstants`.

styx-common/src/test/java/com/spotify/styx/util/TimeUtilTest.java
```diff
@@ -187,53 +191,149 @@ public void shouldAddOffsetWithWeek() {
...
+  public void shouldReturnEmptyList() {
+    final Instant firstTimeHours = parse("2016-01-19T00:00:00.00Z");
+    final Instant lastTimeHours = parse("2016-01-19T00:00:00.00Z");
+
+    assertTrue(
+        instantsInRange(firstTimeHours, lastTimeHours, Schedule.parse("0 * * * *")).isEmpty());
   }
 
   @Test(expected = IllegalArgumentException.class)
   public void shouldGetExceptionIfLastInstantIsBeforeFirstInstant() {
```
danielnorberg: Maybe not part of this PR but this use of `@Test(expected = IllegalArgumentException.class)` might produce false positives for these tests if `parse` throws.

styx-common/src/test/java/com/spotify/styx/util/TimeUtilTest.java
```diff
@@ -187,53 +191,149 @@ public void shouldAddOffsetWithWeek() {
...
+    final Instant firstTimeHours = parse("2016-01-19T03:00:00.00Z");
+    final Instant lastTimeHours = parse("2016-01-19T00:00:00.00Z");
+    final Instant firstTimeDays = parse("2016-01-13T00:00:00.00Z");
+    final Instant lastTimeDays = parse("2016-01-10T00:00:00.00Z");
+    final Instant firstTimeWeeks = parse("2016-01-18T00:00:00.00Z");
+    final Instant lastTimeWeeks = parse("2016-01-11T00:00:00.00Z");
+    final Instant firstTimeMonths = parse("2013-01-01T00:00:00.00Z");
+    final Instant lastTimeMonths = parse("2010-01-01T00:00:00.00Z");
+
+    assertThat(instantsInReversedRange(firstTimeHours, lastTimeHours, Schedule.HOURS),
```
danielnorberg: Is it expected that `instantsInRange(x, y, ...)` and `instantsInReversedRange(y, x, ...)` do not produce the same set of instants?

If the user supplies the same start/end timestamps to the backfill create command but in reverse, should it not produce the exact same workflow instances but just in reverse?

styx-common/src/main/java/com/spotify/styx/util/TimeUtil.java
```diff
@@ -104,15 +106,16 @@ public static Instant nextInstant(Instant instant, Schedule schedule) {
...
+   * according to the {@link Schedule}.
    *
    * @param firstInstant The first instant
    * @param lastInstant  The last instant
    * @param schedule     The schedule of the workflow
-   * @return the number of instances within the range
+   * @return instants within the range
    */
-  public static int instancesInRange(Instant firstInstant, Instant lastInstant, Schedule schedule) {
+  public static List<Instant> instantsInRange(Instant firstInstant, Instant lastInstant,
```
danielnorberg: Check that `firstInstant.isBefore(lastInstant)` ?

styx-common/src/main/java/com/spotify/styx/util/TimeUtil.java
```diff
@@ -122,17 +125,53 @@ public static int instancesInRange(Instant firstInstant, Instant lastInstant, Sc
...
+   * Gets the number of instants between firstInstant (inclusive) and lastInstant (exclusive)
+   * according to the {@link Schedule}. This works in a reversed order, meaning firstInstant should
+   * be after lastInstant.
+   *
+   * @param firstInstant The first instant
+   * @param lastInstant  The last instant
+   * @param schedule     The schedule of the workflow
+   * @return instants within the range
+   */
+  public static List<Instant> instantsInReversedRange(Instant firstInstant, Instant lastInstant,
```
danielnorberg: check that `firstInstant.isAfter(lastInstant)` ?

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -379,4 +380,9 @@ private static boolean isActiveState(RunStateData runStateData) {
 
     return Stream.concat(processedStates.stream(), waitingStates.stream()).collect(toList());
   }
+
+  private static List<Instant> instants(Instant start, Instant end, Schedule schedule) {
+    return start.isAfter(end) ? instantsInReversedRange(start, end, schedule) :
```
danielnorberg: In order to always return the same set of instants, maybe this should be:

```
return start.isAfter(end))
  ? Lists.reverse(instantsInRange(end, start, schedule))
  : instantsInRange(start, end, schedule);
```

If not, at least please format the ternary conditional operator like this (when it does not fit on one line):
```
return start.isAfter(end) 
  ? instantsInReversedRange(start, end, schedule) 
  : instantsInRange(start, end, schedule);
```

honnix: It will be different because of inclusive and exclusive. Will fix the format.

styx-common/src/test/java/com/spotify/styx/util/TimeUtilTest.java
honnix: It is expected due to inclusive (x) and exclusive (y).

danielnorberg: Yeah, and I think that might be quite confusing for the user. Might it be easier to use if `styx backfill create <component> <workflow> 2017-01-01 2018-01-01 <concurrency>` and 
`styx backfill create <component> <workflow> 2018-01-01 2017-01-01 <concurrency>` result in the same partitions, just the order reversed?

styx-common/src/test/java/com/spotify/styx/util/TimeUtilTest.java
```diff
@@ -244,29 +245,44 @@ public void shouldReturnEmptyList() {
...
 
-  @Test(expected = IllegalArgumentException.class)
+  @Test
   public void shouldGetExceptionIfLastInstantIsBeforeFirstInstant() {
     final Instant firstTimeHours = parse("2016-01-19T10:00:00.00Z");
     final Instant lastTimeHours = parse("2016-01-19T09:00:00.00Z");
-    instantsInRange(firstTimeHours, lastTimeHours, Schedule.HOURS);
+    try {
+      instantsInRange(firstTimeHours, lastTimeHours, Schedule.HOURS);
+      fail();
```
danielnorberg: Use `ExpectedException` rule

honnix: Thought about this. What is the guideline for using this rule? In case it is quite a bit code achieving the same goal.

styx-common/src/test/java/com/spotify/styx/util/TimeUtilTest.java
honnix: I think it might be OK because we have been explicitly stating the `end` is exclusive, so unless we want to make both of them inclusive for normal and reversed order, it might be more confusing.

styx-common/src/test/java/com/spotify/styx/util/TimeUtilTest.java
honnix: nvm, it is worth it.

styx-common/src/test/java/com/spotify/styx/util/TimeUtilTest.java
danielnorberg: Maybe this needs some user testing.

danielnorberg: The more I think about this, the more it seems like having an explicit `reverse` flag might be better, both for humans using the cli and for machines calling the api.

E.g. if we add backfill creation to System-Z and then have a `reverse` check box that works by swapping the dates and shifting them 1 partition in one direction, that seems awkward to me. Then I would rather just have a `reverse=true|false` parameter.

And `styx backfill create --reverse ...` seems more clear than swapping start/end.

styx-client/src/test/java/com/spotify/styx/client/StyxApolloClientTest.java
```diff
@@ -206,6 +206,43 @@ public void shouldCreateBackfill() throws Exception {
...
+    final BackfillInput backfillInput = BackfillInput.newBuilder()
+        .start(start)
+        .end(end)
+        .component("foo-comp")
+        .workflow("bar-wf")
+        .concurrency(1)
+        .reverse(true)
+        .build();
+
+    final Backfill backfill = Backfill.newBuilder()
```
NatashaL: The values for `backfillInput` and this `backfill` seem magic. Either extract them, or use them once and then reuse them, like done here https://github.com/spotify/styx/pull/455/files?utf8=%E2%9C%93&diff=split#diff-ae03c6a48f8fa80596ad167c226fdde5R272

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -158,7 +160,8 @@ boolean triggerNextPartitionAndProgress(StorageTransaction tx,
                                           String id,
                                           Workflow workflow,
                                           Instant initialNextTrigger,
-                                          int remainingCapacity) {
+                                          int remainingCapacity,
+                                          boolean reversed) {
     final Backfill momentBackfill = tx.backfill(id).orElseThrow(() ->
```
NatashaL: I understand this is older, but why `momentBackfill`? Why not just backfill?

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -169,13 +172,13 @@ boolean triggerNextPartitionAndProgress(StorageTransaction tx,
 
     final Instant momentNextTrigger = momentBackfill.nextTrigger();
```
NatashaL: Same as before, why use `moment` ?

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -223,11 +226,45 @@ boolean triggerNextPartitionAndProgress(StorageTransaction tx,
...
+
+  private static Instant getNextPartition(Backfill momentBackfill,
+                                   Instant momentNextTrigger,
+                                   boolean reversed) {
+    return reversed
+        ? previousInstant(momentNextTrigger, momentBackfill.schedule())
+        : nextInstant(momentNextTrigger, momentBackfill.schedule());
+  }
+
+  private static boolean capacityReached(Backfill momentBackfill, Instant initialNextTrigger,
```
NatashaL: As far as code-style goes, I don't mind using either ternary of if-else, but let's stick to one and the same, where it applies. `exclusiveEndTrigger`, `getNextPartition` and `capacityReached` can all be written using the same notation

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
honnix: Just to make it clear this is a fresh backfill read from database at this moment.

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -395,6 +432,36 @@ public void shouldPostBackfill() throws Exception {
...
+    assertThat(postedBackfill.id().matches("backfill-[\\d-]+"), is(true));
+    assertThat(postedBackfill.start(), equalTo(Instant.parse("2017-01-01T00:00:00Z")));
+    assertThat(postedBackfill.end(), equalTo(Instant.parse("2017-02-01T00:00:00Z")));
+    assertThat(postedBackfill.workflowId(), equalTo(WorkflowId.create("component", "workflow2")));
+    assertThat(postedBackfill.concurrency(), equalTo(1));
+    assertThat(postedBackfill.description(), equalTo(Optional.empty()));
+    assertThat(postedBackfill.nextTrigger(), equalTo(Instant.parse("2017-01-31T23:00:00Z")));
+    assertThat(postedBackfill.schedule(), equalTo(Schedule.HOURS));
+    assertThat(postedBackfill.allTriggered(), equalTo(false));
+    assertThat(postedBackfill.halted(), equalTo(false));
```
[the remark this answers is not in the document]
honnix: this has been fixed.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
honnix: It was named like this when both the `old` backfill and the one just read were in the same method, and stayed unchanged over refactorings. It is totally ok to rename if it creates confusion now.

PR #457, opened 2018-04-17
[DO NOT MERGE] repro unbounded datastore client memory usage

danielnorberg:


danielnorberg:
This seems to be due to the client reading the library version from resources on every http request to be able to populate the `x-goog-api-client` header. In doing so, it is leaking `InputStream` instances because it is not closing them.

This causes open file handles to leak until the GC runs the finalizers. This explains the unbounded process memory usage growth without java heap & "direct memory" usage growth.

This seems to be fixed in newer versions of the datastore / google cloud / gax libraries.

This PR bumps these dependencies: https://github.com/spotify/styx/pull/434

PR #462, opened 2018-04-17
add debug log so we know when it converges

honnix:
@danielnorberg Does this make sense?

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -267,6 +267,9 @@ public class DatastoreStorage implements Closeable {
               return null;
             }
 
+            LOG.debug("adding the missing entry {} to {}", expectedEntry,
+                KIND_ACTIVE_WORKFLOW_INSTANCE_INDEX_SHARD_ENTRY);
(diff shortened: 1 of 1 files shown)


styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -267,6 +267,9 @@ void concurrentlyIndexActiveWorkflowInstances() {
               return null;
             }
 
+            LOG.debug("adding the missing entry {} to {}", expectedEntry,
```
danielnorberg: Are we running with debug logging enabled?

Perhaps the log message should mention workflow instances and index?

I'd be tempted to track this using a metric but I guess we won't be keeping this around for more than a week or so.

danielnorberg:
👍 except for comments

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
honnix: Yeah we do now. Will add more info to the message. Didn’t want to bother metrics as you said.

honnix:
We can all remove the syncing code I believe. Nothing has been observed.

PR #466, opened 2018-04-18
remove redundant wfi fetch

danielnorberg:
Use the full map of wfis we already have.


honnix:
We have so many wasted CPU cycles (and now IOs). :( I guess we didn't care before because reading WFI was cheap. Will take a look later in details.

honnix:
👍

PR #470, opened 2018-04-19
remove old execution id migration code

danielnorberg:
All currently active WFIs have execution id assigned in PREPARE by the `ExecutionDescriptionHandler`.


honnix:
👍

PR #471, opened 2018-04-19
unused code after sync

honnix:
@danielnorberg PTAL

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -325,10 +325,6 @@ public class StyxScheduler implements AppInit {
     final Storage storage = instrument(Storage.class, storageFactory.apply(environment), stats, time);
     closer.register(storage);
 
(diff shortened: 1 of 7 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -394,7 +390,7 @@ public void create(Environment environment) {
     startScheduler(scheduler, executor);
     startRuntimeConfigUpdate(styxConfig, executor, dequeueRateLimiter);
     startCleaner(cleaner, executor);
-    startConcurrentWorkflowInstanceIndexing(storage, executor);
+
```
danielnorberg: remove empty line?

danielnorberg:
👍

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
honnix: More or less intentionally. Next statement is not startSomething. :D

danielnorberg: ok =)

Issue #435, opened 2018-04-09
State transitions are not retried


ulzha:
When a state transition fails (possible due to an intermittent problem, like Datastore transaction exception), the transition is not retried even once. Styx rather waits for the state to become stale (exceed the configured ttl, normally on the order of minutes or hours) and then issues a timeout event.

This introduces unpredictability in Styx behavior, now that we expect transaction collisions somewhat often.  The possible user experience is that the workflow instance isn't starting for several minutes. The same thing also manifests in flaky system tests that wait for Docker runs to happen.

honnix:
fixed by https://github.com/spotify/styx/pull/688

PR #472, opened 2018-04-19
wip: at-least-once state machine transitions

danielnorberg:
Would address #435


honnix:
It would be very helpful if we can summarize how the change would work.

PR #474, opened 2018-04-19
remove ENDPOINT_ID env var

danielnorberg:


honnix:
👍 Poked users who are still use this.

PR #476, opened 2018-04-19
remove outdated comments

honnix:
Now both retrieving active states and workflows are strongly consistent, so the metric should be as well.

@danielnorberg PTAL
cc @ulzha


ulzha:
:-1: No the lists aren't guaranteed to be strongly consistent. I've been [picking](https://github.com/spotify/styx/pull/448#discussion_r181302157) on our use of "strongly consistent", and this a case in point. Could we rather tweak the misleading documentation in https://github.com/spotify/styx/blob/master/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java ?

honnix:
Yeah, that makes sense for batches of get.

PR #478, opened 2018-04-20
Check for correct exception type when triggering already triggered instances 

NatashaL:
When users call `styx trigger` on an already running partition, they only saw `API error: 500 Internal Server Error`, which was wrong and lacking more information. This corrects the return code to 409 instead, and adds the exception stack trace in cases of status code 500.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
@@ -279,7 +279,7 @@ public class SchedulerResource {
         // TODO: propagate error information using a more specific exception type
         return Response.forStatus(CONFLICT.withReasonPhrase(cause.getMessage()));
       } else {
-        return Response.forStatus(INTERNAL_SERVER_ERROR);
+        return Response.forStatus(INTERNAL_SERVER_ERROR.withReasonPhrase(cause.getMessage()));
       }
     }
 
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -278,8 +279,12 @@ private StatusType eventInjectorHelper(Event event) {
           || cause instanceof IllegalArgumentException) {
         // TODO: propagate error information using a more specific exception type
         return Response.forStatus(CONFLICT.withReasonPhrase(cause.getMessage()));
+      } else if (cause instanceof AlreadyInitializedException) {
```
honnix: It better to use `findCause` in `ExceptionUtil`. The whole exception handling part would be more robust.

honnix:
Also, let's add test cases to cover those changes.

styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -273,13 +275,17 @@ private StatusType eventInjectorHelper(Event event) {
...
-      if (cause instanceof IllegalStateException
-          || cause instanceof IllegalArgumentException) {
+      Throwable cause;
+      if ((cause = findCause(e, IllegalStateException.class)) != null
+          || (cause = findCause(e, IllegalArgumentException.class)) != null) {
         // TODO: propagate error information using a more specific exception type
         return Response.forStatus(CONFLICT.withReasonPhrase(cause.getMessage()));
+      } else if ((cause = findCause(e, AlreadyInitializedException.class)) != null) {
+        return Response.forStatus(CONFLICT.withReasonPhrase(
+            "This workflow instance is already triggered. Did you want to `retry` running it instead? " + cause
```
honnix: I think we can skip `cause.getMessage` here as the message is clear enough.

PR #480, opened 2018-04-23
briefly talk about HA in readme

honnix:
It turned out we never mentioned how to set up API and Scheduler
separately. It's pretty much impossible to describe HA in more details
unless we open up a whole doc talking about thow to set up for different
scenarios than standalone.


danielnorberg:
👍

PR #481, opened 2018-04-26
configurable event processing thread pool size

danielnorberg:
E.g. `styx.event-processing-threads = 64`


honnix:
Update internal configuration to reflect such a configuration option exists? A commented out config or something similar.

PR #482, opened 2018-04-26
Revert "remove redundant wfi fetch"

honnix:
Reverts spotify/styx#466

As we discussed, we might want to consider bring it back. @danielnorberg

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -77,7 +77,6 @@ import java.io.IOException;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Arrays;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


honnix:
cc @NatashaL

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -472,20 +472,22 @@ void pollPods() {
...
       final Optional<WorkflowInstance> workflowInstance = readPodWorkflowInstance(pod);
       if (!workflowInstance.isPresent()) {
         continue;
       }
-      final RunState runState = activeStates.get(workflowInstance.get());
-      if (runState != null && isPodRunState(pod, runState)) {
-        emitPodEvents(Watcher.Action.MODIFIED, pod, runState);
+      // do an extra lookup to get latest state
+      // this is crucial because after fetching all states in datastore, some states might
+      // have transitioned already
```
danielnorberg: We might even observe pods whose states where created after fetching the states.

honnix: Yeah good catch. Could you help fix this? I’m afk now. :)

danielnorberg:
Looks to me like there's additional race conditions in this polling code. Maybe we should spend some time on it.

honnix:
I'm attempted to revert the change to mitigate current issue. Seems quite some cases in the past week. The other race conditions don't seem to have bad impact.

danielnorberg:
👍

PR #483, opened 2018-04-27
Fix decrements when shard has a lot of excess

NatashaL:


honnix:
If we do need to roll out this, we can temporarily revert https://github.com/spotify/styx/pull/434 first to avoid risky deployment.

honnix:
👍

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
```diff
@@ -226,6 +226,34 @@ public void shouldDecrementShardWithExcessUsage() {
     assertEquals(1L, shardedCounter.getCounter(COUNTER_ID1));
   }
 
+  @Test
+  public void shouldDecrementShardWithALotOfExcessUsage() {
```
danielnorberg: Is this an implementation of `decreaseLimitBelowCurrentCounterValue` ? Should we then remove `decreaseLimitBelowCurrentCounterValue` and the TODO?

NatashaL:
> I guess we should also test the scenarios delta < 0 && newShardValue < 0 and delta > 0 && !Range.closed(0L, shardCapacity).contains(newShardValue) to get full coverage of those if statements.

I'd say those are already covered by `shouldFailDecrementingEmptyCounter` and `shouldFailIncrementingFullCounter`

I added a test to cover these blocks https://github.com/spotify/styx/blob/master/styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java#L158 https://github.com/spotify/styx/blob/master/styx-common/src/main/java/com/spotify/styx/util/ShardedCounter.java#L137

honnix:
👍

PR #485, opened 2018-04-27
migrate workflow from having ancestor to flat structure 

honnix:
Instead of using ancestor, make `component id` part of workflow key.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
@@ -41,6 +41,7 @@ import static com.spotify.styx.storage.DatastoreStorage.activeWorkflowInstanceIn
 import static com.spotify.styx.storage.DatastoreStorage.activeWorkflowInstanceKey;
 import static com.spotify.styx.storage.DatastoreStorage.entityToBackfill;
 import static com.spotify.styx.storage.DatastoreStorage.entityToRunState;
+import static com.spotify.styx.storage.DatastoreStorage.getWorkflowOpt;
 import static com.spotify.styx.storage.DatastoreStorage.instantToTimestamp;
 import static com.spotify.styx.storage.DatastoreStorage.runStateToEntity;
 import static com.spotify.styx.util.ShardedCounter.KIND_COUNTER_LIMIT;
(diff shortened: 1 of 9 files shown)


honnix:
Do we want to do concurrent migration in case initial migration didn't read all workflows due to weak consistency? Or we can just let it migrate organically by reading and storing during normal operation?

honnix:
Didn't bother testing some of the to be removed code.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
```diff
@@ -154,28 +155,31 @@ private Entity resourceToEntity(Datastore datastore, Resource resource) {
 
   @Override
   public WorkflowId store(Workflow workflow) throws IOException {
-    final Key componentKey = DatastoreStorage.componentKey(tx.getDatastore().newKeyFactory(), workflow.componentId());
```
danielnorberg: 💯 really want this

danielnorberg:
When deploying this we'll want to have only one api instance running so we don't end up losing some writes.

Another option would be to:
1. first make styx that writes to the old key and read from both new and old wf key.
2. then make styx also write to the new key.

Otherwise one old api instance might write to the old wf key after a new instance has already written to the new wf key.

honnix:
I think we can afford temporarily having only one API instance to avoid deploying twice and additional code change.

honnix:
Need to incorporate the changes made in #526

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
```diff
@@ -179,23 +178,25 @@ public WorkflowId store(Workflow workflow) throws IOException {
...
+            StringValue.newBuilder(json).setExcludeFromIndexes(true).build())
         .set(PROPERTY_NEXT_NATURAL_TRIGGER, instantToTimestamp(triggerSpec.instant()))
-        .set(PROPERTY_NEXT_NATURAL_OFFSET_TRIGGER, instantToTimestamp(triggerSpec.offsetInstant()));
-    tx.put(entity.build());
+        .set(PROPERTY_NEXT_NATURAL_OFFSET_TRIGGER, instantToTimestamp(triggerSpec.offsetInstant()))
+        .build();
 
-    return workflow.id();
+    tx.put(workflowEntity);
+    tx.delete(legacyWorkflowKey);
```
danielnorberg: Only delete if it existed?

danielnorberg:
@honnix I guess we'd also need to first deploy scheduler before deploying api?

honnix:
Yeah. We will need a good plan to roll this out.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorageTransaction.java
honnix: This operation has no impact if not exist as I remember. Will double check.

danielnorberg:
We're seeing a lot of contention errors due to this when people redeploy components with a lot of workflows.

Another less disruptive option could be to add a new api method that allows for the submission of an entire component with multiple workflows. Then Styx could iterate over those workflows and store them in a more datastore-sympathetic manner and avoid contention failures.

honnix:
We may not need this PR after all. `https://cloud.google.com/datastore/docs/firestore-or-datastore` and particularly `Writes to an entity group are no longer limited to 1 per second.`

danielnorberg:
We could ease rolling this out if we change the PR to write both the old and the new workflow key in the same TX. A later PR could stop writing the old workflow keys. Then the migration steps become:

1. Write old & new. Read new and fall back to old.
2. Run migration to copy all old to new.
3. Stop writing and reading old.

* No downtime needed.
* No need to have only one scheduler instance running.
* No need to wait for firestore.

honnix:
> Run migration to copy all old to new.

Does it mean to read all workflows from styx and write them back?

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -320,6 +320,8 @@ public void create(Environment environment) {
     final Config staleStateTtlConfig = config.getConfig(STYX_STALE_STATE_TTL_CONFIG);
     final TimeoutConfig timeoutConfig = TimeoutConfig.createFromConfig(staleStateTtlConfig);
 
+    migrateWorkflows(storage);
```
danielnorberg: Would prefer to run workflow migration as a separate application instead of blocking scheduler startup.

danielnorberg:
This is now done, closing.

honnix:
🔥

PR #487, opened 2018-04-30
check backfill halted status before moving on

honnix:
during triggering a big backfill with high concurrency, it could have
been halted by user. we should check the flag before moving on.

@NatashaL PTAL

It's not trivial to unit test this thing and that why I made the method visible.


NatashaL:
👍

PR #488, opened 2018-04-30
Log errors with service-account key operations

NatashaL:


styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
```diff
@@ -88,7 +88,7 @@ public void tryDeleteKey(String keyName) {
     try {
       deleteKey(keyName);
     } catch (IOException e) {
-      LOG.debug("Ignoring error while deleting key {}", keyName, e);
+      LOG.debug("Ignoring error while deleting key {} {}", keyName, e);
```
honnix: This shouldn't be needed I think. You might did filtering on stackdriver log I guess, but that wouldn't give you the following line (which is the stacktrace in this case).

honnix:
It doesn’t feel needed to adapt our logs for stackdriver.

styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
NatashaL: Ah you're totally right

PR #489, opened 2018-05-04
[wip] remove resource kind

honnix:
The idea is to keep using `CounterLimit` and ditch `Resource` kind in Datastore.
In order to have minimum confusion, `Resource` usage has been removed from storage layer. So at storage side, we call it `CounterLimit`, while at API and scheduler side, we call it `Resource`.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -61,7 +61,6 @@ import com.google.common.hash.Hashing;
 import com.spotify.styx.model.Backfill;
 import com.spotify.styx.model.BackfillBuilder;
 import com.spotify.styx.model.ExecutionDescription;
-import com.spotify.styx.model.Resource;
 import com.spotify.styx.model.Schedule;
 import com.spotify.styx.model.StyxConfig;
 import com.spotify.styx.model.Workflow;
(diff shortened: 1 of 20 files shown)


honnix:
Some of the changes are covered by unit tests in different modules. That's why `codecov/path` is not happy while the overall coverage is increased. :(

honnix:
This wasn't an easy choice due to the naming. If the kind were named as `ResourceShard` if would be a lot easier. Keeping `Resource` will end up with `Resource` + `CounterShard` which to me is quite confusing.

The idea of this PR is to keep counter stuff to storage, while everywhere else we stick to resource.

Pros and cons, I have no strong opinion.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -732,44 +732,24 @@ private static Instant timestampToInstant(Timestamp ts) {
     return Instant.ofEpochSecond(ts.getSeconds(), ts.getNanos());
   }
 
-  Optional<Resource> getResource(String id) {
-    Entity entity = datastore.get(datastore.newKeyFactory().setKind(KIND_RESOURCE).newKey(id));
+  Optional<Tuple2<String, Long>> getCounterLimit(String id) {
```
danielnorberg: Avoid using `Tuple`. Create a bespoke value class (using e.g. auto-value or auto-matter) instead.

honnix: My initial idea was not to introduce another class/concept, but I guess it would be much easier to read the code, also clearer type.

danielnorberg:
Let's get this done.

honnix:
I will bring this PR back. I promise.

honnix:
Too many conflicts due to refactoring a datastore code. Will start from scratch.

PR #490, opened 2018-05-06
align with ShardedCounter for ease of reading

honnix:


styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -487,16 +487,13 @@ void updateShards(final Storage storage,
       final Long usage = entity.getValue();
       LOG.info("Syncing {} -> {}", resource, usage);
       try {
-        int result = (int) (usage / NUM_SHARDS);
```
danielnorberg: Does this code belong in `StyxScheduler.java`? Can it be moved to `ShardedCounter.java` instead?

honnix: That’s also weird. Maybe move to Storage?

honnix: Better address this in #489 if we want to.

honnix: OK to get it merged for the scope of this PR?

NatashaL:
👍

PR #493, opened 2018-05-10
remove redundant wfi reads - speed up polling

danielnorberg:


honnix:
Overall LGTM.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -566,6 +582,8 @@ private void emitPodEvents(Watcher.Action action, Pod pod, RunState runState) {
       }
 
       try {
+        // TODO: spoofing counter values like this can give unexpected results, e.g. if we emit two events here the
+        // first one might be discarded and the second one accepted.
```
honnix: We can continue with https://github.com/spotify/styx/pull/445 if we see high risk here.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -457,58 +458,70 @@ public void init() {
...
-      final Optional<RunState> runState = lookupPodRunState(pod, workflowInstance.get());
-      if (runState.isPresent()) {
-        emitPodEvents(Watcher.Action.MODIFIED, pod, runState.get());
+
+      final RunState runState = activeStates.get(workflowInstance.get());
+      if (runState != null && isPodRunState(pod, runState)) {
+        emitPodEvents(Watcher.Action.MODIFIED, pod, runState);
         cleanupWithRunState(workflowInstance.get(), pod.getMetadata().getName());
       } else {
+        // The pod is stale as we fetched the active states _after_ listing all pods.
```
honnix: There is a small issue here and not so harmful. After fetching all pods, while batch fetching active states, watcher receives pod main container termination event and transitions state (active state gets removed from datastore), so we will not get that active state. We would believe the pod is still running but no state, so we will delete it immediately not respecting graceful deletion period. We could read the pod again here but it would be too slow and not worth it.

danielnorberg: Currently `cleanup()` does read the pod again though. I guess we might want to optimize away that though, in which case this could become a problem.

honnix: That reading has been removed in the sidecar PR I remember.

honnix: nvm, that commits got lost.

PR #497, opened 2018-05-16
fix pod creation / deletion race

danielnorberg:
Looks like this behavior was introduced in https://github.com/spotify/styx/pull/282


danielnorberg:
@honnix wdyt about this, is it a real race? Afaict this behavior has not changed recently, but I noticed it now.

danielnorberg:
Looking at the event trace from a successfully executed pod, there's a window of time where it is still `Pending` with an empty `containerStatuses` before the container creation process has started.

honnix:
I think the old behavior was if not container status don’t do anything.

honnix:
👍

PR #499, opened 2018-05-16
state manager: log verified events

danielnorberg:
It would be useful to be able to scrape logs only for events that did not get discarded due to being stale etc.


NatashaL:
I believe that is being logged here https://github.com/spotify/styx/blob/a4fe912803f8b4bc6142e8bbc7f1821062976e8b/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TransitionLogger.java#L45 which is post transition, ensuring the transition transaction really went through. Granted it's a bit more verbose

NatashaL:
Ah nvm, that log is missing the actual event causing the transition. 

Let's have this too then 👍

danielnorberg:
Yeah would've been nice I guess if the event could be logged there as well but that seems like maybe a bigger refactoring.

danielnorberg:
Maybe the `"{}{} transition -> {} {}"` logging could move to the QSM.

PR #501, opened 2018-05-16
mark scope as provided

honnix:
@danielnorberg PTAL

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -116,11 +116,13 @@
         <groupId>com.google.auto.value</groupId>
         <artifactId>auto-value</artifactId>
         <version>1.5.3</version>
+        <scope>provided</scope>
       </dependency>
       <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -181,6 +183,7 @@
         <groupId>io.norberg</groupId>
         <artifactId>auto-matter-jackson</artifactId>
         <version>0.14.1</version>
+        <scope>provided</scope>
```
danielnorberg: This must not be provided.

honnix: So ok.

danielnorberg:
👍

danielnorberg:
I guess they would. In which case maybe remove the redundant `provided` scope from the submodules?

honnix:
Ah sure. I didn’t even check the sub modules. :D

honnix:
i will close this PR because the children specify the scope already.

PR #502, opened 2018-05-16
(WIP) More informative error messages

ulzha:


ulzha:
About to set similar things straight on the server side as well...

styx-cli/src/test/java/com/spotify/styx/cli/CliMainTest.java
```diff
@@ -526,7 +526,7 @@ public void testClientError() {
       assertThat(e.status(), is(ExitStatus.ClientError));
     }
 
-    verify(cliOutput).printError("Client error: " + exception.getMessage());
+    verify(cliOutput).printError(Throwables.getStackTraceAsString(exception));
```
danielnorberg: Let's not have the cli print the exception stack trace. It is quite unpleasant for users.

ulzha: I quite agree, and that's what I was about to continue on working :+1:

danielnorberg:
close?

PR #504, opened 2018-05-17
state manager: log transaction failures

danielnorberg:
Will be helpful in troubleshooting why instances fail to transition.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
@@ -26,6 +26,7 @@ import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Throwables;
 import com.spotify.styx.model.Event;
 import com.spotify.styx.model.Resource;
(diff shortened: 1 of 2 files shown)


danielnorberg:
@NatashaL PTAL

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -174,12 +190,17 @@ private void initialize(WorkflowInstance workflowInstance) {
...
         throw new RuntimeException(e);
       } else {
+        log.debug("Transaction failure when triggering workflow instance: {}: {}",
+            workflowInstance, e.getMessage(), e);
         throw new RuntimeException(e);
       }
-    } catch (IOException e) {
+    } catch (Exception e) {
+      log.debug("Failure when triggering workflow instance: {}: {}", workflowInstance, e.getMessage(), e);
+      Throwables.propagateIfPossible(e, RuntimeException.class);
```
ulzha: This method is deprecated.

danielnorberg: `Throwables.propagate()` is deprecated, but `Throwables.propagateIfPossible()` is not. However, the newer `Throwables.throwIfUnchecked()` might be preferrable to use.

PR #505, opened 2018-05-17
do not delete pods if the WFI is still RUNNING

danielnorberg:
We want to keep the pod around until the WFI has transitioned out of
RUNNING so we can get the exit code and not falsely label executions as
having failed just because the pod "expired" before the RunState managed
to transition (e.g. due to datastore contention and slowness).


danielnorberg:
This regression seems to have been introduced in https://github.com/spotify/styx/pull/279

honnix:
👍

PR #506, opened 2018-05-17
assign UUIDs to all requests and their logs

danielnorberg:
* For each incoming request, accept or assign a unique ID
* Include the UUID in all log messages logged by the thread(s) handling the request.
* Pass on the request id when proxying from API to Scheduler.
* Return the UUID in the response as the `X-Request-Id` header.
* Make the CLI print the request UUID on API errors - searching for this UUID in our logs will yield all log messages produced when handling the user request.
* Return root cause message in 500 error response


styx-common/pom.xml
```diff
@@ -137,5 +137,10 @@
       <artifactId>JUnitParams</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
```
honnix: let's use `slf4j-simple` instead.

honnix:
How would this work internally where we don't have the logback config?

danielnorberg:
@honnix we will have a logback config internally.

styx-common/pom.xml
danielnorberg: Yeah, this is so that ocular inspection is possible.

danielnorberg:
@ulzha PTAL

ulzha:
I learned a new TLA today, "MDC" \o/

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -126,26 +133,62 @@ private Middlewares() {
...
             }
           } else {
-            return r;
+            response = r;
           }
+          return response.withHeader(X_REQUEST_ID, requestId);
         });
       } catch (ResponseException e) {
-        return completedFuture(e.getResponse());
+        MDC.remove(REQUEST_ID);
```
honnix: Would `finally` work in this case?

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
```diff
@@ -452,7 +457,8 @@ private Request decorateRequest(
             return response;
           default:
             final String message = response.status().code() + " " + response.status().reasonPhrase();
-            throw new ApiErrorException(message, response.status().code(), authToken.isPresent());
+            final String responseRequestId = response.headers().getOrDefault(X_REQUEST_ID, requestId);
```
ulzha: If there's a nonnull ID in headers() different from requestId, then that seems like a "should never happen" kind of situation that would be good to highlight in error log, not just shrug about it.

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -126,26 +133,62 @@ private Middlewares() {
...
+      final String requestIdHeader = requestContext.request().headers().get(X_REQUEST_ID);
+      final String requestId = (requestIdHeader != null)
+          ? requestIdHeader
+          : UUID.randomUUID().toString().replace("-",""); // UUID with no dashes, easier to deal with
+
+      MDC.put(REQUEST_ID, requestId);
+
       try {
         return innerHandler.invoke(requestContext).handle((r, t) -> {
+          MDC.remove(REQUEST_ID);
```
ulzha: Isn't it easier and more robust to remove it once, in `finally`? There's even a try-with-resource example in https://gquintana.github.io/2017/12/01/Structured-logging-with-SL-FJ-and-Logback.html.

styx-common/src/main/java/com/spotify/styx/util/MDCUtil.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+      }
+    };
+  }
+
+  private static Map<String, String> safeCopyOfContextMap() {
+    try {
+      return MDC.getCopyOfContextMap();
+    } catch (Exception e) {
+      log.warn("Failed to copy MDC");
```
ulzha: Should be error level? because this is so unexpected to throw, to my understanding...

ulzha:
Cool, pretty thorough as far as I can tell. A few minor things noted ^.

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
danielnorberg: Should that fail the request or should it just result in an angry log message?

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
danielnorberg: This removal inside the async handler was misplaced. Good catch. 

Didn't know about the try-with-resource helper. Will try to use that instead.

styx-common/src/main/java/com/spotify/styx/util/MDCUtil.java
danielnorberg: Sure.

styx-client/src/main/java/com/spotify/styx/client/StyxApolloClient.java
ulzha: No very strong opinion. A loud fail seems :+1:

PR #509, opened 2018-05-23
fix api -> scheduler request id propagation

danielnorberg:
Make the styx api decorate proxied requests with the request id when the incoming user request did not have a X-Request-Id header and the request id is only accessible through the MDC.

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/SchedulerProxyResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/SchedulerProxyResource.java
@@ -23,6 +23,7 @@ package com.spotify.styx.api;
 import static com.spotify.styx.api.Api.Version.V3;
 
 import com.spotify.apollo.Client;
+import com.spotify.apollo.Request;
 import com.spotify.apollo.RequestContext;
 import com.spotify.apollo.Response;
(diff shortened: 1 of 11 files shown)


danielnorberg:
@ulzha PTAL

styx-api-service/src/main/java/com/spotify/styx/api/SchedulerProxyResource.java
```diff
@@ -73,7 +75,20 @@ public SchedulerProxyResource(String schedulerServiceBaseUrl, Client client) {
...
+  }
+
+  private Request withRequestId(Request request) {
+    if (request.headers().containsKey("X-Request-Id")) {
+      return request;
+    }
+    // Unfortunately it is not possible for middleware to set headers on
+    // incoming requests so we manually propagate it here from the MDC.
+    final String requestId = MDC.get("request-id");
+    if (requestId == null) {
```
ulzha: When is this expected to be null?

danielnorberg: Might be null if the middleware code changes, or a thread pool is introduced without MDC propagation, etc, causing the MDC to not have the request id here. I think it is more important to let the request proxying work instead of failing with a NPE if the request id was unavailable.

PR #512, opened 2018-05-23
Always capture exception causes

ulzha:
Went looking for any occurrences where an exception cause is swallowed (caught and then neither logged nor rethrown), a bit like this:

  $ git grep -A 5 "} catch" | grep -v ", e)" | grep -C 2 -iP "log\."

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
@@ -101,7 +101,7 @@ public class ServiceAccountKeyManager {
     } catch (GoogleJsonResponseException e) {
       // TODO: handle 403 correctly once google fixes their API
       if (e.getStatusCode() == 403 || e.getStatusCode() == 404) {
-        LOG.debug("Couldn't find key to delete {}", keyName);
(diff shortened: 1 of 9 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
```diff
@@ -101,7 +101,7 @@ public void deleteKey(String keyName) throws IOException {
     } catch (GoogleJsonResponseException e) {
       // TODO: handle 403 correctly once google fixes their API
       if (e.getStatusCode() == 403 || e.getStatusCode() == 404) {
-        LOG.debug("Couldn't find key to delete {}", keyName);
+        LOG.debug("Couldn't find key to delete {}", keyName, e);
```
danielnorberg: Don't think the stack trace is needed here, it's not really an exception.

danielnorberg:
👍

styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
ulzha: On other 404 checks I did consider the cause captured without additional logging, but because here there's even a TODO about the 403 situation I went the pedantic route. One bit of information captured :]

PR #513, opened 2018-05-24
hide sensitive header contents only

danielnorberg:
Instead of hiding the presence of the header in the request. This is useful for debugging.

Also adds some tests to make codecov happy.


ulzha:
:+1:

PR #516, opened 2018-05-24
log workflow read exception

danielnorberg:
Let's log this exception as well. There's nothing else logging it upstream when hit by a user request.

@ulzha ptal

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -663,7 +663,7 @@ public class DatastoreStorage implements Closeable {
       return OBJECT_MAPPER
           .readValue(entity.getString(PROPERTY_WORKFLOW_JSON), Workflow.class);
     } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


ulzha:
:+1:

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -663,7 +663,7 @@ static Workflow parseWorkflowJson(Entity entity, WorkflowId workflowId) throws I
       return OBJECT_MAPPER
           .readValue(entity.getString(PROPERTY_WORKFLOW_JSON), Workflow.class);
     } catch (IOException e) {
-      LOG.error("Failed to read workflow for {}, {}", workflowId.componentId(), workflowId.id());
+      LOG.error("Failed to read workflow for {}, {}", workflowId.componentId(), workflowId.id(), e);
       throw e;
```
ulzha: Interesting, I like to rely on such rethrows ^ to mean that we're promising ourselves to handle the exception with due thoroughness upon catching upstream.

danielnorberg:
Yeah maybe the exception handler middleware should log all exceptions?
That’s where it ends up.

PR #521, opened 2018-05-31
revert container naming: use "styx-run"

danielnorberg:
Unique container naming causes logs to end up in separate stackdriver logs and breaks log exporting.


Zatine:
Looks good to me! 👍

PR #523, opened 2018-06-02
read global resource limit from StyxConfig

honnix:
May not be the best approach. But let's kick off the discussion.


honnix:
@danielnorberg @NatashaL fyi

danielnorberg:
Instead of propagating the magic of the global resource and have more special cases in our code, I'd rather we use the regular resource commands to control it.

And to prevent that users change it (or other important resource limits) using `styx resource edit`, we should implement authorization for resource changes.

But 👍 for now I guess. 😢

honnix:
@danielnorberg feel free to ship it.

PR #520, opened 2018-05-30
store workflow in transaction

danielnorberg:
Avoid potential corruption, inconsistencies or partial updates due to races, failures, restarts, etc.

PR #525, opened 2018-06-04
Revert "store workflow in transaction"

honnix:
Reverts spotify/styx#520

This doesn't work unfortunately. We cannot read workflow from transaction after `store` it so `updateNextNaturalTrigger` failed.

We got `com.spotify.styx.util.ResourceNotFoundException`.

@NatashaL PATL


NatashaL:
👍

honnix:
cc @danielnorberg we need a bit of rework.

PR #527, opened 2018-06-07
Datastore metrics

danielnorberg:


styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
```diff
@@ -537,8 +540,8 @@ public void shouldDeleteWorkflow() throws Exception {
   @Test
   public void shouldReturnErrorWhenDeleteNonexistWorkflow() throws Exception {
     sinceVersion(Api.Version.V3);
-    
-    when(storage.workflow(WORKFLOW.id())).thenReturn(Optional.empty());
+
+    doReturn(Optional.empty()).when(storage).workflow(WORKFLOW.id());
```
Zatine: Should this be `storage.workflow(WORKFLOW.id())` (as before) instead of `(storage).workflow(WORKFLOW.id()`?

Zatine: Saw it in another place as well and looked up the syntax and it seems correct, so nevermind.

danielnorberg: Yeah the problem here is that `storage` is a "spy", and spies sometimes don't work with the `when(...).then(...)` syntax. In this case that syntax broke when I added the `State` parameter to the `AggregateStorage` class constructor.

Zatine: 👍

honnix:
A more general comment. I think we should have operation rate metrics in stackdriger.

danielnorberg:
@honnix Unfortunately the stackdriver metrics seem too coarse to be very useful. For instance, they do not give numbers per entity `kind` and it's not possible to group by request origin (i.e. styx `api` vs `scheduler`). I want datastore metrics with more granular detail and I consider this PR to be just the beginning. For now it produces read metrics tagged by `kind`, operation type and `role`. Future iterations could be to also tag by context, i.e. the api request uri, or scheduler tick, etc.

danielnorberg:
@honnix poke. Would really like to get this moving so we can start understanding our datastore usage and figure out how to reduce costs.

honnix:
Will take a look today. Didn't get time. It's still in my to-do list. ;)

styx-common/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -377,4 +397,9 @@ private Histogram tickHistogram(String type) {
     return tickHistograms.computeIfAbsent(
         type, (op) -> registry.histogram(TICK_DURATION.tagged("type", type)));
   }
+
+  private Meter datastoreOperationMeter(String operation, String kind) {
+    return datastoreOperationMeters.computeIfAbsent(Tuple.of(operation, kind),
+        t -> registry.meter(DATASTORE_OPERATION_RATE.tagged("operation", operation, "kind", kind)));
```
honnix: align the naming convertion? s/t/(op)/

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -417,13 +443,17 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
         .map(entity -> entity.getKey().getName())
         .map(name -> activeWorkflowInstanceKey(datastore.newKeyFactory(), name))
         .collect(toList());
+    stats.recordDatastoreEntityReads(KIND_ACTIVE_WORKFLOW_INSTANCE_INDEX_SHARD_ENTRY, keys.size());
 
     // Strongly consistently read values for the above keys
-    return Lists.partition(keys, MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH_READ).stream()
+    final Map<WorkflowInstance, RunState> states = Lists.partition(keys, MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH_READ)
```
honnix: why the change here?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -417,13 +443,17 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
...
+        .stream()
         .map(batch -> forkJoinPool.submit(withMDC(() -> this.readRunStateBatch(batch))))
         .collect(toList()).stream() // collect here to execute batch reads in parallel
         .flatMap(task -> task.join().stream())
         .collect(toMap(RunState::workflowInstance, Function.identity()));
+    return states;
+
```
honnix: empty line.

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
```diff
@@ -374,7 +377,8 @@ public void shouldFailDecrementingEmptyShard() throws IOException {
 
   @Test(expected = ShardNotFoundException.class)
   public void shouldThrowExceptionOnUninitializedShards() {
-    when(counterSnapshotFactory.create(COUNTER_ID1)).thenReturn(new ShardedCounter.Snapshot(COUNTER_ID1, 100, new HashMap<>()));
+    doReturn(new ShardedCounter.Snapshot(COUNTER_ID1, 100, new HashMap<>()))
```
honnix: Is this change necessary?

styx-common/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
danielnorberg: But it's a tuple of operation and kind. And it's not used by the lambda. I would use `_` if I could but that's not allowed anymore/yet.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: This change is no longer needed, will revert.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: Will remove

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
honnix: Sure but it is not storage here but that counterSnapshotFactory.

styx-common/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
honnix: Ok ignore this comment.

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
danielnorberg: Yes but counterSnapshotFactory is using the spy storage.

honnix: OK then. I missed that connection.

danielnorberg: Sorry I meant that `counterSnapshotFactory` is a spy and that breaks `when(...).then(...)`.

honnix:
👍

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
honnix: Ah, so it never worked before. That's interesting.

honnix:
I see #529 is a revamp of this? And that is still WIP?

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
danielnorberg: It has worked before (by accident, luck?), but it broke in this PR because now we inject a mock Stats into AggregateStorage. And then in the `when(...)` the _real_ `counterSnapshotFactory.create` was called, which it the mocked `Stats`, which broke Mockito. And that's the gist of the problem. When doing `when(...)` on a spy, the _real_ method gets called immediately then and there, and we don't want that, that breaks things.

danielnorberg:
@honnix I'd be happy to go with #529 instead. It's a bigger PR though because it instruments the whole Datastore interface.

styx-common/src/test/java/com/spotify/styx/util/ShardedCounterTest.java
honnix: Yeah. It was never the right way working with `spy`, and fortunately this change surfaced it.

honnix:
Now the state is lying: changes approved. :D

We will need to review it again.

styx-common/src/main/java/com/spotify/styx/storage/InstrumentedDatastoreBatchWriter.java
```diff
@@ -0,0 +1,124 @@
...
+import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.FullEntity;
+import com.google.cloud.datastore.Key;
+import com.spotify.styx.monitoring.Stats;
+import java.util.List;
+import java.util.Objects;
+
+interface InstrumentedDatastoreBatchWriter
+    // XXX: DatastoreBatchWriter is not a public interface, Batch is an almost identical subtype
+    // https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3387
```
ulzha: :star:

ulzha:
:+1: Seems fine.

I guess this will overcount in cases when transactions get rolled back yet meters have already been marked?

Happy that the problems in tests got surfaced.

danielnorberg:
@ulzha Yes, this will overcount writes in transactions that are never committed. As for transactions that roll back due to failure to commit, I think it might be correct to count those. I believe this to not be a problem atm.

danielnorberg:
@ulzha I guess an alternative here to reduce size would be to introduce yet another level of indirection and have our own smaller interface in front of the full `Datastore` interface and instrument that instead.

PR #530, opened 2018-06-15
tracing of requests & internals

danielnorberg:
add tracing to styx api/scheduler requests and internals

A next step after this PR is to instrument the storage methods etc with spans to get more granular breakdown of what's going on in there.

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
+++ b/styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
@@ -22,6 +22,8 @@ package com.spotify.styx.api;
 
 import static com.spotify.apollo.Status.INTERNAL_SERVER_ERROR;
 import static com.spotify.styx.serialization.Json.OBJECT_MAPPER;
+import static io.opencensus.trace.AttributeValue.stringAttributeValue;
+import static io.opencensus.trace.Status.UNKNOWN;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
@@ -43,13 +45,17 @@ import com.spotify.apollo.route.Middleware;
 import com.spotify.apollo.route.SyncHandler;
 import com.spotify.styx.util.MDCUtil;
 import io.norberg.automatter.AutoMatter;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.Tracer;
 import java.io.IOException;
+import java.net.URI;
 import java.security.GeneralSecurityException;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.CompletionStage;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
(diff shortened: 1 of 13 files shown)


styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
```diff
@@ -220,6 +226,33 @@ static GoogleIdToken verifyIdToken(String s, GoogleIdTokenVerifier verifier) {
...
+        span.setStatus(UNKNOWN);
+        span.end();
+        Throwables.throwIfUnchecked(e);
+        throw new RuntimeException(e);
+      }
+      response.whenComplete((r, ex) -> {
+        if (ex != null) {
+          span.setStatus(UNKNOWN);
+        }
+        span.end();
```
ulzha: Doing `end()` in multiple places. Consider try-with-resource instead? https://static.javadoc.io/io.opencensus/opencensus-api/0.13.2/io/opencensus/trace/SpanBuilder.html#startScopedSpan--

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -420,7 +425,7 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
 
     // Strongly consistently read values for the above keys
     return Lists.partition(keys, MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH_READ).stream()
-        .map(batch -> forkJoinPool.submit(withMDC(() -> this.readRunStateBatch(batch))))
+        .map(batch -> CompletableFuture.supplyAsync(() -> this.readRunStateBatch(batch)))
```
ulzha: `executor` not passed, by accident?

styx-common/src/main/java/com/spotify/styx/api/Middlewares.java
danielnorberg: The request handler might be async, in which case it returns a future and we'd end up closing the span before it's (asynchronously) completed.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: good catch!

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -420,7 +425,7 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
 
     // Strongly consistently read values for the above keys
     return Lists.partition(keys, MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH_READ).stream()
-        .map(batch -> forkJoinPool.submit(withMDC(() -> this.readRunStateBatch(batch))))
+        .map(batch -> CompletableFuture.supplyAsync(() -> this.readRunStateBatch(batch), forkJoinPool))
```
ulzha: Erm I think you meant `executor`, like in the other replaced usages of `withMDC`.

danielnorberg: Good catch again!

PR #533, opened 2018-07-11
Scheduler tracing

danielnorberg:


NatashaL:
@danielnorberg How did you choose which codepaths to trace, or what information exactly are you looking to find with this tracing? so I now what to look for myself

danielnorberg:
@NatashaL I was primarily looking to trace the `Scheduler.tick()`.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -400,7 +404,7 @@ public void create(Environment environment) {
     startRuntimeConfigUpdate(styxConfig, tickExecutor, dequeueRateLimiter);
     startCleaner(cleaner, tickExecutor);
 
-    setupMetrics(stateManager, workflowCache, storage, dequeueRateLimiter, stats);
+    setupMetrics(queuedStateManager, workflowCache, storage, dequeueRateLimiter, stats);
```
NatashaL: With this we'll miss tracing of calls made to QueuedStateManager for getting metrics. Is that intended? Can we send the proxy instead?

danielnorberg: The `setupMetrics` method needs to access the `QueuedStateManager` directly, it cannot use the `StateManager` interface.

NatashaL: 👍 sounds like it should be refactored. but that can go in another PR then

NatashaL:
👍 to merge after tests pass

PR #534, opened 2018-07-11
retry resource limited instances later

danielnorberg:
Avoid spinning and exhausting dequeue rate limit on instances that are blocked on exhausted resources.

Note: Even if a workflow instance is rescheduled for later dequeue, the Scheduler will keep reading it in every tick.

Note: This will cause execution of workflow instances that were blocked on resources to be delayed and not immediately start when the resource becomes available.


danielnorberg:
Closing this for now. Might revisit if we also come up with a more efficient queueing / scheduled dequeueing mechanism.

PR #538, opened 2018-07-13
scheduler: check resource limit before dequeue

danielnorberg:
An alternative to https://github.com/spotify/styx/pull/534.

By doing a best-effort resource limit check before dequeue, we can avoid exponential backoff delay and also avoid consuming the dequeue rate limit.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -71,7 +71,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -45,6 +45,7 @@ import com.spotify.styx.model.WorkflowId;
 import com.spotify.styx.model.WorkflowInstance;
 import com.spotify.styx.monitoring.Stats;
 import com.spotify.styx.state.InstanceState;
+import com.spotify.styx.state.Message;
 import com.spotify.styx.state.RunState;
 import com.spotify.styx.state.RunState.State;
 import com.spotify.styx.state.StateManager;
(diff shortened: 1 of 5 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -317,12 +319,15 @@ public void create(Environment environment) {
     closer.register(publisher);
 
     final ScheduledExecutorService executor = executorFactory.create(3, schedulerTf);
-    closer.register(executorCloser("scheduler", executor));
+    closer.register(executorCloser("scheduled-executor", executor));
```
ulzha: Putting on my uninitiated reader hat again: now we have both executor and schedulerExecutor in Scheduler, pretty confusing to read. And no clue which one of those is controlled by "styx.scheduler-threads" configuration parameter. And not sure why one of them uses newWorkStealingPool and the other uses some different kind of factory. (Would be great to add meaningful distinctions between the names, and meaningful connection with the configuration parameter name.)

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -277,15 +259,64 @@ private void dequeueInstance(Map<String, Resource> resources,
...
+    }
+
+    // Racy: some resources may have been removed (become unknown) by now; in that case the
+    // counters code during dequeue will treat them as unlimited...
+    sendDequeue(instanceState, instanceResourceRefs);
+  }
+
+  private boolean limitReached(final String resourceId) {
+    try {
+      final CounterSnapshot counterSnapshot = shardedCounter.getCounterSnapshot(resourceId);
```
ulzha: The CounterSnapshot was meant to be private as an implementation detail, made public to ease testing at some point, and so it goes :)

danielnorberg: I guess we could expose a nicer api for doing this check instead of interacting directly with the `CounterSnapshot`.

styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
```diff
@@ -110,12 +112,16 @@
...
+  @Mock ShardedCounter shardedCounter;
+  @Mock CounterSnapshot counterSnapshot;
 
   @Before
   public void setUp() {
     workflows = new HashMap<>();
     when(gate.executionBlocker(any()))
         .thenReturn(WorkflowExecutionGate.NO_BLOCKER);
+    when(shardedCounter.getCounterSnapshot(anyString())).thenReturn(counterSnapshot);
+    when(counterSnapshot.pickShardWithSpareCapacity(anyLong())).thenReturn(0);
```
ulzha: Might have been appropriate to refactor this test too in accordance with `counterHasSpareCapacity`.

PR #541, opened 2018-08-07
styx environment and logging format

honnix:
@danielnorberg @ulzha PTAL

As discussed, it would be better to have a dedicated env.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -147,6 +147,8 @@ public class StyxScheduler implements AppInit {
   public static final String STYX_SCHEDULER_TICK_INTERVAL = "styx.scheduler.tick-interval";
   public static final String STYX_TRIGGER_TICK_INTERVAL = "styx.trigger.tick-interval";
   public static final String STYX_SCHEDULER_THREADS = "styx.scheduler-threads";
+  private static final String STYX_ENVIRONMENT = "styx.environment";
+  private static final String STYX_ENVIRONMENT_PRODUCTION = "production";
 
   public static final int DEFAULT_STYX_EVENT_PROCESSING_THREADS = 32;
   public static final int DEFAULT_STYX_SCHEDULER_THREADS = 32;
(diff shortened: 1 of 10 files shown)


danielnorberg:
I think I want the opposite. Add a env var like `STYX_LOGGING=text` to quick-styx instead and look for that in our loggers.

honnix:
This is not only for logging. We have use cases that users want to detect whether running on Styx. So we can give them a dedicated variable.

ulzha:
Yeah I also think explicit STYX_LOGGING will be more obvious to work with down the line. I don't see a big benefit of immediately tying the choice of logging format to the fact of running in production Styx; the users may well want to be able to get per-task logging locally/in qstyx, and they should not have to specify a lying environment variable about running in production Styx to achieve that.

It does make sense to provide the injection of "running on Styx" information, separately. In general I'd say that a boolean toggle screams of a missed opportunity for extensibility; how about STYX_ENVIRONMENT that can be one of {production,testing,qstyx,...}?

danielnorberg:
👍 for `STYX_LOGGING` and `STYX_ENVIRONMENT`

honnix:
@danielnorberg @ulzha PTAL again.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -681,11 +683,17 @@ private static DockerRunner createDockerRunner(
       LOG.info("Creating LocalDockerRunner");
       return closer.register(DockerRunner.local(scheduler, stateManager));
     } else {
+      final String styxEnvironment;
+      if (!config.hasPath(STYX_ENVIRONMENT)) {
+        styxEnvironment = STYX_ENVIRONMENT_PRODUCTION;
```
danielnorberg: is production a good/safe default?

honnix: No idea. ;) Maybe `development` or `NA`?

NatashaL:
@honnix is there more to be done here? can we merge? @ulzha any more feedback from you?

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -681,11 +683,17 @@ private static DockerRunner createDockerRunner(
       LOG.info("Creating LocalDockerRunner");
       return closer.register(DockerRunner.local(scheduler, stateManager));
     } else {
+      final String styxEnvironment;
+      if (!config.hasPath(STYX_ENVIRONMENT)) {
```
danielnorberg: Would it be better perhaps to not have a default environment? I.e. fail if `styx.environment` is not set.

danielnorberg: Afact this is how we handle `styx.mode`.

honnix: Is it? I thought we just created different classes depending on `styx.mode`.

honnix: Ah ok it fails by getting non-configured value.

ulzha:
:+1:

danielnorberg:
:+1:

PR #543, opened 2018-08-08
revert addition of `commitSha` to `created`

danielnorberg:
There are no existing `created` records with `commitSha` and the event is no longer emitted.


danielnorberg:
@NatashaL @juliaromare PTAL

juliaromare:
Looks good

PR #544, opened 2018-08-09
Add method to client to get workflow's instance

AlejandroVera:
This is the equivalent to this API endpoint: https://spotify.github.io/styx/api.html#workflows-workflow-instance-get

### DIFF
--- a/styx-client/src/main/java/com/spotify/styx/client/StyxWorkflowClient.java
+++ b/styx-client/src/main/java/com/spotify/styx/client/StyxWorkflowClient.java
@@ -23,6 +23,7 @@ package com.spotify.styx.client;
 import com.spotify.styx.model.Workflow;
 import com.spotify.styx.model.WorkflowConfiguration;
 import com.spotify.styx.model.WorkflowState;
+import com.spotify.styx.model.data.WorkflowInstanceExecutionData;
 import java.util.List;
 import java.util.concurrent.CompletionStage;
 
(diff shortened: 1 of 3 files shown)


styx-client/src/main/java/com/spotify/styx/client/StyxWorkflowClient.java
```diff
@@ -74,6 +75,18 @@
...
 
+  /**
+   * Get execution data of an instance of a {@link Workflow}
+   *
+   * @param componentId component id
+   * @param workflowId  workflow id
+   * @param parameter   parameter
+   * @return the {@link WorkflowInstanceExecutionData}
+   */
+  CompletionStage<WorkflowInstanceExecutionData> workflowInstance(final String componentId,
```
danielnorberg: Should this be called `workflowInstanceExecutions` instead?

AlejandroVera: Yeah, I wasn't too sure about that. I end up with this name because in the API (https://spotify.github.io/styx/api.html#workflows-workflow-instance-get) says "Get Workflow Instance" in the description of the endpoint... :$

AlejandroVera: Renamed @danielnorberg !

NatashaL:
LGTM 👍

PR #546, opened 2018-08-15
Backfill status performance improvement

honnix:
* read active states by backfill trigger id
* skip replaying of running workflow instances
* parallel stream -> fork join pool

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
@@ -61,6 +61,7 @@ import com.spotify.styx.util.ReplayEvents;
 import com.spotify.styx.util.ResourceNotFoundException;
 import com.spotify.styx.util.TimeUtil;
 import com.spotify.styx.util.WorkflowValidator;
+import java.io.Closeable;
 import java.io.IOException;
 import java.time.Instant;
 import java.util.Collection;
@@ -72,26 +73,33 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ForkJoinTask;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import okio.ByteString;
 
-public final class BackfillResource {
+public final class BackfillResource implements Closeable {
 
   static final String BASE = "/backfills";
   private static final String SCHEDULER_BASE_PATH = "/api/v0";
   private static final String UNKNOWN = "UNKNOWN";
   private static final String WAITING = "WAITING";
+  private static final int CONCURRENCY = 64;
 
   private final Storage storage;
   private final String schedulerServiceBaseUrl;
   private final WorkflowValidator workflowValidator;
+
+  private final ForkJoinPool forkJoinPool;
 
   public BackfillResource(String schedulerServiceBaseUrl, Storage storage,
       WorkflowValidator workflowValidator) {
     this.schedulerServiceBaseUrl = Objects.requireNonNull(schedulerServiceBaseUrl);
     this.storage = Objects.requireNonNull(storage);
     this.workflowValidator = Objects.requireNonNull(workflowValidator, "workflowValidator");
+    this.forkJoinPool = new ForkJoinPool(CONCURRENCY);
   }
 
   public Stream<Route<AsyncHandler<Response<ByteString>>>> routes() {
@@ -131,7 +139,17 @@ public final class BackfillResource {
     );
   }
 
-  public BackfillsPayload getBackfills(RequestContext rc) {
+  @Override
+  public void close() {
+    forkJoinPool.shutdownNow();
+    try {
+      forkJoinPool.awaitTermination(30, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+    }
+  }
+
+  private BackfillsPayload getBackfills(RequestContext rc) {
     final Optional<String> componentOpt = rc.request().parameter("component");
     final Optional<String> workflowOpt = rc.request().parameter("workflow");
     final boolean includeStatuses = rc.request().parameter("status").orElse("false").equals("true");
@@ -155,18 +173,21 @@ public final class BackfillResource {
       throw new RuntimeException(e);
     }
 
-    final List<BackfillPayload> backfillPayloads = backfills.parallel()
-        .map(backfill -> BackfillPayload.create(
+    final List<BackfillPayload> backfillPayloads = backfills
+        .map(backfill -> forkJoinPool.submit(() -> BackfillPayload.create(
             backfill,
             includeStatuses
             ? Optional.of(RunStateDataPayload.create(retrieveBackfillStatuses(backfill)))
-            : Optional.empty()))
+            : Optional.empty())))
+        .collect(toList())
+        .stream()
+        .map(ForkJoinTask::join)
         .collect(toList());
 
     return BackfillsPayload.create(backfillPayloads);
   }
 
-  public Response<BackfillPayload> getBackfill(RequestContext rc, String id) {
+  private Response<BackfillPayload> getBackfill(RequestContext rc, String id) {
     final boolean includeStatuses = rc.request().parameter("status").orElse("true").equals("true");
     final Optional<Backfill> backfillOpt = storage.backfill(id);
     if (!backfillOpt.isPresent()) {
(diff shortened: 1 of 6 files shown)


styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -350,7 +347,13 @@ public void shouldGetBackfillStatusWithStatusByDefault() throws Exception {
     storage.writeEvent(SequenceEvent.create(Event.submitted(wfi, "exec-1"),                              4L, 4L));
     storage.writeEvent(SequenceEvent.create(Event.started(wfi),                                          5L, 5L));
     storage.writeActiveState(wfi, RunState.create(wfi, State.RUNNING,
-        StateData.zero(), Instant.now(), 5L));
+        StateData.newBuilder()
```
honnix: to be reviewed.

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -376,7 +379,13 @@ public void shouldGetBackfillStatusWithStatusByDefaultReversed() throws Exceptio
...
```
honnix: to be reviewed.

honnix:
Messed up by code formatter, sorry about this. I commented the actual changes in the test code.

honnix:
We have verified these changes improved quite a lot when running locally. That's why we would like to ship them to production and verify there.

honnix:
@danielnorberg wdyt?

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -72,26 +73,33 @@
...
   private final WorkflowValidator workflowValidator;
 
+  private final ForkJoinPool forkJoinPool;
+
   public BackfillResource(String schedulerServiceBaseUrl, Storage storage,
       WorkflowValidator workflowValidator) {
     this.schedulerServiceBaseUrl = Objects.requireNonNull(schedulerServiceBaseUrl);
     this.storage = Objects.requireNonNull(storage);
     this.workflowValidator = Objects.requireNonNull(workflowValidator, "workflowValidator");
+    this.forkJoinPool = new ForkJoinPool(CONCURRENCY);
```
danielnorberg: We're starting to have a lot of bespoke thread pools with throughout the code base, each with life cycle management etc. Maybe we should start injecting a shared FJP from `StyxScheduler` instead and centralize setup and cleanup?

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -245,7 +245,7 @@ public void shouldFilterBackfillsOnComponentEvenWhenInactive() throws Exception
     storage.storeBackfill(BACKFILL_4.builder().allTriggered(true).build());
 
     final String uri = path(String.format("?showAll=true&component=%s",
-                                          BACKFILL_4.workflowId().componentId()));
+        BACKFILL_4.workflowId().componentId()));
```
danielnorberg: Please avoid unnecessary whitespace changes in otherwise unchanged code. Makes the PR noisier and harder to review.

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -350,7 +350,13 @@ public void shouldGetBackfillStatusWithStatusByDefault() throws Exception {
     storage.writeEvent(SequenceEvent.create(Event.submitted(wfi, "exec-1"),                              4L, 4L));
     storage.writeEvent(SequenceEvent.create(Event.started(wfi),                                          5L, 5L));
     storage.writeActiveState(wfi, RunState.create(wfi, State.RUNNING,
-        StateData.zero(), Instant.now(), 5L));
+        StateData.newBuilder()
```
danielnorberg: How come this needs to be changed?

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
honnix: Makes sense. We might actually need multiple FJPs for different components to get better throughput.

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
```diff
@@ -376,7 +382,13 @@ public void shouldGetBackfillStatusWithStatusByDefaultReversed() throws Exceptio
...
```
danielnorberg: How come this needs to be changed?

styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
```diff
@@ -59,10 +57,7 @@ private ReplayEvents() {
 
     RunState restoredState = RunState.fresh(workflowInstance, time);
 
-    final long lastConsumedEvent =
-        Optional.ofNullable(activeWorkflowInstances.get(workflowInstance))
-            .map(RunState::counter)
-            .orElse(sequenceEvents.last().counter());
+    final long lastConsumedEvent = sequenceEvents.last().counter();
```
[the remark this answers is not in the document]
honnix: Good catch. Will fix.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -365,9 +386,9 @@ private static boolean isActiveState(RunStateData runStateData) {
     final List<RunStateData> processedStates;
     final List<RunStateData> waitingStates;
 
-    Map<WorkflowInstance, RunState> activeWorkflowInstances;
+    final Map<WorkflowInstance, RunState> activeWorkflowInstances;
     try {
-      activeWorkflowInstances = storage.readActiveStates();
+      activeWorkflowInstances = storage.readActiveStatesByTriggerId(backfill.id());
```
danielnorberg: Maybe note here that it is ok for this to be weakly consistent?

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
danielnorberg: Yeah, so let's have the test verify that they are not used.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -417,4 +428,24 @@ private static boolean isActiveState(RunStateData runStateData) {
         : Stream.concat(processedStates.stream(), waitingStates.stream()).collect(toList());
   }
 
+  private RunStateData getRunStateData(Backfill backfill,
+      Map<WorkflowInstance, RunState> activeWorkflowInstances, Instant instant) {
+    final WorkflowInstance wfi = WorkflowInstance
+        .create(backfill.workflowId(), toParameter(backfill.schedule(), instant));
+    if (activeWorkflowInstances.containsKey(wfi)) {
+      RunState state = activeWorkflowInstances.get(wfi);
```
danielnorberg: The active state here might not be triggered by the backfill.

It might for example be a naturally triggered active state that is still running and that will be be re-triggered later by the backfill.

It could also be an instance that was ad-hoc re-triggered _after_ the backfill completed running. It could also be a an instance triggered by another backfill.

So for example, for a backfill that has already finished, we might return statuses that indicates that the backfill hasn't finished running yet. That seems incorrect.

danielnorberg: We seem to correctly check for the backfill trigger id in the replay code: https://github.com/spotify/styx/pull/546/files#diff-6762e6046164a0a7533c9d5d13973226R83

danielnorberg:
If a backfill is created and some of the instances are not triggered because they are currently active, what does that mean for the status of the backfill? Like (I belive) @ulzha mentioned we should perhaps handle triggering of already active instances in a better way. E.g. by queueing that re-trigger to happen after the current execution completes.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
honnix: This map should only contain states belonging to this backfill. Or?

danielnorberg: Indeed. Please add a test to verify this behavior. I.e. store an active state that is not triggered by the backfill we're querying for and verify that it is not used and that the events are replayed instead.

honnix: Ok I think I understand.

honnix:
I don't think we allow that to happen. Iirc, API would return error indicating not possible to create backfill because some of the partition is active.

honnix:
@danielnorberg PTAL again.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
honnix: Addressed in later commit.

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
honnix: Addressed in later commit.

styx-api-service/src/test/java/com/spotify/styx/api/BackfillResourceTest.java
honnix: Addressed in later commit.

ulzha:
:+1:

PR #547, opened 2018-08-17
[poc] cache backfill status

honnix:
exploring possible ways of further optimisation, `space for time` is one of those =D
@danielnorberg @Zatine


danielnorberg:
Can we do something that will work for backfills that are in progress as well?

E.g., keep per-backfill execution statistics in storage, so that querying for the aggregate status of a backfill becomes cheap?

honnix:
Yeah that we can do, though that would become a different API. This API was designed to return all details of the backfill unfortunately/accidentally.

honnix:
captured the idea internally, so close for now.

PR #549, opened 2018-08-20
scheduler: log tick information

danielnorberg:


styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -185,6 +185,11 @@ private void tick0() {
             .sorted(comparingLong(i -> i.runState().timestamp()))
             .collect(toList());
 
+    final String message = String.format("Instances: active=%d, eligible=%d, timedOut=%d",
+        activeStates.size(), eligibleInstances.size(), timedOutInstances.size());
+    LOG.info(message);
+    tracer.getCurrentSpan().addAnnotation(message);
```
honnix: Shall we verify this invocation in test case?

honnix:
minor comment, otherwise LGTM.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
danielnorberg: It's not possible to mock `tracer.getCurrentSpan()` 😫

honnix: nvm, i thought that `tracer` was injected.

honnix:
👍

PR #551, opened 2018-08-20
wip: more efficient scheduling queue

danielnorberg:
* Have `cold` and `hot` queues
* `cold` queue is for instances scheduled "far" into the future.
* `hot` queue is for instances that are "soon" due for execution.
* Every N minutes, scan over the cold queue and move instances due for execution within N minutes to the `hot` queue.
* The frequent scheduler tick only looks at the `hot` queue.

Goal: Reduce the amount of datastore reads in `Scheduler.tick`.


danielnorberg:
@honnix wdyt about doing something like this?

PR #552, opened 2018-08-20
stats: cache active states

danielnorberg:
I think this might be the source of our excessive datastore reads.

### DIFF
--- a/styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerTest.java
+++ b/styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerTest.java
@@ -20,6 +20,7 @@
 
 package com.spotify.styx;
 
+import static com.spotify.styx.model.Schedule.DAYS;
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerTest.java
```diff
@@ -172,6 +189,29 @@ public void shouldFailToResetShardsOfResource() throws IOException {
     }
   }
 
+  @Test
+  public void testSetupMetrics() {
```
honnix: 🌟

honnix:
Ignore codecov in this case.

👍 ship it

danielnorberg:
Got 100% diff coverage, let's not give up too easily, we can do it with a little bit of effort 😄 

And now the test actually does some useful work, although it can be improved. It verifies that all the gauges run without throwing exceptions and that expensive method calls are cached.

PR #554, opened 2018-08-21
k8s: optimize watcher

danielnorberg:
* Make pod watcher asynchronous to avoid falling behind and getting disconnected.
* Let events for a pod coalesce in order to avoid re-processing for each small incremental pod update event.
* Process pod update event batches on the same thread as the poller, eliminate race conditions.
* Parallelize processing of pod update events to defeat IO latency.

### DIFF
--- a/styx-common/src/test/java/com/spotify/styx/util/GrpcContextUtilTest.java
+++ b/styx-common/src/test/java/com/spotify/styx/util/GrpcContextUtilTest.java
@@ -1,0 +1,166 @@
+/*
+ * -\-\-
+ * Spotify Styx Scheduler Service
+ * --
+ * Copyright (C) 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.styx.util;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.verify;
+
+import com.google.common.collect.ImmutableList;
+import io.grpc.Context;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class GrpcContextUtilTest {
+
+  private static final Context.Key<String> TEST_KEY = Context.key("test");
+
+  @Rule public ExpectedException exception = ExpectedException.none();
+
+  @Mock ExecutorService executorService;
+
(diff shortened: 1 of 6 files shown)


danielnorberg:
Seems the `codecov/patch` check doesn't take into account line number changes. I.e., if the line number of a line without coverage changes to a line number of a line that had coverage due to lines being added, then codecov flags that line as having regressed from being covered to not being covered =/

danielnorberg:
Actually seems like the newly added `GrpcContextUtilTest` wasn't run by circle-ci, causing the coverage to fail. Looking into why.

danielnorberg:
Jacoco output seems to be missing some coverage: https://4747-74431644-gh.circle-artifacts.com/0/jacoco-aggregate/styx-common/com.spotify.styx.util/GrpcContextUtil.java.html#L45

danielnorberg:
My bad. Seems I configured jacoco to exclude nested classes a year and a half ago 😫

styx-common/src/test/java/com/spotify/styx/util/GrpcContextUtilTest.java
```diff
@@ -0,0 +1,166 @@
...
+  public void shouldPropagateContextForCfSupplyAsync() throws Exception {
+    final Supplier<String> get = TEST_KEY::get;
+    final CompletableFuture<String> cf = Context.current().withValue(TEST_KEY, "foobar")
+        .call(() -> CompletableFuture.supplyAsync(get, sut));
+    assertThat(cf.get(30, SECONDS), is("foobar"));
+  }
+
+  @Test
+  public void shouldNotBeConstructable() throws ReflectiveOperationException {
+    final Constructor<GrpcContextUtil> constructor = GrpcContextUtil.class.getDeclaredConstructor();
```
honnix: Aiming for 100% coverage. :)

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -507,8 +518,10 @@ public void init() {
         pollPodsIntervalSeconds,
         TimeUnit.SECONDS);
 
-    watch = client.pods()
-        .watch(new PodWatcher());
+    final PodWatcher watcher = new PodWatcher();
+    executor.scheduleWithFixedDelay(guard(watcher::processPodUpdates), 0, 5, TimeUnit.SECONDS);
```
honnix: Let's give this magic number a static field.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -674,17 +687,63 @@ private String readStatus(Pod pod) {
 
     private static final int RECONNECT_DELAY_SECONDS = 1;
 
+    private final ConcurrentMap<String, WorkflowInstance> podUpdates = new ConcurrentHashMap<>();
+
     @Override
     public void eventReceived(Action action, Pod pod) {
+
```
honnix: Make this javadoc instead.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -674,17 +687,63 @@ private String readStatus(Pod pod) {
...
+            // Remove from change set before processing in order to not lose updates
+            final WorkflowInstance instance = podUpdates.remove(podName);
+            return CompletableFuture.runAsync(() -> processPodUpdate(podName, instance), eventExecutor);
+          })
+          .collect(Collectors.toList())
+          .forEach(CompletableFuture::join);
+    }
+
+    private void processPodUpdate(String podName, WorkflowInstance instance) {
+      final Pod pod = client.pods().withName(podName).get();
```
honnix: This will result lots of additional pod readings so we should test this thoroughly in staging.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -168,7 +178,8 @@
     this.pollPodsIntervalSeconds = pollPodsIntervalSeconds;
     this.podDeletionDelaySeconds = podDeletionDelaySeconds;
     this.time = Objects.requireNonNull(time);
-    this.executor = register(closer, Objects.requireNonNull(executor), "kubernetes-docker-runner");
+    this.executor = register(closer, Objects.requireNonNull(executor), "kubernetes-poll");
+    this.eventExecutor = currentContextExecutorService(register(closer, new ForkJoinPool(32), "kubernetes-event"));
```
honnix: Make `32` as a constant?

honnix:
Minor comments, otherwise LGTM. Feel free to ship to staging for performance test.

styx-common/src/test/java/com/spotify/styx/util/GrpcContextUtilTest.java
danielnorberg: Yeah. We should have some reusable utility `verifyNotInstantiable(Class<?> cls)` method

PR #560, opened 2018-08-24
scheduler: cache resource limit lookup in tick

danielnorberg:
Alternative to #555

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -67,6 +67,8 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -221,13 +223,15 @@ private void dequeueInstances(
       Map<WorkflowId, Workflow> workflows,
       List<InstanceState> eligibleInstances) {
 
+    final ConcurrentMap<String, Boolean> resourceExhaustedCache = new ConcurrentHashMap<>();
```
honnix: Where do we invalidate the cache?

danielnorberg: A new instance is created for for every tick.

honnix: Ah ok i read it as instance variable.

danielnorberg:
@honnix Still reviewing this?

honnix:
+1

PR #563, opened 2018-08-24
refactor "Resource limit reached" message handling

danielnorberg:
Use same message emission code in both `Scheduler` and
`QueuedStateManager` to ensure that duplicate messages are not emitted.

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/MessageUtil.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/MessageUtil.java
@@ -1,0 +1,48 @@
+/*
+ * -\-\-
+ * Spotify Styx Scheduler Service
+ * --
+ * Copyright (C) 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.styx;
+
+import static java.util.stream.Collectors.toList;
+
+import com.spotify.styx.model.Event;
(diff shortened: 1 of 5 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -305,14 +305,7 @@ private void tryUpdatingCounter(RunState runState,
...
     if (!depletedResourceIds.isEmpty()) {
-      final List<Resource> depletedResources = depletedResourceIds.stream().map(x ->
-          Resource.create(x, shardedCounter.getCounterSnapshot(x).getLimit()))
-          .collect(toList());
-      final Message message = Message.info(
-          String.format("Resource limit reached for: %s", depletedResources));
-      if (!runState.data().message().map(message::equals).orElse(false)) {
-        receiveIgnoreClosed(Event.info(runState.workflowInstance(), message), runState.counter());
-      }
+      MessageUtil.emitResourceLimitReachedMessage(this, runState, depletedResourceIds);
```
honnix: This changes the content of the message from `Resource.toString()` to plain resource name, to align with `Scheduler`. Interesting, so we have been sending different messages for a while.

styx-scheduler-service/src/main/java/com/spotify/styx/MessageUtil.java
```diff
@@ -0,0 +1,48 @@
...
+public class MessageUtil {
+
+  private MessageUtil() {
+    throw new UnsupportedOperationException();
+  }
+
+  public static void emitResourceLimitReachedMessage(StateManager stateManager, RunState runState,
+      List<String> depletedResources) {
+    if (depletedResources.isEmpty()) {
+      throw new IllegalArgumentException();
```
honnix: We can cover this branch in unit test to get better result of codecov.

honnix:
LGTM with minor comments.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
danielnorberg: Yeah, that's what prompted this refactor. To ensure that we send the same message. Is even in the PR description =)

styx-scheduler-service/src/main/java/com/spotify/styx/MessageUtil.java
danielnorberg: 👍

Issue #545, opened 2018-08-13
Initial workflow happens after "offset" interval which means a long wait time for larger offsets


anish749:
During workflow initialisation, the next trigger instant is `offset` duration _after_ the current time. This means that for larger values of offset (like 30 days, 90 days) the natural trigger would be after 30 days with the partition instant being the current time. While for smaller values (a few hours or 1 day offsets) this works well, but makes it harder to test and iterate fast for larger offsets. Larger offset values are typically used for tracking lagging metrics.

Changing the [initial trigger](https://github.com/spotify/styx/blob/master/styx-api-service/src/main/java/com/spotify/styx/api/workflow/WorkflowInitializer.java#L96) to subtract the offset would solve the issue and would allow for a trigger to happen at the next schedule interval but with partition value which is offset by the given interval behind time. However, this would break the API as documented, since the workflow should trigger after the offset duration, according to the doc.

Proposal:
Add a separate flag in workflow which would allow only the initial trigger to happen at the next schedule interval. (At the next hour if hourly, at midnight, if daily, etc) with a partition value at 'offset' duration back. The following natural triggers can however happen every 'schedule'.

honnix:
Thanks for reporting. Having a flag indicating whether to subtract or add offset sounds good to me.

ulzha:
Unclear at "subtract or add", that's what the offset's sign is for.

I think the issue is just about starting to trigger the workflow ASAP after creation, instead of waiting for offset duration.

I think in the documentation we aren't really stating that the first trigger for a newly created offset workflow will occur only after offset duration. Maybe let's b̶r̶e̶a̶k̶fix it? I haven't heard arguments in favor of the current behavior, only complaints.

honnix:
That was my initial reaction as well. We could just alway subtract offset to get the first `next natural trigger`, and that means an immediate triggering of a partition.

danielnorberg:
This should be fixed now.

anish749:
thanks @danielnorberg

PR #565, opened 2018-08-28
Trigger at next available schedule

anish749:
1. Trigger at next available schedule with an offset which goes back in time. This allows workflow with larger offsets to be scheduled normally.

2. Add support for parsing negative offsets.
The offset format is specified to be `ISO 8601 Duration format`, which allows negative durations. Styx should allow this in order to confirm to this format.

This PR fixes https://github.com/spotify/styx/issues/545


styx-common/src/main/java/com/spotify/styx/util/TimeUtil.java
```diff
@@ -210,36 +214,33 @@ public static boolean isAligned(Instant instant, Schedule schedule) {
    * @return A zoned date time with the offset applied
    */
   public static ZonedDateTime addOffset(ZonedDateTime time, String offset) {
-    final int tPos = offset.indexOf('T');
-    final String periodOffset;
-    final String durationOffset;
+    final Pattern pattern = Pattern.compile("([-+]?)P([-+0-9YMWD]+)?(T([-+0-9HMS.,]+)?)?",
```
danielnorberg: Would be nice to not have to use regex but if we have to, at least make the compiled pattern a `static final` constant

styx-common/src/main/java/com/spotify/styx/util/TimeUtil.java
```diff
@@ -210,36 +214,33 @@ public static boolean isAligned(Instant instant, Schedule schedule) {
...
+        Pattern.CASE_INSENSITIVE);
+    Matcher matcher = pattern.matcher(offset);
```
danielnorberg: final

danielnorberg:
Update docs to reflect new behavior

danielnorberg:
So with this new behavior, enabling a new workflow with schedule `daily` and offset `P90D` seem like it will immediately cause the triggering of 90 past instances. Is that desirable? Feels a little bit like accidental backfill without concurrency control.

Also, trying to understand the need here. With the current behavior unchanged, if running those past instances/partitions is desirable, couldn't that be done by manual triggering or backfill?

anish749:
@honnix @danielnorberg @ulzha Any thoughts?

danielnorberg:
Sorry for the delay. Will try to take a look at this again tomorrow.

honnix:
Sorry for taking so long. However this is a breaking change that we need to understand the overall impact. Would it be possible to make it an opt-in solution?

anish749:
I agree, would you suggest an opt-in by specifying a flag in the workflow specification yaml file?

honnix:
To me it is more an initialization thing than workflow configuration. Maybe a query parameter in the URL? That would however require many many changes, covering from server to client to cli. :(

styx-api-service/src/main/java/com/spotify/styx/api/workflow/WorkflowInitializer.java
```diff
@@ -96,8 +96,8 @@ public WorkflowInitializer(Storage storage, Time time) {
   private TriggerInstantSpec initializeNaturalTrigger(Workflow workflow) {
     final Instant now = time.get();
     final Schedule schedule = workflow.configuration().schedule();
-    final Instant nextTrigger = lastInstant(now, schedule);
-    final Instant nextWithOffset = workflow.configuration().addOffset(nextTrigger);
-    return TriggerInstantSpec.create(nextTrigger, nextWithOffset);
+    final Instant nextTrigger = nextInstant(now, schedule); // Next trigger instant, rounded off by schedule
+    final Instant nextWithOffset = workflow.configuration().subtractOffset(nextTrigger); // gives offset by date
```
danielnorberg: I think this logic is wrong.

Consider the case where a `hourly` workflow has a offset of `30m`. The `TriggerInstantSpec` would then be created with an `instant` of `HH:30:00` but is should be `HH:00:00`. 

I think maybe what you want to do here is only subtract the offset from `now` and not make any other changes.

```java
// Pretend we're in the past for the purpose of scheduling the first trigger
final Instant now = workflow.configuration().subtractOffset(time.get()); 
```

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -168,9 +168,9 @@ private Runnable tryTriggering(Workflow workflow,
       }
 
       final Schedule schedule = workflow.configuration().schedule();
-      final Instant nextTrigger = nextInstant(instantSpec.instant(), schedule);
-      final Instant nextWithOffset = workflow.configuration().addOffset(nextTrigger);
-      final TriggerInstantSpec nextSpec = TriggerInstantSpec.create(nextTrigger, nextWithOffset);
+      final Instant nextTrigger = nextInstant(instantSpec.offsetInstant(), schedule);
```
danielnorberg: With the above proposed change, I do not think these changes are needed.

styx-scheduler-service/src/test/java/com/spotify/styx/StyxSchedulerServiceFixture.java
```diff
@@ -306,9 +306,9 @@ void workflowDeleted(Workflow workflow) throws IOException {
   private TriggerInstantSpec initializeNaturalTrigger(Workflow workflow) {
     final Instant now = time.get();
     final Schedule schedule = workflow.configuration().schedule();
-    final Instant nextTrigger = lastInstant(now, schedule);
-    final Instant nextWithOffset = workflow.configuration().addOffset(nextTrigger);
-    return TriggerInstantSpec.create(nextTrigger, nextWithOffset);
+    final Instant nextTrigger = nextInstant(now, schedule);
```
danielnorberg: The above proposed change should be probably be applied here as well.

danielnorberg:
I think this has potential to be a good default behavior that we don't need an opt-in flag for.

styx-api-service/src/main/java/com/spotify/styx/api/workflow/WorkflowInitializer.java
honnix: Agreed and then it will not cause the issue mentioned in another comment that it becomes backfill without concurrency control.

honnix:
@danielnorberg I saw you have a branch working on this. Still wip?

anish749:
I am sorry, I wasn't able to look into this much this week.

danielnorberg:
@honnix haven't worked on it more than that wip branch

danielnorberg:
Closing in favor of #651

Issue #355, opened 2018-01-12
DatastoreExceptions are not handled


danielnorberg:
The `Storage` and `DatastoreStorage` interfaces declared `throws IOException` but if there is a datastore error `DatastoreException` is thrown. The `DatastoreException` is a runtime exception and is not a `IOException` subclass. Thus all storage call sites that do `catch (IOException e)` will not actually handle datastore exceptions.

E.g. https://github.com/spotify/styx/blob/master/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java#L600

honnix:
fixed by #566

PR #566, opened 2018-08-29
handle datastore exceptions

danielnorberg:
Wraps the `Datastore` client in a `CheckedDatastore` wrapper that throws `IOException` instead of `DatastoreException`.

This forces all callers to handle these errors and leverages all the exception handling we already had in place that was written assuming that storage errors result in `IOException`.

Addresses #355

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
@@ -52,7 +52,7 @@ public class AggregateStorage implements Storage {
 
   public AggregateStorage(Connection connection, Datastore datastore, Duration retryBaseDelay) {
     this(new BigtableStorage(connection, retryBaseDelay),
-         new DatastoreStorage(datastore, retryBaseDelay));
+         new DatastoreStorage(new CheckedDatastore(datastore), retryBaseDelay));
   }
 
   AggregateStorage(BigtableStorage bigtableStorage, DatastoreStorage datastoreStorage) {
@@ -219,17 +219,17 @@ public class AggregateStorage implements Storage {
   }
 
   @Override
-  public Map<Integer, Long> shardsForCounter(String counterId) {
+  public Map<Integer, Long> shardsForCounter(String counterId) throws IOException {
     return datastoreStorage.shardsForCounter(counterId);
   }
 
   @Override
-  public void deleteShardsForCounter(String counterId) {
+  public void deleteShardsForCounter(String counterId) throws IOException {
     datastoreStorage.deleteShardsForCounter(counterId);
   }
 
   @Override
-  public long getLimitForCounter(String counterId) {
+  public long getLimitForCounter(String counterId) throws IOException {
     return datastoreStorage.getLimitForCounter(counterId);
   }
 
@@ -259,7 +259,7 @@ public class AggregateStorage implements Storage {
   }
 
   @Override
-  public Optional<Backfill> backfill(String id) {
+  public Optional<Backfill> backfill(String id) throws IOException {
     return datastoreStorage.getBackfill(id);
   }
 
--- a/styx-common/src/main/java/com/spotify/styx/storage/CheckedDatastoreReaderWriter.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/CheckedDatastoreReaderWriter.java
@@ -1,0 +1,108 @@
+/*
+ * -\-\-
+ * Spotify Styx Scheduler Service
+ * --
+ * Copyright (C) 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.styx.storage;
+
+import static com.google.common.collect.Iterables.toArray;
+
+import com.google.cloud.datastore.DatastoreException;
+import com.google.cloud.datastore.DatastoreReaderWriter;
+import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.EntityQuery;
+import com.google.cloud.datastore.FullEntity;
+import com.google.cloud.datastore.Key;
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Supplier;
+
+/**
+ * A wrapper for {@link DatastoreReaderWriter} that translates unchecked {@link DatastoreException}s to checked
+ * {@link DatastoreIOException}s.
+ */
+class CheckedDatastoreReaderWriter {
+
+  private final DatastoreReaderWriter rw;
+
+  CheckedDatastoreReaderWriter(DatastoreReaderWriter rw) {
+    this.rw = Objects.requireNonNull(rw);
+  }
+
+  Entity put(FullEntity<?> entity) throws IOException {
+    return call(() -> rw.put(entity));
+  }
+
+  Entity add(FullEntity<?> entity) throws IOException {
+    return call(() -> rw.add(entity));
+  }
+
+  void update(Entity entity) throws IOException {
+    run(() -> rw.update(entity));
+  }
+
+  Entity get(Key key) throws IOException {
+    return call(() -> rw.get(key));
+  }
+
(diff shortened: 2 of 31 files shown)


styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -690,7 +690,7 @@ private void retryWorkflowInstance() throws ExecutionException, InterruptedExcep
             .help("Run backfill in reverse, from end (exclusive) to beginning (inclusive)")
             .action(Arguments.storeTrue());
     final Argument backfillCreateConcurrency =
-        backfillCreate.addArgument("concurrency").help("The number of jobs to run in parallel").type(Integer.class);
+        backfillCreate.addArgument("concurrency").help("The number of jobs to query in parallel").type(Integer.class);
```
danielnorberg: Oh no intellij, what have you done

styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
```diff
@@ -52,7 +52,7 @@
 
   public AggregateStorage(Connection connection, Datastore datastore, Duration retryBaseDelay) {
     this(new BigtableStorage(connection, retryBaseDelay),
-         new DatastoreStorage(datastore, retryBaseDelay));
+         new DatastoreStorage(new CheckedDatastore(datastore), retryBaseDelay));
```
honnix: I like the name `CheckedDatastore`.

styx-common/src/main/java/com/spotify/styx/storage/CheckedDatastoreReaderWriter.java
```diff
@@ -0,0 +1,108 @@
...
+
+  void get(Iterable<Key> keys, IOConsumer<Entity> f) throws IOException {
+    run(() -> rw.get(toArray(keys, Key.class)).forEachRemaining(IOConsumer.unchecked(f)));
+  }
+
+  List<Entity> query(EntityQuery query) throws IOException {
+    return call(() -> ImmutableList.copyOf(rw.run(query)));
+  }
+
+  void query(EntityQuery query, IOConsumer<Entity> f) throws IOException {
```
honnix: Is the iterator returned by `rw.run()` lazy? Or Datastore returns all result anyway in one network interaction?

honnix: I guess in our use case it doesn't really matter because everything can fit into memory.

styx-common/src/main/java/com/spotify/styx/storage/StorageTransaction.java
```diff
@@ -70,7 +70,7 @@ WorkflowId storeWorkflowWithNextNaturalTrigger(Workflow workflow, TriggerInstant
    * @param id Id of the backfill
    * @return Optionally a backfill, if one was found for the given id
    */
-  Optional<Backfill> backfill(String id);
+  Optional<Backfill> backfill(String id) throws IOException;
```
honnix: Update javadoc as well?

styx-common/src/main/java/com/spotify/styx/util/ShardedCounterSnapshotFactory.java
```diff
@@ -82,12 +81,12 @@ private static void initShardRange(Storage storage, String counterId, int startI
                                      int endIndex) {
     try {
       storage.runInTransaction(tx -> {
-        IntStream.range(startIndex, endIndex).forEach(index -> {
+        for (int index = startIndex; index < endIndex; index++) {
```
honnix: I like this old school better. It also gives space for compiler optimisation.

styx-common/src/main/java/com/spotify/styx/storage/CheckedDatastoreReaderWriter.java
```diff
@@ -0,0 +1,108 @@
...
+  void update(Entity entity) throws IOException {
+    run(() -> rw.update(entity));
+  }
+
+  Entity get(Key key) throws IOException {
+    return call(() -> rw.get(key));
+  }
+
+  List<Entity> get(Iterable<Key> keys) throws IOException {
+    return call(() -> ImmutableList.copyOf(rw.get(toArray(keys, Key.class))));
```
honnix: At least this `get` is lazy according to Datastore javadoc, and `hasNext` or `next` may throw `DatastoreException`. :/

```
/**
   * Returns an {@link Entity} for each given {@link Key} that exists in the Datastore. The order of
   * the result is unspecified. Results are loaded lazily, so it is possible to get a
   * {@code DatastoreException} from the returned {@code Iterator}'s
   * {@link Iterator#hasNext hasNext} or {@link Iterator#next next} methods. {@link ReadOption}s can
   * be specified if desired.
   *
   * <p>Example of getting multiple entity objects.
   * <pre> {@code
   * String firstKeyName = "my_first_key_name";
   * String secondKeyName = "my_second_key_name";
   * KeyFactory keyFactory = datastore.newKeyFactory().setKind("MyKind");
   * Key firstKey = keyFactory.newKey(firstKeyName);
   * Key secondKey = keyFactory.newKey(secondKeyName);
   * Iterator<Entity> entitiesIterator = datastore.get(Lists.newArrayList(firstKey, secondKey));
   * List<Entity> entities = Lists.newArrayList();
   * while (entitiesIterator.hasNext()) {
   *   Entity entity = entitiesIterator.next();
   *   // do something with the entity
   *   entities.add(entity);
   * }
   * }</pre>
   *
   * @throws DatastoreException upon failure
   * @see #get(Key)
   */
```

honnix:
Looks good overall. The missed exception due to iterator is an issue we need to address.

styx-common/src/main/java/com/spotify/styx/storage/CheckedDatastoreReaderWriter.java
danielnorberg: Datastore does paging when fetching results, so it is lazy.

styx-common/src/main/java/com/spotify/styx/storage/StorageTransaction.java
danielnorberg: We should then add `@throws IOException` to all the other methods in this class as well but that feels like another PR.

styx-common/src/main/java/com/spotify/styx/util/ShardedCounterSnapshotFactory.java
danielnorberg: Yeah, I felt like this was an egregious use of stream, but that's bonus. The real reason for this change is that `tx.store()` now throws `IOException`.

styx-common/src/main/java/com/spotify/styx/storage/CheckedDatastoreReaderWriter.java
danielnorberg: Yeah, so call `List<Entity> get(Iterable<Key> keys)` if you expect the results to be small, and call `void get(Iterable<Key> keys, IOConsumer<Entity> f)` if you expect results to be bigger and/or want to avoid collecting all entities into a list.

Both methods handle `hasNext ` and `next` throwing `DatastoreException`.

danielnorberg: Maybe we're missing some tests here proving that `DatastoreException` is handled.

danielnorberg:
@honnix where is it missing an exception due to iterator?

honnix:
I guess nowhere. I missed those runs and calls.

danielnorberg:
@honnix good to merge?

styx-common/src/main/java/com/spotify/styx/storage/CheckedDatastoreReaderWriter.java
honnix: Shall we reflect the above comment about when to use what with some javadoc?

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -351,7 +337,7 @@ public void updateNextNaturalTrigger(WorkflowId workflowId, TriggerInstantSpec t
     final Iterable<List<WorkflowId>> batches = Iterables.partition(workflowIds,
         MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH_READ);
     return StreamSupport.stream(batches.spliterator(), false)
-        .map(batch -> CompletableFuture.supplyAsync(() -> this.getBatchOfWorkflows(batch), executor))
+        .map(batch -> asyncIO(() -> this.getBatchOfWorkflows(batch)))
```
danielnorberg: Seems this code is not properly propagating IOException

danielnorberg: Fixed

honnix:
I don't feel very confident with this big change. Let's have another brain. :) @NatashaL

NatashaL:
No comments from me 👍

honnix:
@danielnorberg You have got +1 so ship it.

PR #567, opened 2018-08-30
use guava Streams.concat

danielnorberg:


honnix:
Beautiful. Less code. 👍

PR #568, opened 2018-08-30
refactor fjp/executor shutdown

danielnorberg:
Have a single well-tested method for shutting down executor services and use that instead of duplicating variations of the same shutdown logic everywhere.

One of the tests breaks code coverage on Java 8 so it is disabled there but it will be run by our Java 10+ CI pipeline.

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
@@ -23,6 +23,7 @@ package com.spotify.styx.api;
 import static com.spotify.apollo.StatusType.Family.SUCCESSFUL;
 import static com.spotify.styx.api.Api.Version.V3;
 import static com.spotify.styx.serialization.Json.serialize;
+import static com.spotify.styx.util.CloserUtil.register;
 import static com.spotify.styx.util.ParameterUtil.toParameter;
 import static com.spotify.styx.util.StreamUtil.cat;
 import static com.spotify.styx.util.TimeUtil.instantsInRange;
@@ -31,6 +32,7 @@ import static java.util.stream.Collectors.toList;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.collect.Iterables;
+import com.google.common.io.Closer;
 import com.spotify.apollo.Client;
 import com.spotify.apollo.Request;
 import com.spotify.apollo.RequestContext;
@@ -75,7 +77,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ForkJoinTask;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import okio.ByteString;
@@ -88,6 +89,8 @@ public final class BackfillResource implements Closeable {
   private static final String WAITING = "WAITING";
   private static final int CONCURRENCY = 64;
 
+  private final Closer closer = Closer.create();
+
   private final Storage storage;
   private final String schedulerServiceBaseUrl;
   private final WorkflowValidator workflowValidator;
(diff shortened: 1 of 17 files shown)


styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -182,10 +183,11 @@
 
   private static final int REQUEST_CONCURRENCY = 32;
 
+  private final Closer closer = Closer.create();
```
honnix: We can get `closer` from Apollo env, but passing it all the way down from `StyxScheduler` is awful. :(

styx-common/src/test/java/com/spotify/styx/storage/AggregateStorageTest.java
```diff
@@ -53,6 +55,17 @@ public void setUp() throws Exception {
     sut = new AggregateStorage(bigtable, datastore);
   }
 
+  @After
+  public void tearDown() throws Exception {
+    sut.close();
```
honnix: Does `sut` stand for `Storage Under Test`? :D

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -321,16 +320,17 @@ public void create(Environment environment) {
     final Publisher publisher = publisherFactory.apply(environment);
     closer.register(publisher);
 
+    // TODO: is the shutdown timeout of 1 second here sane?
```
honnix: Probably not. :) But it has been there for long.

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -72,11 +73,12 @@
 
   private static final Tracer tracer = Tracing.getTracer();
 
+  private final Closer closer = Closer.create();
```
honnix: Same here.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -136,8 +138,9 @@
   static final String KEEPALIVE_CONTAINER_NAME = "keepalive";
   static final String MAIN_CONTAINER_NAME = "styx-run";
 
-  private final ScheduledExecutorService executor;
+  private final Closer closer = Closer.create();
```
honnix: Same.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -400,18 +396,6 @@ public void close() throws IOException {
       return;
     }
     running = false;
-
-    eventProcessingExecutor.shutdown();
```
honnix: Why is this not needed?

honnix:
Overall a good cleanup! A little bit concern regarding all those closers.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -88,6 +89,8 @@
   private static final String WAITING = "WAITING";
   private static final int CONCURRENCY = 64;
 
+  private final Closer closer = Closer.create();
```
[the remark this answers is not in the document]
danielnorberg: But then we have to pass in a closer in the `BackfillResource` constructor, and `BackfillResource` loses control of the lifecycle of [a resource that it creates](https://github.com/spotify/styx/blob/be201540ec72b00dfd3b2f03caf5797120ddf444/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java#L105). That doesn't make sense to me.

styx-common/src/test/java/com/spotify/styx/storage/AggregateStorageTest.java
danielnorberg: `System Under Test`

http://xunitpatterns.com/SUT.html
https://stackoverflow.com/questions/7321407/what-is-sut-and-where-did-it-come-from
https://en.wikipedia.org/wiki/System_under_test

danielnorberg:
I do not intend to change the life-cycle contracts of these classes in this PR

I believe the `Closer` to be a light-weight tool. Instantiating them where needed is reasonable.

What is your concern regarding the closers?

honnix:
Not really now you explained it.

honnix:
Ship it

PR #581, opened 2018-09-11
encode urls as uris so that URI.create is ok

bergman:

## Description
encode urls in a more compatible way

## Motivation and Context
apollo client seems to use URI.create before passing to underlying
client and URI.create throws on unencoded characters like [ and ]

I have included unit tests.

## Checklist for PR author(s)
- [ ] All tests pass

### DIFF
--- a/styx-client/src/test/java/com/spotify/styx/client/StyxApolloClientTest.java
+++ b/styx-client/src/test/java/com/spotify/styx/client/StyxApolloClientTest.java
@@ -60,7 +60,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
+import java.util.concurrent.CompletionException;
 import junitparams.JUnitParamsRunner;
 import junitparams.Parameters;
 import okhttp3.HttpUrl;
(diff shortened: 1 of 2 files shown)


honnix:
codecov is unhappy about delta changes. :) Want to spend time to cover more?

bergman:
I added a few more tests that weird characters are encoded correctly. Let's see what coverage is now.

styx-client/src/test/java/com/spotify/styx/client/StyxApolloClientTest.java
```diff
@@ -340,16 +360,16 @@ public void testTokenSuccess() {
...
-  public void testTokenFailure() throws Exception {
+  public void testTokenFailure() throws IOException, java.security.GeneralSecurityException {
     final IOException rootCause = new IOException("netsplit!");
     when(auth.getToken(any())).thenThrow(rootCause);
     final StyxApolloClient styx = new StyxApolloClient(client, CLIENT_HOST, auth);
     final CompletableFuture<Void> f = styx.triggerWorkflowInstance("foo", "bar", "baz")
         .toCompletableFuture();
     try {
-      f.get();
+      f.join();
```
honnix: Why the change here?

styx-client/src/test/java/com/spotify/styx/client/StyxApolloClientTest.java
```diff
@@ -340,16 +360,16 @@ public void testTokenSuccess() {
   }
 
   @Test
-  public void testTokenFailure() throws Exception {
+  public void testTokenFailure() throws IOException, java.security.GeneralSecurityException {
```
honnix: We don't need to be so specific I think for test cases, but fine.

styx-client/src/test/java/com/spotify/styx/client/StyxApolloClientTest.java
bergman: unchecked exception, nicer to handle than `ExecutionException` etc

styx-client/src/test/java/com/spotify/styx/client/StyxApolloClientTest.java
bergman: Agreed, will revert

styx-client/src/test/java/com/spotify/styx/client/StyxApolloClientTest.java
bergman: You're right, in this context it doesn't help much. I was too eager :)

bergman:
@honnix How about going for the okhttp PR instead of this one? #584

honnix:
> How about going for the okhttp PR instead of this one? #584
Sounds greater!

bergman:
Closing since #584 was merged

PR #587, opened 2018-09-12
simplify and make it easier to read

honnix:

## Description

Refactoring.

## Motivation and Context

Reduce per method complexity and make it easier to read.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
@@ -39,21 +39,16 @@ public final class ReplayEvents {
       WorkflowInstance workflowInstance,
       Storage storage,
       String backfillId) {
-    final SettableTime time = new SettableTime();
-    boolean backfillFound = false;
 
-    final SortedSet<SequenceEvent> sequenceEvents;
-    try {
-      sequenceEvents = storage.readEvents(workflowInstance);
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
-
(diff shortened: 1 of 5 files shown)


honnix:
codacy doesn't understand custom assertion :(

honnix:
@danielnorberg PTAL

danielnorberg:
Dang, I thought codacy / PMD would recognice custom asserts. This rule has cause trouble before and it has at least never helped me find a case where I forgot to add an assert to my tests. I'd be fine with disabling it.

Alternatively we could switch back to returning bool.

styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
```diff
@@ -35,7 +35,6 @@ private ReplayEvents() {
     throw new UnsupportedOperationException();
   }
 
-  // TODO: fix NPath complexity
```
danielnorberg: Codacy doesn't seem to think that the NPath complexity issue has been resolved

danielnorberg:
We might need to refactor a bit more to get the complexity down.

styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
honnix: Weird. This method is not that complex now I think.

honnix: Now it has been resolved.

honnix:
This is included in  #588

danielnorberg:
Codacy still says “The method getBackfillRunState() has an NPath complexity of 500”

PR #588, opened 2018-09-13
include initial and latest timestamps in RunStateData

honnix:

## Description

Include those timestamps so that for completed workflow instance the client will be able to tell when it started and when it finished.

## Motivation and Context

Timestamps can be used to check how long it took for the workflow instance to complete.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/doc/api.apib
+++ b/doc/api.apib
@@ -459,14 +459,9 @@ Workflow Instance belonging to a Backfill entity.
                 },
                 "statuses": {
                     "active_states": [{
+                        "initial_timestamp": 1536838486145,
+                        "latest_timestamp": 1536838486185,
                         "state": "DONE",
-                        "workflow_instance": {
-                            "parameter": "2017-01-01T00",
-                            "workflow_id": {
-                                "component_id": "styx-canary",
-                                "id": "StyxCanary"
-                            }
-                        },
                         "state_data": {
                             "consecutive_failures": 0,
                             "execution_description": {
@@ -497,6 +492,13 @@ Workflow Instance belonging to a Backfill entity.
                                 "trigger_id": "backfill-1489054446085-53384"
                             },
                             "trigger_id": "backfill-1489054446085-53384"
+                        },
+                        "workflow_instance": {
+                            "parameter": "2017-01-01T00",
+                            "workflow_id": {
+                                "component_id": "styx-canary",
+                                "id": "StyxCanary"
+                            }
                         }
                     }]
                 }
@@ -571,14 +573,9 @@ Workflow Instance belonging to a Backfill entity.
             },
             "statuses": {
                 "active_states": [{
+                    "initial_timestamp": 1536838486145,
+                    "latest_timestamp": 1536838486185,
                     "state": "DONE",
-                    "workflow_instance": {
-                        "parameter": "2017-01-01T00",
-                        "workflow_id": {
-                            "component_id": "styx-canary",
-                            "id": "StyxCanary"
-                        }
-                    },
                     "state_data": {
                         "consecutive_failures": 0,
                         "execution_description": {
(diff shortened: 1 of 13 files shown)


styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
```diff
@@ -35,47 +36,80 @@ private ReplayEvents() {
     throw new UnsupportedOperationException();
   }
 
-  // TODO: fix NPath complexity
-  public static Optional<RunState> getBackfillRunState(
+  public static Optional<RunStateData> getBackfillRunStateData(
```
honnix: This has compile time incompatibility.

styx-common/src/main/java/com/spotify/styx/api/RunStateDataPayload.java
```diff
@@ -20,51 +20,59 @@
...
-    public abstract StateData stateData();
+    StateData stateData();
 
-    @JsonCreator
-    public static RunStateData create(
-        @JsonProperty("workflow_instance") WorkflowInstance workflowInstance,
-        @JsonProperty("state") String state,
-        @JsonProperty("state_data") StateData stateData) {
-      return new AutoValue_RunStateDataPayload_RunStateData(workflowInstance, state, stateData);
+    Optional<Long> initialTimestamp();
```
honnix: This is not forward compatible with old styx-client.

honnix:
@danielnorberg PTAL

danielnorberg:
Codacy still says “The method getBackfillRunState() has an NPath complexity of 500”

styx-common/src/main/java/com/spotify/styx/util/ReplayEvents.java
```diff
@@ -35,47 +36,80 @@ private ReplayEvents() {
...
+          "triggerExecution".equals(EventUtil.name(sequenceEvent.event()));
+
+      if (triggerExecutionEventMet) {
         if (backfillFound) {
-          return Optional.of(restoredState);
+          break;
         }
+
         restoredState = RunState.fresh(workflowInstance, time);
+        initialTime.set(time.get());
```
danielnorberg: This will include all the time spent waiting for dependencies, which for scheduled executions of most workflows is a long time.

We should perhaps only measure the time spent running the container.

honnix:
In the PR results it says fixed. :( It cannot be still that complex.

honnix:
https://app.codacy.com/app/styx/styx/pullRequest?prid=2185568

honnix:
I think It kinda makes sense to include waiting time for estimation.

danielnorberg:
Will backfill instances be waiting for dependencies?

honnix:
Good question, don't know. Maybe or maybe not. :D

danielnorberg:
Maybe look at some backfills and see how they typically behave?

honnix:
If it doesn't wait for dependencies, there is no waiting time and that would be reflected by the two timestamps as well.

danielnorberg:
Are we only using the runtime of previous backfills to estimate runtime of new backfills? Or are we also using the runtime of naturally triggered instances?

danielnorberg:
Ok looks like this PR is only returning runtime of previous backfill instances. In which case nvm and 👍

honnix:
We thought we would only use previous backfills or even the current running backfill only, and see how good/bad that would be. Estimation, guestimation...

doc/api.apib
```diff
@@ -665,14 +669,8 @@ Resources related to inspect detailed status of Workflow Instances.
 
         {
             "active_states": [{
+                "latest_timestamp": 1536838486185,
```
Zatine: Should this not have the `initial_timestamp`?

honnix: Unfortunately it won't. This is active state and we don't store initial_timestamp in datastore, and we probably don't want to read from bigtable to get it.
@danielnorberg and I talked about those missing timestamps in general. So that seems to be an improvement later.

PR #591, opened 2018-09-19
api: add instances?tail=true to get latest WFIs

danielnorberg:

## Description
Allow fetching the _latest_ (by parameter) instances of a workflow.

## Motivation and Context
Users want an easy way to to get the _latest_ instances of a workflow for health checking purposes.

Added unit tests.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
@@ -50,7 +50,6 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
--- a/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
@@ -301,12 +301,18 @@ public final class WorkflowResource {
    * Calculate an offset suitable for getting the {@code limit} latest instances of a workflow.
    */
   private String calculateTailOffset(int limit, Schedule schedule) {
-    // TODO: Optimize away for loop using Schedule.WellKnown
-    Instant offset = Instant.now();
-    for (int i = 0; i < limit; i++) {
-      offset = TimeUtil.previousInstant(offset, schedule);
-    }
-    return ParameterUtil.toParameter(schedule, offset);
+    final Instant offsetInstant = schedule.wellKnown().unit()
+        // Calculate a point in time using the well know schedule time unit
+        .map(unit -> TimeUtil.previousInstant(Instant.now().minus(limit, unit), schedule))
+        // Fall back to iteratively walking back in time
+        .orElseGet(() -> {
+          Instant offset = Instant.now();
+          for (int i = 0; i < limit; i++) {
+            offset = TimeUtil.previousInstant(offset, schedule);
+          }
+          return offset;
+        });
+    return ParameterUtil.toParameter(schedule, offsetInstant);
   }
 
   private Response<WorkflowInstanceExecutionData> instance(
--- a/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
@@ -25,7 +25,6 @@ import static com.spotify.styx.api.Middlewares.json;
 import static com.spotify.styx.serialization.Json.OBJECT_MAPPER;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
 import com.spotify.apollo.Request;
 import com.spotify.apollo.RequestContext;
@@ -51,6 +50,7 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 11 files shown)


styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
```diff
@@ -119,6 +121,11 @@ void writeEvent(SequenceEvent sequenceEvent) throws IOException {
         scan.setStartRow(Bytes.toBytes(offsetInstance.toKey() + '#'));
       }
 
+      // If `tail` is:
+      //   true: Return the _latest_ instances by scanning over _all_ the rows for this workflow (from offset).
+      //   false: Return the _newest_ instances by only scanning over as many rows as necessary.
```
Zatine: What is the difference between latest/newest here? To me it sounds like they both refer to the most recent instances based on parameter, but I suspect that isn't the case? Just wondering if there are more descriptive words we could use.

danielnorberg: That was a typo, good catch 👍

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -279,6 +297,18 @@ public WorkflowResource(Storage storage, WorkflowValidator workflowValidator,
...
   }
 
+  /**
+   * Calculate an offset suitable for getting the {@code limit} latest instances of a workflow.
+   */
+  private String calculateTailOffset(int limit, Schedule schedule) {
+    // TODO: Optimize away for loop using Schedule.WellKnown
+    Instant offset = Instant.now();
+    for (int i = 0; i < limit; i++) {
+      offset = TimeUtil.previousInstant(offset, schedule);
```
honnix: I remember this is pretty heavy computation in case of cron because of overhead in `previousInstant`. So we might want to put down the loop.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -279,6 +297,18 @@ public WorkflowResource(Storage storage, WorkflowValidator workflowValidator,
     return Response.forPayload(data);
   }
 
+  /**
+   * Calculate an offset suitable for getting the {@code limit} latest instances of a workflow.
+   */
+  private String calculateTailOffset(int limit, Schedule schedule) {
+    // TODO: Optimize away for loop using Schedule.WellKnown
+    Instant offset = Instant.now();
```
honnix: Is `now()` a good start? We could use `nextNaturalTrigger` in `WorkflowState` to be more accurate.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -268,6 +272,20 @@ public WorkflowResource(Storage storage, WorkflowValidator workflowValidator,
...
+        if (tail && Strings.isNullOrEmpty(offsetParam)) {
+          // Tailing and no offset supplied. If we do not supply an offset here we might end up scanning a lot of
+          // historical instances. Calculate a suitable offset by stepping back `limit` instances based on the
+          // configured schedule of the workflow.
+          final Optional<Workflow> workflow = storage.workflow(workflowId);
+          if (!workflow.isPresent()) {
+            return Response.forStatus(Status.NOT_FOUND.withReasonPhrase("Could not find workflow."));
+          }
+          final Schedule schedule = workflow.get().configuration().schedule();
+          offset = calculateTailOffset(limit, schedule);
```
honnix: If there is a gap, we will now fetch fewer instances than `limit`, right?

honnix:
How about we compute `start` and `stop` based on `nextNaturalTrigger` and `schedule`, and then fetch instances using `executionData(WorkflowId workflowId, String start, String stop)`? We might fetch fewer instances than desired due to gap with this approach.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
danielnorberg: Not sure what put down the loop means but at least now it's optimized for "well known" time units.

honnix: Something like we did for `TimeUtil::instantsInRange()`.

danielnorberg:
AC: `If there is a gap, ignore and return few number of instances`

styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
```diff
@@ -127,9 +134,17 @@ void writeEvent(SequenceEvent sequenceEvent) throws IOException {
           final int lastHash = key.lastIndexOf('#');
 
           final WorkflowInstance wfi = WorkflowInstance.parseKey(key.substring(0, lastHash));
-          workflowInstancesSet.add(wfi);
-          if (workflowInstancesSet.size() == limit) {
-            break;
+          final boolean added = workflowInstancesSet.add(wfi);
+          if (added) {
+            if (tail) {
+              tailInstances.addLast(wfi);
```
honnix: Can this logic be simplified using `EvictingQueue`? So we don't need to explicitly remove the head.

honnix:
> AC: If there is a gap, ignore and return few number of instances

Yes, and that's why it seems using `start` and `stop` would be simpler than using `offset` and `limit`. Now we compute `offset` anyway which acts like `stop`.

styx-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
danielnorberg: Then we don't know which wfi to remove from `workflowInstancesSet`. Sure, we could refactor this even more to avoid the need for `workflowInstancesSet` but not sure it's worth it.

honnix:
Add test cases to cover cron branch?

PR #600, opened 2018-10-02
meter publishing result

honnix:

## Description

Report publishing success or failure.

## Motivation and Context

We had a bad incident some time ago that publishing didn't work and Styx retried 420 times and that delayed workflow execution badly. We want to have a metric that we can monitor and take early action when it happens again.

Unit test.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
+++ b/styx-common/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
@@ -148,6 +148,13 @@ public final class MetricsStats implements Stats {
       .tagged("what", "workflow-consumer-error-rate")
       .tagged("unit", "error");
 
+  static final MetricId PUBLISHING_RATE = BASE
+      .tagged("what", "publishing-rate");
+
+  static final MetricId PUBLISHING_ERROR_RATE = BASE
+      .tagged("what", "publishing-error-rate")
+      .tagged("unit", "error");
+
   static final MetricId TICK_DURATION = BASE
       .tagged("what", "tick-duration")
       .tagged("unit", UNIT_MILLISECOND);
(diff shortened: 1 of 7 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/PublisherHandler.java
```diff
@@ -40,16 +46,28 @@
...
+  private final Retrier retrier;
   private final Publisher publisher;
+  private final Stats stats;
 
-  public PublisherHandler(Publisher publisher) {
+  public PublisherHandler(Publisher publisher, Stats stats) {
+    this(publisher, stats,
+        Retrier.builder()
+            .errorMessage("publish deploy event")
+            .retryDelay(Duration.ofSeconds(1))
```
danielnorberg: Exponential backoff?

honnix: Could be, but not in this PR I think.

danielnorberg: If we really think it is important to send this event, then we should persist it to durable storage.

honnix: I don't actually know the initial requirement of ITGC.

honnix: This has been very reliable, meaning ~100% in the past years. Not sure it is really worth persisting that includes more moving parts.

danielnorberg:
👍

styx-common/src/main/java/com/spotify/styx/monitoring/MetricsStats.java
```diff
@@ -148,6 +148,13 @@ public boolean isInstance(Metric metric) {
       .tagged("what", "workflow-consumer-error-rate")
       .tagged("unit", "error");
 
+  static final MetricId PUBLISHING_RATE = BASE
+      .tagged("what", "publishing-rate");
+
+  static final MetricId PUBLISHING_ERROR_RATE = BASE
+      .tagged("what", "publishing-error-rate")
+      .tagged("unit", "error");
```
ulzha: BTW I guess it's a wrong unit for a rate.

honnix: I think we once had a PR to remove all those faulty units. No idea why they make their way back. :(

PR #602, opened 2018-10-02
Unhalt backfill

honnix:

## Description

Support unhalting a halted backfill.

## Motivation and Context

Users halt a running backfill for certain reason and afterwards might want to unhalt it instead of creating a new one and trying to figure out what partitions have been backfilled already manually.

A caveat is halted instances will NOT be revived. So this unhalting may not be that useful after all.

unit test and running cil manually.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
+++ b/styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
@@ -210,6 +210,9 @@ public final class CliMain {
             case HALT:
               backfillHalt();
               break;
+            case UNHALT:
+              backfillUnhalt();
+              break;
             case SHOW:
               backfillShow();
               break;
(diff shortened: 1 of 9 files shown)


styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -865,6 +880,7 @@ public Subparser parser(Subparsers subCommands) {
     CREATE("", "Create a backfill"),
     EDIT("e", "Edit a backfill"),
     HALT("h", "Halt a backfill"),
+    UNHALT("u", "Unhalt a halted backfill"),
```
danielnorberg: Could we call this `resume` instead? 😄

danielnorberg:
That caveat sounds bad. Perhaps we can improve things by also adding a new `styx backfill pause` command that only halts the backfill without halting the instances.

honnix:
Exactly

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
honnix: `resume` is not technically/strictly antonym of `halt`. :D 

This is hard, and it was hard when choosing `halt` as I remember.

honnix:
Basically users don't want to have that bad caveat. However halting backfill without halting instances are also bad because that would leave those instances running for nothing; and after resuming, those instances may have already finished (with wrong logic possibly).

honnix:
This is not exactly what users are after, read as they want halted instances to revive.

PR #621, opened 2018-11-07
remove resource kind

honnix:

## Description

Rework of #489

Remove `Resource` kind because it is not used by scheduler.

So the idea (or the way I choose to convince myself regarding resource vs. counter) is, resource is the concept we expose to users while counter is only referred internally.

## Motivation and Context

`Resource` kind is not used for controlling resource usage anymore.

Unit test

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
@@ -215,7 +215,7 @@ public class AggregateStorage implements Storage {
 
   @Override
   public void storeResource(Resource resource) throws IOException {
-    datastoreStorage.postResource(resource);
+    datastoreStorage.storeResource(resource);
   }
 
   @Override
--- a/styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
@@ -214,6 +214,11 @@ public class AggregateStorage implements Storage {
   }
 
   @Override
+  public void storeResource(Resource resource) throws IOException {
+    datastoreStorage.postResource(resource);
+  }
+
+  @Override
   public void storeBackfill(Backfill backfill) throws IOException {
     datastoreStorage.storeBackfill(backfill);
   }
@@ -224,18 +229,8 @@ public class AggregateStorage implements Storage {
   }
 
   @Override
-  public void deleteShardsForCounter(String counterId) throws IOException {
-    datastoreStorage.deleteShardsForCounter(counterId);
-  }
-
-  @Override
   public long getLimitForCounter(String counterId) throws IOException {
     return datastoreStorage.getLimitForCounter(counterId);
-  }
-
-  @Override
-  public void storeResource(Resource resource) throws IOException {
-    datastoreStorage.postResource(resource);
   }
 
   @Override
@@ -266,15 +261,5 @@ public class AggregateStorage implements Storage {
   @Override
   public <T, E extends Exception> T runInTransaction(TransactionFunction<T, E> f) throws IOException, E {
     return datastoreStorage.runInTransaction(f);
-  }
-
-  @Override
-  public void deleteLimitForCounter(String counterId) throws IOException {
-    datastoreStorage.deleteLimitForCounter(counterId);
-  }
-
-  @Override
-  public void updateLimitForCounter(String counterId, long limit) throws IOException {
-    datastoreStorage.updateLimitForCounter(counterId, limit);
   }
 }
--- a/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -768,7 +768,10 @@ public class DatastoreStorage implements Closeable {
         .build(), entity -> shards.add(entity.getKey()));
 
     for (List<Key> batch : Lists.partition(shards, MAX_NUMBER_OF_ENTITIES_IN_ONE_BATCH_WRITE)) {
-      datastore.delete(batch.toArray(new Key[0]));
+      storeWithRetries(() -> {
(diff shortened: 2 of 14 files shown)


styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -736,25 +736,48 @@ void postResource(Resource resource) throws IOException {
...
+  private void deleteShardsForCounter(String counterId) throws IOException {
+    final List<Key> shards = new ArrayList<>();
+    datastore.query(EntityQuery.newEntityQueryBuilder()
+        .setKind(KIND_COUNTER_SHARD)
+        .setFilter(PropertyFilter.eq(PROPERTY_COUNTER_ID, counterId))
+        .build(), entity -> shards.add(entity.getKey()));
+
+    for (List<Key> batch : Lists.partition(shards, MAX_NUMBER_OF_ENTITY_GROUPS_IN_ONE_TRANSACTION)) {
+      storeWithRetries(() -> runInTransaction(transaction -> {
+        datastore.delete(batch.toArray(new Key[0]));
```
danielnorberg: This doesn't actually perform the delete in a transaction as it is calling `datastore.delete()` instead of `transaction.delete()` but in this case I'm not sure it is even necessary to use transactions, seeing as we anyway cannot atomically delete all of the shards atomically in one transaction.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -736,25 +736,48 @@ void postResource(Resource resource) throws IOException {
...
-    return Resource.create(entity.getKey().getName(), entity.getLong(PROPERTY_CONCURRENCY));
+  /**
+   * Delete resource by id. Deletes both counter shards and counter limit if it exists.
+   *
+   * <p>Due to Datastore limitations (modify max 25 entity groups per transaction),
+   * deletion of shards is done in batches of 25 shards.
+   */
+  void deleteResource(String id) throws IOException {
+    storeWithRetries(() -> runInTransaction(tx -> {
+      tx.deleteCounterLimit(id);
```
danielnorberg: No need for transaction here, a single delete is atomic anyway.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -736,25 +736,48 @@ void postResource(Resource resource) throws IOException {
     storeWithRetries(() -> runInTransaction(transaction -> {
```
danielnorberg: Not due to this PR but strange that this method is called `postResource`.

danielnorberg:
👍

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
honnix: This is small enough to get refactored in this PR. So I did it.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -769,10 +769,7 @@ private void deleteShardsForCounter(String counterId) throws IOException {
         .build(), entity -> shards.add(entity.getKey()));
 
     for (List<Key> batch : Lists.partition(shards, MAX_NUMBER_OF_ENTITY_GROUPS_IN_ONE_TRANSACTION)) {
```
danielnorberg: iirc batch size != `MAX_NUMBER_OF_ENTITY_GROUPS_IN_ONE_TRANSACTION`

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: ⭐️

styx-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
```diff
@@ -215,7 +215,7 @@ public void deleteResource(String id) throws IOException {
 
   @Override
   public void storeResource(Resource resource) throws IOException {
-    datastoreStorage.postResource(resource);
+    datastoreStorage.storeResource(resource);
```
danielnorberg: ⭐️

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
honnix: Yeah, right now it is not a part of transaction.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
honnix: It feels somewhat more reliable using transaction than plain batch deletion.

styx-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -761,20 +763,26 @@ void deleteResource(String id) throws IOException {
   }
 
   private void deleteShardsForCounter(String counterId) throws IOException {
-    final List<Key> shards = new ArrayList<>();
+    final List<Shard> shards = new ArrayList<>();
     datastore.query(EntityQuery.newEntityQueryBuilder()
         .setKind(KIND_COUNTER_SHARD)
         .setFilter(PropertyFilter.eq(PROPERTY_COUNTER_ID, counterId))
-        .build(), entity -> shards.add(entity.getKey()));
+        .build(), this::entityToShard);
```
honnix: This feels stupid converting back and force, in order to make `StorageTransaction` look cleaner (not involving `Key` in the interface).

honnix: Not valid comment.

danielnorberg:
👍

PR #639, opened 2018-12-13
swallow 400 as well for member check

honnix:

## Description

As commented in the code, we swallow 400 as well and move on to the next group.

## Motivation and Context

To not to fail too early.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
@@ -304,6 +304,7 @@ public interface ServiceAccountUsageAuthorizer {
           // hasMember API returns 404 if the group does not exist, while returning 400 if the principal
           // email does not exist or does not have the same domain as the group if it is enforced
           if (statusCode == 400 || statusCode == 404) {
+            log.info("Failed to check membership for {} against group {}", principalEmail, group, cause);
             return false;
           }
         }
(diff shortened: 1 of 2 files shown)


honnix:
Codecov is still complaining. Will check later.

styx-api-service/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
```diff
@@ -299,9 +299,13 @@ private boolean isMemberOfGroup(String principalEmail, String group) {
         return TRUE.equals(isMember);
       } catch (ExecutionException e) {
         final Throwable cause = e.getCause();
-        if (cause instanceof GoogleJsonResponseException
-            && ((GoogleJsonResponseException) cause).getStatusCode() == 404) {
-          return false;
+        if (cause instanceof GoogleJsonResponseException) {
+          final int statusCode = ((GoogleJsonResponseException) cause).getStatusCode();
+          // hasMember API returns 404 if the group does not exist, while returning 400 if the principal
+          // email does not exist or does not have the same domain as the group if it is enforced
```
danielnorberg: Feels unpleasant to silently swallow 400 bad request responses, as they could also be caused by legitimately broken requests or other problems. Let's at least log them.

And I'm curious about how we hit this case. How do user emails that do not exist make it through authentication? And do we have groups in our iam policies that do not belong to our domain?

honnix: The case we hit was for a legal service account, we check it against a group (with domain restrictions) having the role, and when that fails and we are not moving on to the next group.

honnix: It is being logged.

Zatine:
👍 looks good to me

PR #640, opened 2018-12-13
use empty list in case of null

honnix:

## Description

`policy.getBindings()` may return `null` if there is no binding.

## Motivation and Context

To fix NPE.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
@@ -254,9 +254,9 @@ public interface ServiceAccountUsageAuthorizer {
           .orElseThrow(() -> new ResponseException(Response.forStatus(
               BAD_REQUEST.withReasonPhrase("Project does not exist: " + projectId))));
 
-      final List<String> members = Optional.ofNullable(policy.getBindings()).orElse(Collections.emptyList()).stream()
+      final List<String> members = emptyListIfNull(policy.getBindings()).stream()
           .filter(binding -> serviceAccountUserRole.equals(binding.getRole()))
-          .flatMap(binding -> binding.getMembers().stream())
+          .flatMap(binding -> emptyListIfNull(binding.getMembers()).stream())
           .collect(toList());
 
       return memberStatus(principalEmail, members);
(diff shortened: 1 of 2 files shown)


styx-api-service/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
```diff
@@ -253,7 +254,7 @@ private String lookupServiceAccountProjectId(String email) {
           .orElseThrow(() -> new ResponseException(Response.forStatus(
               BAD_REQUEST.withReasonPhrase("Project does not exist: " + projectId))));
 
-      final List<String> members = policy.getBindings().stream()
+      final List<String> members = Optional.ofNullable(policy.getBindings()).orElse(Collections.emptyList()).stream()
```
NatashaL: This looks duplicate. Can we move it into `memberStatus`, and have `memberStatus(principalEmail, policy)` instead?

NatashaL:
A test would also be good to have. Otherwise it looks good

styx-api-service/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
honnix: I will add tests but would prefer doing massage before calling the method. It is the matter of contract, I think.

honnix: Plus, it looks the same, but they are actually different policy/binding classes.

styx-api-service/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
```diff
@@ -266,12 +267,11 @@ private String lookupServiceAccountProjectId(String email) {
           .orElseThrow(() -> new ResponseException(Response.forStatus(
               BAD_REQUEST.withReasonPhrase("Service account does not exist: " + serviceAccount))));
 
-      final List<String> members = policy.getBindings().stream()
+      final List<String> members = Optional.ofNullable(policy.getBindings()).orElse(Collections.emptyList()).stream()
           .filter(binding -> serviceAccountUserRole.equals(binding.getRole()))
           .flatMap(binding -> binding.getMembers().stream())
```
danielnorberg: Might `getMembers` also return null?

styx-api-service/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
danielnorberg: Yeah I wanted to do that code deduplication but it's different types so doesn't get much nicer than this.

danielnorberg:
👍

styx-api-service/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
honnix: :(

Zatine:
👍

PR #644, opened 2018-12-28
record demanded resource count

honnix:

## Description

After resource reference validation, record the demand of using.

## Motivation and Context

To be able to understand when out of resource, roughly how bad the situation is in terms
of queuing WFI.

Unit test

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -282,6 +282,8 @@ public class Scheduler {
       return;
     }
 
+    instanceResourceRefs.forEach(stats::recordResourceDemanded);
+
     // Check resource limits. This is racy and can give false positives but the transactional
     // checking happens later. This is just intended to avoid spinning on exhausted resources.
     final List<String> depletedResources = instanceResourceRefs.stream()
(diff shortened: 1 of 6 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -282,6 +282,8 @@ private void dequeueInstance0(final StyxConfig config, Map<String, Resource> res
       return;
     }
 
+    instanceResourceRefs.forEach(stats::recordResourceDemanded);
```
danielnorberg: This should perhaps use a histogram and emit aggregated counts instead. I believe a meter won't produce any intelligible numbers here. For example, if the scheduling interval changes, then the resource demand meter value would change, which doesn't make sense.

honnix: On the other hand, does it make sense to interpret it as requested resources `now`?

danielnorberg: I guess what we want is numbers the gauge the "backlog" of instances that cannot run because they're blocking on an exhausted resource. Emitting an aggregate demand count would provide this.

https://github.com/spotify/styx/pull/650

PR #645, opened 2018-12-28
Custom running state timeout

honnix:

## Description

`runningTimeout` added to workflow configuration to customize timeout value for `running` state.

## Motivation and Context

Users want to configure smaller value than the global default one.

Unit test.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,10 @@ the [Service Account] that Styx itself runs as should be granted `Service Accoun
 role for the `service_account` of the workflow. This can be done by following
 [Granting Roles to Service Accounts].
 
+#### `running_timeout` **[string]**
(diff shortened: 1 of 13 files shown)


NatashaL:
Looks good. I wonder if we should have a lower boundary for the running timeout too? We don't want users to set the timeout too low and cause a flood of timeouts in the events queue. Also, could low timeouts be more flaky due to overhead in communication with kubernetes?

honnix:
I think that should be smeared out by backoff. It does make sense to validate negative duration though.

honnix:
@NatashaL I put lower limit as 60 seconds (reasonable?). PTAL again.

NatashaL:
Sounds reasonable 👍 :shipit:

PR #646, opened 2019-01-08
Add running_timeout check before submitting

NatashaL:

## Description

Early halt workflow instances that have a timeout larger than the running state ttl.

## Motivation and Context

We can't do upperlimit validation of the running_timeout in the api without having to talk to the scheduler. If a user configures a running_timeout larger than the configured ttl for the running state, we wouldn't check this until the workflow has timed-out.  Therefore this PR adds an early check and halts the workflow before it even tries to execute, in the case when running_timeout is too large.

Unit test

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/model/ExecutionDescription.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/ExecutionDescription.java
@@ -21,6 +21,7 @@
 package com.spotify.styx.model;
 
 import io.norberg.automatter.AutoMatter;
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


NatashaL:
@Zatine @danielnorberg ptal

honnix:
We might just configure the upper limit in API service conf file, and would fail workflow deployment.

NatashaL:
@danielnorberg if by merging them you mean have one shared configuration, I looked around a bit and the only approach I found that was kind of sensible was having a new module that will contain this config and use a plugin to share it with other modules. See https://maven.apache.org/plugins/maven-remote-resources-plugin/examples/sharing-resources.html how does that approach sound? I haven't tried it tho, and I'm not sure how we'd expose it to apollo so that it gets picked. Maybe I'm overcomplicating this. Did you have some concrete idea in mind?

NatashaL:
Closing in favour of #654

PR #647, opened 2019-01-09
authz: administrators

danielnorberg:

## Description
A configurable list of users/SAs/groups that are system administrators
and should be authorized to manipulate any workflow and use any
service account.

## Checklist for PR author(s)

- [ ] All tests pass


Zatine:
Looks good to me!

styx-common/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
```diff
@@ -199,6 +203,10 @@ private ServiceAccountUsageAuthorizationResult authorizationCheck(WorkflowId wor
       return ServiceAccountUsageAuthorizationResult.builder()
           .serviceAccountProjectId(projectId)
           .accessMessage(firstPresent(
+              // Check if the principal is an admin
+              () -> memberStatus(principalEmail, administrators)
```
honnix: If we configure `group:...` for staging, this will always result RuntimeException because of missing proper permission, right?

danielnorberg: Yeah

honnix: So we configure individuals then?

PR #653, opened 2019-01-14
Validate the upper bound of running time configuration

narape:
## Description
Somebody could specify an runtime timeout that exceed the implicit
maximum defined in the scheduler.

So to avoid giving false expectations, now an maximum upper value is
defined in the configuration and the workflow validator now enforce that
the value specified does not exceed this limit.

I have included unit tests.

## Checklist for PR author(s)

- [ ] All tests pass


narape:
codecov didn't triggered so closing and reopening

narape:
it didn't work either.

PR #654, opened 2019-01-14
Validate the upper bound of running time configuration

narape:
## Description
Somebody could specify an runtime timeout that exceed the implicit
maximum defined in the scheduler.

So to avoid giving false expectations, now an maximum upper value is
defined in the configuration and the workflow validator now enforce that
the value specified does not exceed this limit.

I have included unit tests.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -395,6 +395,12 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
+        <groupId>com.spotify</groupId>
+        <artifactId>hamcrest-future</artifactId>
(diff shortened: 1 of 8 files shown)


narape:
@NatashaL please review

NatashaL:
There's this example config file https://github.com/spotify/styx/blob/master/styx-standalone-service/src/main/resources/styx-standalone.conf that's used when running both the scheduler and api locally. I think we should add the new config field there too and document that it should be the same as `styx.stale-state-ttls.running`

PR #656, opened 2019-01-17
endpoint for testing SA authz

danielnorberg:

## Checklist for PR author(s)

- [ ] All tests pass


narape:
There are still some issues that I don't like about this code but I will address that in a follow up PR.
I think `ServiceAccountUsageAuthorizer` is doing too much:
* logic to grant or reject authorization
* Enforcing or not access based on config and the response of the previous point.

I will create a new PR to split this class/interface in two

honnix:
Please disregard the comment about `testXxx`. Found an example in gcp lib doing it the same way.

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -73,14 +75,41 @@ public StatusResource(Storage storage) {
...
     return Api.prefixRoutes(routes, V3);
   }
 
+  private Response<TestServiceAccountUsageAuthorizationResponse> testServiceAccountUsageAuthorization(
+      TestServiceAccountUsageAuthorizationRequest request) {
+    final Either<Response<?>, ServiceAccountUsageAuthorizer.ServiceAccountUsageAuthorizationResult> maybeResult =
+        accountUsageAuthorizer.authorizeServiceAccountUsage(request.serviceAccount(), request.principal());
+
+    if (maybeResult.isLeft()) {
+      throw new ResponseException(maybeResult.left().get());
```
honnix: It seems we should return `maybeResult.left().get()` as it is, than throwing a runtime exception which will be interpreted as internal server error.

honnix: This should be handled already by Middleware, not sure how it ended up with internal error.

honnix: OK we are not attaching the right middleware. Fixed.

styx-api-service/src/test/java/com/spotify/styx/api/StatusResourceTest.java
```diff
@@ -86,6 +91,12 @@
 
   private Storage storage;
 
+  private static final String AUTH_SERVICE_ACCOUNT = "foo@bar.iam.gserviceaccounts.com";
+  private static final String AUTH_PRINCIPAL = "bar@example.com";
+  private static final ByteString AUTH_PAYLOAD2 = ByteString.encodeUtf8(
```
danielnorberg: `2`?

Please avoid hand-written json.

danielnorberg:
Added client method + cli command and did some refactoring.

Note that the cli command prints a message with (useful) information and help from the server if authorization failed. The same message that we pass back if authorization fails on a regular request.

styx-api-service/src/test/java/com/spotify/styx/api/StatusResourceTest.java
honnix: We have quite some hand-written json in test cases. I think it serves a purpose to ensure data structure changes are caught by test cases, when hand-written plain string gets deserialized into Java object.

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -638,6 +653,23 @@ private void retryWorkflowInstance() throws ExecutionException, InterruptedExcep
...
+    final TestServiceAccountUsageAuthorizationResponse response = styxClient.testServiceAccountUsageAuthorization(
+        serviceAccount, principal).toCompletableFuture().get();
+
+    if (response.authorized()) {
+      cliOutput.printMessage("The principal " + principal + " is authorized to use the service account "
+                             + serviceAccount + ". " + response.message().orElse(""));
+    } else {
+      cliOutput.printMessage("The principal " + principal + " is not authorized to use the service account "
+                             + serviceAccount + ". " + response.message().orElse(""));
+      throw CliExitException.of(ExitStatus.UnknownError);
```
honnix: Arguable whether this needs to be an error.

styx-cli/src/main/java/com/spotify/styx/cli/CliMain.java
```diff
@@ -1000,6 +1045,31 @@ public Subparser parser(Subparsers subCommands, CliContext cliContext) {
...
+
+    private final String alias;
+    private final String description;
+
+    AuthCommand(String alias, String description) {
+      this.alias = alias;
+      this.description = description;
+    }
+
+    public Subparser parser(Subparsers subCommands, CliContext cliContext) {
```
honnix: This is duplicated everywhere. :( Not part of this PR, but we can consider refactoring this as a default method into an interface somehow.

styx-api-service/src/test/java/com/spotify/styx/api/StatusResourceTest.java
honnix: I think we can talk about and address this issue separately and refactor if we decide to.

PR #659, opened 2019-01-22
target java 11

danielnorberg:

## Description
Target Java 11 instead of Java 8.

## Motivation and Context
* We want to use Java 11 features.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>com.spotify</groupId>
     <artifactId>foss-root</artifactId>
-    <version>5</version>
+    <version>9</version>
   </parent>
 
   <artifactId>styx</artifactId>
@@ -484,8 +484,14 @@
       </dependency>
       <dependency>
         <groupId>com.google.errorprone</groupId>
+        <artifactId>error_prone_core</artifactId>
+        <version>2.3.2</version>
+        <scope>provided</scope>
+      </dependency>
+      <dependency>
+        <groupId>com.google.errorprone</groupId>
         <artifactId>error_prone_annotations</artifactId>
-        <version>2.2.0</version>
+        <version>2.3.2</version>
       </dependency>
       <dependency>
         <groupId>io.grpc</groupId>
@@ -520,6 +526,14 @@
     </dependencies>
   </dependencyManagement>
 
+  <dependencies>
+    <dependency>
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_core</artifactId>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+
   <profiles>
     <profile>
       <id>default</id>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -520,6 +526,14 @@
     </dependencies>
   </dependencyManagement>
 
+  <dependencies>
+    <dependency>
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_core</artifactId>
+      <scope>provided</scope>
```
honnix: Shall we move `scope` to management section to align with `error_prone_core`?

pom.xml
```diff
@@ -528,94 +542,44 @@
...
-          <!--https://github.com/google/error-prone/issues/1091-->
-          <!--https://github.com/google/error-prone/issues/1127-->
           <plugins>
             <plugin>
-              <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-compiler-plugin</artifactId>
               <version>3.8.0</version>
               <configuration>
-                <release>8</release>
+                <source>11</source>
```
honnix: Move them to `properties` to be more visible? `maven.compiler.source` and such.

honnix: And perhaps there will be no need to customize this plugin.

danielnorberg:
Note that we need to have our internal styx cli packaging target java 11 before merging and releasing this.

pom.xml
danielnorberg: I'm not a fan of having the `scope` in the management section, it's too far removed from the use-site and seems all too prone to make mistakes. The way we put `scope` in the management section has been disturbing me for a long time. I'd rather not have it in the management section at all.

pom.xml
danielnorberg: We still need to customize it in order to configure errorprone and linting etc.

pom.xml
honnix: Either way I think. As long as it is consistent. :D

honnix:
👍

pom.xml
```diff
@@ -528,94 +542,44 @@
...
+                     -Xep:PackageLocation:ERROR
+                     -Xep:SwitchDefault:ERROR
+                     -Xep:ThrowsUncheckedException:ERROR
+                     -Xep:TypeParameterNaming:ERROR
+                     -Xep:UngroupedOverloads:WARN
+                     -Xep:UnnecessaryStaticImport:ERROR
+                     -Xep:WildcardImport:ERROR
+                     -XepDisableWarningsInGeneratedCode
+                     -XepExcludedPaths:.*/generated-(test-)?sources/.*
+                     -Xep:Var:OFF
```
bergman: Why the explicit disable of these particular rule? They are not enabled by default afaik.

honnix: I thought they are here to enable, not disable. Maybe I misunderstood. @danielnorberg

danielnorberg: Might be copy pasta from internal root pom

danielnorberg: OFF means disable

honnix: Ah OK, I misunderstood the comment. It was only for that line. `these particular rule` confused me. :D

pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>com.spotify</groupId>
     <artifactId>foss-root</artifactId>
-    <version>5</version>
+    <version>9</version>
   </parent>
 
   <artifactId>styx</artifactId>
```
NatashaL: Should we bump the major or middle(?) version too?

danielnorberg: 👍

danielnorberg: https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

Semver seems to suggest that we should bump either minor (middle) or patch.

pom.xml
danielnorberg: That line and the ones below I guess

danielnorberg: And yes, this was copy-pasta from internal root pom. Not sure why they were disabled there.

pom.xml
danielnorberg: Turns out we needed this because the foss-root pom explicitly sets (and thus overrides) this:
```
            <source>1.8</source>
            <target>1.8</target>
```

pom.xml
bergman: They were disabled there because of another setting that would enable all non-default checks as warnings and some of them were buggy. It should not be needed if you do not have that setting (and error-prone authors do not recommend that setting).

danielnorberg: Aha TIL. Thanks

PR #664, opened 2019-01-29
restore old behaviour when modifying workflow

honnix:

## Description

https://github.com/spotify/styx/pull/651 changed the behaviour and this PR makes that only applicable
to creating new workflows.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/workflow/WorkflowInitializer.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/workflow/WorkflowInitializer.java
@@ -20,6 +20,7 @@
 
 package com.spotify.styx.api.workflow;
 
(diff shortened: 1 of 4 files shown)


honnix:
Not sure this is worth. Getting complicated.

danielnorberg:
This would revert to the undesired behavior that we tried to address in https://github.com/spotify/styx/pull/651 when making changes to a workflow. For example, if a workflow has an offset of `P30D` and e.g. the offset is changed to `P31D`, then the next execution will  now happen 31 days later. And to add insult to injury, 30 days worth of instances will be skipped.

honnix:
So not a valid change. I thought it was only the issue when creating a new workflow.

PR #669, opened 2019-02-07
authz: principal blacklist

danielnorberg:

## Description
Allow blacklisting specific undesirable principals from authorizing.

## Motivation and Context
Protect against users granting authorization access to principals like e.g. a global default CI/CD service account.

Those workflows would otherwise be modifiable by (more or less) _any_ user, which is not intended.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
+++ b/styx-common/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
@@ -497,7 +497,8 @@ public interface ServiceAccountUsageAuthorizer {
                                               String gsuiteUserEmail,
                                               String serviceName,
                                               String message,
-                                              List<String> administrators, List<String> blacklist) {
+                                              List<String> administrators,
+                                              List<String> blacklist) {
 
     final HttpTransport httpTransport;
     try {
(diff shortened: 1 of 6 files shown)


styx-common/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
```diff
@@ -472,7 +497,7 @@ static ServiceAccountUsageAuthorizer create(String serviceAccountUserRole,
                                               String gsuiteUserEmail,
                                               String serviceName,
                                               String message,
-                                              List<String> administrators) {
+                                              List<String> administrators, List<String> blacklist) {
```
honnix: Let's make it to a new line.

danielnorberg: 👍

danielnorberg:
I guess https://github.com/spotify/styx/blob/master/doc/api.apib should also be updated

honnix:
I will do that.

honnix:
Let's take it in a separated PR though.

honnix:
Take liberty merging it.

PR #674, opened 2019-02-07
Revert "use later version of typesafe config"

honnix:
Reverts spotify/styx#668


honnix:
@danielnorberg PTAL

danielnorberg:
Why reverting this? No longer needed?

PR #679, opened 2019-02-22
client: auth: GCE default credentials

danielnorberg:

## Description
Add support to styx client for authenticating using GCE default credentials.

## Motivation and Context
When using default credentials, `GoogleIdTokenAuth` will check if using GCE credentials and in that case fetch id tokens from the GCE instance metadata server.

This is so that users using GCE default credentials can use the styx client lib / cli without needing to provision service account keys.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
+++ b/styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
@@ -26,11 +26,16 @@ import com.google.api.client.auth.oauth2.TokenResponse;
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.api.client.googleapis.util.Utils;
 import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpHeaders;
 import com.google.api.client.http.HttpTransport;
+import com.google.api.client.http.UriTemplate;
 import com.google.api.client.json.JsonFactory;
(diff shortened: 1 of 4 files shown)


honnix:
@danielnorberg Worked on a PR to remove the boolean flag which seems a bit awkward. https://github.com/spotify/styx/pull/682 Not sure it makes sense given we want to rewrite this part using cloud lib.

+1 to this PR otherwise. We can ignore the delta complaining in this case.

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
```diff
@@ -70,9 +88,24 @@ private String getToken(String targetAudience, GoogleCredential credential)
     }
   }
 
+  private String getDefaultGCEIdToken(String targetAudience) throws IOException {
+    // https://cloud.google.com/compute/docs/instances/verifying-instance-identity#request_signature
+    final String metadataHost = System.getenv().getOrDefault("GCE_METADATA_HOST", DEFAULT_GCE_METADATA_HOST);
+    final String uriTemplate = "http://" + metadataHost + GCE_METADATA_IDENTITY_PATH;
```
bergman: Given the default value
```java
private static final String DEFAULT_GCE_METADATA_HOST = "http://169.254.169.254";
```

this will prefix `http://` twice, leading to url like `http://http://169.254.169.254`

PR #680, opened 2019-02-22
2.0.x: auth: support GCE default credentials

danielnorberg:
backport of #679


styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
```diff
@@ -41,19 +46,32 @@
...
   private static final JsonFactory JSON_FACTORY = Utils.getDefaultJsonFactory();
+  private static final String DEFAULT_GCE_METADATA_HOST = "http://169.254.169.254";
+  private static final String GCE_METADATA_IDENTITY_PATH =
+      "/computeMetadata/v1/instance/service-accounts/default/identity{?audience,format}";
 
   private final HttpTransport httpTransport;
   private final Optional<GoogleCredential> credential;
+  private final boolean isGCE;
 
   GoogleIdTokenAuth(
```
bergman: While you're here... how about making this one public? Or maybe it's one layer above, in the client setup. Basically, it would be cool if I could pass my own credentials (or `ImpersonatedCredential`) when creating the styx client.

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
```diff
@@ -70,9 +88,24 @@ private String getToken(String targetAudience, GoogleCredential credential)
     }
   }
 
+  private String getDefaultGCEIdToken(String targetAudience) throws IOException {
```
bergman: Does this cover other non-materialized keys situations? Maybe the metadata server situation is the only one, not sure.

styx-client/src/test/java/com/spotify/styx/client/auth/GoogleIdTokenAuthTest.java
```diff
@@ -0,0 +1,84 @@
...
+    final Optional<String> token = idTokenAuth.getToken("http://styx.foo.bar");
+    assertThat(token, is(Optional.of(TEST_ID_TOKEN)));
+    final RecordedRequest detectionRequest = metadataServer.takeRequest();
+    assertThat(detectionRequest.getPath(), is("/"));
+    final RecordedRequest tokenRequest = metadataServer.takeRequest();
+    assertThat(tokenRequest.getPath(), is("/computeMetadata/v1/instance/service-accounts/default/identity"
+                                          + "?audience=http://styx.foo.bar&format=full"));
+    assertThat(tokenRequest.getHeader("Metadata-Flavor"), is("Google"));
+  }
+}
```
bergman: Files should end with a line terminator :)

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
danielnorberg: I'd like to do this in a separate PR, along with switching from `GoogleCredential` to `GoogleCredentials` along with exposing `StyxClientFactory` methods that allow passing in credentials.

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
danielnorberg: No, but this PR is only about the GCE default credentials (via metadata server) use case.

honnix: I misunderstood the problem when creating Jira ticket. Supporting non-materialized keys situation is more urgent than support GCE default.

honnix:
+1 with comments the same as https://github.com/spotify/styx/pull/679

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
danielnorberg: Yes, styx client & cli should work with our internal `toogle` without this PR.

PR #681, opened 2019-02-27
avoid GCP API interaction if not necessary

honnix:

## Description

Apply blacklist and admin check before making any GCP API interaction.

## Motivation and Context

There are cases that getting service account would fail due to non-exist GCP project or non-exist
service account:

```
API error: 500 Internal Server Error (Request ID: 63671a603b6d4bd5a08487c505f4731b): UncheckedExecutionException: java.lang.RuntimeException: java.util.concurrent.ExecutionException: com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden {   "code" : 403,   "errors" : [ {     "domain" : "global",     "message" : "Permission iam.serviceAccounts.get is required to perform this operation on service account projects/-/serviceAccounts/...",     "reason" : "forbidden"   } ],   "message" : "Permission iam.serviceAccounts.get is required to perform this operation on service account projects/-/serviceAccounts/...",   "status" : "PERMISSION_DENIED" }: GoogleJsonResponseException: 403 Forbidden {   "code" : 403,   "errors" : [ {     "domain" : "global",     "message" : "Permission iam.serviceAccounts.get is required to perform this operation on service account projects/-/serviceAccounts/....",     "reason" : "forbidden"   } ],   "message" : "Permission iam.serviceAccounts.get is required to perform this operation on service account projects/-/serviceAccounts/....",   "status" : "PERMISSION_DENIED" }
```
When this happens, neither user or admins would be able to delete the dangling workflow. With this change, at least admins will be able to force taking any action without depending on GCP.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
+++ b/styx-common/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
@@ -51,6 +51,7 @@ import com.google.api.services.cloudresourcemanager.model.GetIamPolicyRequest;
 import com.google.api.services.iam.v1.Iam;
 import com.google.api.services.iam.v1.IamScopes;
 import com.google.api.services.iam.v1.model.ServiceAccount;
+import com.google.common.base.Suppliers;
 import com.google.common.base.Throwables;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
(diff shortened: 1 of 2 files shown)


styx-common/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
```diff
@@ -214,44 +214,50 @@ public void authorizeServiceAccountUsage(WorkflowId workflowId, String serviceAc
...
-              .map(type -> String.format("Principal %s has role %s on service account %s %s",
-                  principalEmail, serviceAccountUserRole, serviceAccount, type)));
-
-      final ServiceAccountUsageAuthorizationResultBuilder result = ServiceAccountUsageAuthorizationResult.builder()
-              .serviceAccountProjectId(projectId)
-              .authorized(accessMessage.isPresent())
-              .message(accessMessage.orElseGet(() -> denialMessage(serviceAccount, principalEmail, projectId)));
-
-      return result.build();
+                                                                                          String principalEmail) {
```
narape: It is weird the division of concerns between `checkServiceAccountUsageAuthorization` and `checkServiceAccountUsageAuthorization0` and the naming also.

What about we keep the public method high level and then we create small private methods for every check. Something like this:
```java
@Override
    public ServiceAccountUsageAuthorizationResult checkServiceAccountUsageAuthorization(String serviceAccount,	    public ServiceAccountUsageAuthorizationResult checkServiceAccountUsageAuthorization(String serviceAccount,
                                                                                        String principalEmail) {	                                                                                        String principalEmail) {
  return checkIsServiceAccountBlacklisted(serviceAccount, principalEmail)
      .or(() -> checkIsPrincipalAdmin(serviceAccount, principalEmail))
      .orElseGet(() -> {
         final String projectId = serviceAccountProjectId(serviceAccount);
         return checkRole(serviceAccount, principalEmail, projectId)
                  .or(() -> denialMessage(serviceAccount, principalEmail, projectId));
       });
}
```

narape: We will also be using method with names instead of comments to explain logic

styx-common/src/main/java/com/spotify/styx/api/ServiceAccountUsageAuthorizer.java
```diff
@@ -214,50 +215,64 @@ public void authorizeServiceAccountUsage(WorkflowId workflowId, String serviceAc
     @Override
     public ServiceAccountUsageAuthorizationResult checkServiceAccountUsageAuthorization(String serviceAccount,
                                                                                         String principalEmail) {
-      // First check if principal is blacklisted
+      final Supplier<String> projectIdSupplier = Suppliers.memoize(() -> serviceAccountProjectId(serviceAccount));
```
narape: Nice good to know. It is cleaner this way than my original suggestion

PR #683, opened 2019-03-05
client: configurable credentials

danielnorberg:

## Description
* Switch to "cloud client" GoogleCredentials
* Support `ImpersonatedCredentials`

## Motivation and Context
* Make it easier for services to use the styx client and manage workflows on behalf of users.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
+++ b/styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
@@ -20,10 +20,11 @@
 
 package com.spotify.styx.client.auth;
 
+import com.google.api.client.auth.oauth2.ClientParametersAuthentication;
 import com.google.api.client.auth.oauth2.RefreshTokenRequest;
 import com.google.api.client.auth.oauth2.TokenRequest;
 import com.google.api.client.auth.oauth2.TokenResponse;
-import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.client.googleapis.util.Utils;
 import com.google.api.client.http.GenericUrl;
 import com.google.api.client.http.HttpHeaders;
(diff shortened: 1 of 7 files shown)


danielnorberg:
Perhaps we should set up a GCP project where we can test this code (and in the future do e2e tests etc).

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
```diff
@@ -103,22 +114,85 @@ private String getDefaultGCEIdToken(String targetAudience) throws IOException {
...
-        targetAudience, credential.getServiceAccountId(), credential.getTokenServerEncodedUrl());
+        targetAudience, credential.getAccount(), credential.getTokenServerUri().toString());
     request.put("assertion", JsonWebSignature.signUsingRsaSha256(
-        credential.getServiceAccountPrivateKey(), JSON_FACTORY, header, payload));
+        credential.getPrivateKey(), JSON_FACTORY, header, payload));
     final TokenResponse response = request.execute();
     return (String) response.get("id_token");
   }
 
+  private String getImpersonatedIdToken(ImpersonatedCredentials credentials, String targetAudience) throws IOException {
```
narape: We should add unit tests for these new kind of credentials

Also this class looks complicated with a lot of nontrivial logic smashed all together

danielnorberg: Yeah like I mentioned above, we should probably set up a GCP project where we can test this code. I do not believe that _unit_ tests would give much confidence here due to tight coupling with the IAM service api. We cannot emulate or mock IAM well enough to know whether it would work in production. We need to integration test this against the real IAM api.

danielnorberg:
Added tests that will run against the real IAM api if there are GCP credentials available. Note that these will currently not run as part of the regular circleci pipeline as we do not have any credentials there. For now let's set up the internal styx-release pipeline to run these tests.

Unfortunately I'm not sure how we can run real integration tests for user credentials in a CI pipeline as they require a set of real user credentials, but to compensate for that there is at least a unit test for user credentials that mock out the token server. Not great but at least something to lock down current behavior which has been manually tested and verified to work.

All tests can also be run manually by dataex members.

styx-client/src/main/java/com/spotify/styx/client/auth/GoogleIdTokenAuth.java
narape: It makes sense

danielnorberg:
I've now set up circle-ci to run these tests.

PR #687, opened 2019-03-06
emit events with counter

danielnorberg:

## Description
Emit events with the counter of the origin run state.

## Motivation and Context
Be more resilient against scheduler races.

## Checklist for PR author(s)

- [ ] All tests pass


danielnorberg:
@honnix unfortunately, no. The one without the counter is still called in places that doesn't have a counter. E.g. `SchedulerResource` when halting, etc.

We could get rid of all of those by making them first fetch the current run state so they have a counter to inject, but that seems like a bigger and less obviously desirable change.

PR #688, opened 2019-03-06
At least once transition

danielnorberg:

## Description
Prevent instances getting stuck by making event transitions at-least-once.

* Make output handlers tolerate being called more than once for a single state (i.e. they should be idempotent or harmless).
* Ensure that output handlers are called at least once by setting up a scheduled tick that calls the output handlers for all active instance.

## Motivation and Context
Styx currently implements at-most-once state transitions. If a transition fails, it will never be retried.

Because of this, some instances can get stuck when a sequence of transitions fail due to ephemeral failure. For example, the transition from `TERMINATED` to `DONE` could fail due to datastore contention and then the instance would stay in `TERMINATED` until it times out.

By having a scheduled tick that calls the output handler again, the above example would be resolved by the `TerminationHandler` getting called again until the transition to `DONE` succeeds.

## Checklist for PR author(s)

- [ ] All tests pass


styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/PublisherHandler.java
```diff
@@ -26,22 +26,23 @@
...
+import java.util.function.BiConsumer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * An {@link OutputHandler} that integrates {@link RunState.State} values with a {@link Publisher}.
+ * An event consumer that integrates {@link RunState.State} values with a {@link Publisher}.
  */
-public class PublisherHandler implements OutputHandler {
+public class PublisherHandler implements BiConsumer<SequenceEvent, RunState> {
```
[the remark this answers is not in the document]
honnix: Not as simple as I thought due to the fanout thing.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -122,6 +122,14 @@ public QueuedStateManager(
     this.log = Objects.requireNonNull(logger, "logger");
   }
 
+  @Override
+  public void tick() {
+    var states = getActiveStates();
```
honnix: There will be competition with the state machine. Do we need to ensure all handlers are sort of idempotent, and we  rely on the counter to take care of duplicated events fired from the handlers?

honnix: `TransitionLogger` might output `transition ->` for each `tick` when the workflow instance is running. It is not harmful, but could be confusing.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/PublisherHandler.java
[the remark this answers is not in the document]
honnix: Tried in https://github.com/spotify/styx/pull/691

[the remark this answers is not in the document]
danielnorberg: I guess it could be rewritten to act on the event instead of switching on the new state. But that's just a cosmetic change.

Another approach here could be to still have `PublisherHandler` be an `OutputHandler` so it gets called at-least-once, and e.g. keep "internal" (possibly) persisted state or specially craft pubsub messages to be idempotent and/or effectively publish exactly-once. I considered looking into that but I do not believe the at-most-once behavior of PublisherHandler to be a regression (it matches current behavior before this PR) and so can be considered out of scope for this PR. I'd rather revisit in another PR if necessary.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
```diff
@@ -122,6 +122,14 @@ public QueuedStateManager(
...
+    for (final RunState runState : states.values()) {
```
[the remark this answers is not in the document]
danielnorberg: sure

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
```diff
@@ -89,17 +80,26 @@ public void transitionInto(RunState state) {
...
                 state.counter() + 1);
           } catch (IsClosedException isClosedException) {
             LOG.warn("Failed to send 'runError' event", isClosedException);
           }
+          return;
+        }
+
+        final Event submitted = Event.submitted(state.workflowInstance(), runSpec.executionId());
+        try {
+          stateManager.receive(submitted, state.counter());
```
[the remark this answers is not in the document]
danielnorberg: Good catch, we should not use `state.counter() + 1` at all.

Added a comment on the reason for emitting `submitted` after starting the execution.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
danielnorberg: Yes, all handlers should be idempotent. Should I rephrase the PR description?

Perhaps the `TransitionLogger` should also be an event consumer.

danielnorberg: Wrt competition, note that our schedulers already compete.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/PublisherHandler.java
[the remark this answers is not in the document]
danielnorberg: Here's a PR that make the `PublisherHandler` observe the event instead of the state. It feels a bit pointless though. https://github.com/spotify/styx/pull/692

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
danielnorberg: Another comment wrt competition, I would like to work on reducing contention overall in another PR.

honnix: > all handlers should be idempotent

We can put this into `OutputHandler` interface.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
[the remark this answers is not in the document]
honnix: > Emit submitted event first to guarantee it is observed before events from the pod

Isn't this still valid? If we have received events from pod, this `submitted` event will become invalid due to counter mismatching.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
```diff
@@ -89,17 +80,26 @@ public void transitionInto(RunState state) {
...
+        } catch (IsClosedException isClosedException) {
+          LOG.warn("Could not emit 'submitted' event", isClosedException);
         }
         break;
 
       case TERMINATED:
       case FAILED:
       case ERROR:
+        // TODO: remove this effectively unused cleanup?
```
honnix: We should remove `cleanup` in DockerRunner interface as well if we remove this. It was kept for separation of concern.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
honnix: I'm fine changing `TransitionLogger` as well, because it will be quite confusing when looking at the log.

styx-standalone-service/src/main/resources/styx-standalone.conf
```diff
@@ -2,12 +2,12 @@ styx.mode = "production"
...
-  creating       = "PT1M"
+  new            = "PT30M"
+  creating       = "PT30M"
   submitted      = "PT10M"
   running        = "PT24H"
-  terminated     = "PT1M"
-  failed         = "PT1M"
+  terminated     = "PT30M"
+  failed         = "PT30M"
   awaiting_retry = "PT8H"
```
honnix: Just discovered this is not really used. But let's kill it in a separated PR.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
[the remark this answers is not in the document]
danielnorberg: The reason to emit the `submitted` event was eliminated two years ago: https://github.com/spotify/styx/pull/198

The original reason was so that events from k8s could observe the execution id that was created (and persisted) as part of the `submitted` event.

But now we're persisting the execution id in the `submit` event, so emitting `submitted` after starting the execution is safe.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
danielnorberg: I'd be happy to remove that and the rotting/unused `LocalDockerRunner` as well and just require k8s. For local execution there's minikube and docker desktop (with k8s bundled) anyway.

styx-scheduler-service/src/main/java/com/spotify/styx/state/QueuedStateManager.java
danielnorberg: Done

styx-standalone-service/src/main/resources/styx-standalone.conf
honnix: I meant awaiting_retry. Sorry for confusion.

danielnorberg: Aha

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
honnix: I can leave it to another PR.

honnix:
This is in a mergable state to me. I can take it over and ship it to staging after a final check later.

danielnorberg:
Maven repo seems flaky, CI fails because it can't download some dependency 😫

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java
danielnorberg: Yeah

danielnorberg:
Flaky system test. Circle-ci should have a "run tests until successful" feature ;P

honnix:
Sell Styx to circle ci? :D

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TransitionLogger.java
```diff
@@ -23,26 +23,37 @@
...
+    this.log = Objects.requireNonNull(log, "log");
   }
 
   @Override
-  public void transitionInto(RunState state) {
-    final String name = state.state().name().toLowerCase();
-    LOG.info("{}{} transition -> {} {}", prefix, state.workflowInstance(), name, stateInfo(state));
+  public void accept(SequenceEvent sequenceEvent, RunState state) {
+    var stateName = state.state().name().toLowerCase(Locale.ROOT);
+    log.info("{} transition #{} {}({}) -> {} {}",
```
honnix: This will break `user/processed-internal-event` logs metric in styx GCP project. We can fix later.

danielnorberg: 😫 
Scraping unstructured logs for metrics is not a great practice.

danielnorberg: We don't event seem to be using this metric. It's not in any dashboard I can find. If we care about it, let's make it a real metric.

PR #691, opened 2019-03-07
skip at most once handlers

honnix:

## Description

Introducing `AtMostOnceOutputHandler` so we can skip when trying to transition.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/state/FanOutputHandler.java
+++ b/styx-common/src/main/java/com/spotify/styx/state/FanOutputHandler.java
@@ -20,7 +20,12 @@
 
 package com.spotify.styx.state;
 
+import static java.util.function.Predicate.not;
+
+import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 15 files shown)


styx-common/src/main/java/com/spotify/styx/state/FanOutputHandler.java
```diff
@@ -48,4 +52,17 @@ public void transitionInto(RunState state) {
...
+
+  @Override
+  public void transitionInto(RunState state) {
+    transitionInto(state, outputHandlers);
+  }
+
+  @Override
+  public void tryTransitionInto(RunState state) {
+    final List<OutputHandler> outputHandlers = StreamSupport.stream(this.outputHandlers.spliterator(), false)
+        .filter(not(outputHandler -> outputHandler instanceof AtMostOnceOutputHandler))
```
danielnorberg: So the `FanOutputHandler` would not make forward any calls to `AtMostOnceOutputHandler` instances at all? That's surprising behavior.

Then there is nothing calling the `PublisherHandler`, unless you'd add some additional logic e.g. to the QueuedStateManager to specifically call the `AtMostOnceOutputHandler` instances. At which point this is starting to add more complexity than I'd want.

honnix: Given tryTransitionInto is designed to be called outside of state machine, yes. But I agree, I don't like this very much either.

PR #692, opened 2019-03-08
make PublisherHandler inspect event

danielnorberg:

## Description
Make `PublisherHandler` inspect event type instead of state.

## Motivation and Context
This is a purely cosmetic change without any practical implications in the current styx implementation, as observing the transition to the `SUBMITTED` and `RUNNING` states is equivalent to observing the `submitted` and `started` events.

It could be meaningful if we were to e.g. introduce transitions `SUBMITTED` -> `SUBMITTED` and/or `RUNNING` -> `RUNNING` by emitting `info` events in these states, but not sure we're going to do that.

In fact, I think the `info` event is harmful and we should instead work to eliminate it and replace it with direct mutation of the `message` field in `RunState` and/or the ability to append a new `message` in any event. 

## Checklist for PR author(s)

- [ ] All tests pass


danielnorberg:
Closing this for now as it's not a very useful change.

PR #693, opened 2019-03-08
🔥 partial active instance reads 🔥

danielnorberg:

## Description
Allow partial success when reading active instances from datastore.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/FutureUtil.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/FutureUtil.java
@@ -20,15 +20,12 @@
 
 package com.spotify.styx.util;
 
-import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableList;
-import java.io.IOException;
-import java.util.List;
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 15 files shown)


styx-common/src/main/java/com/spotify/styx/util/FutureUtil.java
```diff
@@ -42,28 +39,20 @@ private FutureUtil() {
     return future;
   }
 
+  public static <T> CompletableFuture<T> exceptionallyCompletedFuture(final Throwable t, final Class<T> cls) {
```
honnix: Is this only for test?

danielnorberg: not needed, removed

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -596,7 +598,11 @@ synchronized void tryPollPods() {
         continue;
       }
 
-      final RunState runState = activeStates.get(workflowInstance.get());
+      if (unavailableWorkflowInstance.test(workflowInstance.get())) {
+        LOG.debug("Ignoring unavailable workflow instance: {}", workflowInstance.get());
```
honnix: We should emit a meter and set up an alert if it goes high.

PR #695, opened 2019-03-08
🔥 temporarily shift the shard by one

honnix:

## Description

1 and 108 are currently broken. We don't use them and overload a bit to 2 and 109.

## Checklist for PR author(s)

- [ ] All tests pass


danielnorberg:
so happy we never had to roll this out =)

An improvement here to consider if this ever becomes necessary in the future could be to "create new shards" instead of shifting, e.g. by prefixing with `backup-` or something in order to not put additional strain on existing shards (or the tablets storing them).

PR #702, opened 2019-03-13
Mitigate datastore contention

danielnorberg:

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -281,7 +281,11 @@ public class Scheduler {
     }
 
     // Check for execution blocker
-    if (executionIsBlocked(config, instance, runState)) {
+    var blocker = executionBlocker(config, instance);
+    if (blocker.isPresent()) {
+      var retry = Event.retryAfter(instance, blocker.get().delay().toMillis());
+      stateManager.receiveIgnoreClosed(retry, runState.counter());
+      LOG.debug("Dequeue rescheduled: {}: {}", instance, blocker.get());
       return;
     }
 
@@ -290,30 +294,21 @@ public class Scheduler {
     sendDequeue(instance, runState, instanceResourceRefs);
   }
 
-  private boolean executionIsBlocked(StyxConfig config, WorkflowInstance instance, RunState runState) {
+  private Optional<ExecutionBlocker> executionBlocker(StyxConfig config, WorkflowInstance instance) {
     if (!config.executionGatingEnabled()) {
-      return false;
+      return Optional.empty();
     }
 
-    Optional<ExecutionBlocker> blocker = Optional.empty();
     try {
-      blocker = gate.executionBlocker(instance)
+      return gate.executionBlocker(instance)
           .toCompletableFuture().get(10, TimeUnit.SECONDS);
     } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
       throw new RuntimeException(e);
     } catch (ExecutionException | TimeoutException e) {
       LOG.warn("Failed to check execution blocker for {}, assuming there is no blocker", instance, e);
+      return Optional.empty();
     }
-
-    if (blocker.isEmpty()) {
-      return false;
-    }
-
-    var retry = Event.retryAfter(instance, blocker.get().delay().toMillis());
-    stateManager.receiveIgnoreClosed(retry, runState.counter());
-    LOG.debug("Dequeue rescheduled: {}: {}", instance, blocker.get());
-    return true;
   }
 
   private Optional<Workflow> readWorkflow(WorkflowId workflowId) {
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -22,13 +22,9 @@ package com.spotify.styx;
 
 import static com.google.common.base.CaseFormat.LOWER_UNDERSCORE;
 import static com.google.common.base.CaseFormat.UPPER_CAMEL;
-import static com.spotify.styx.state.StateUtil.getActiveInstanceStates;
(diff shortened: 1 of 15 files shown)


danielnorberg:
Seems there might be some test flakyness

narape:
Here it seems that it failed during statics analysis

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -476,18 +478,19 @@ private boolean isNonDeletePeriodExpired(ContainerStatus cs) {
...
       return true;
     }
-    final Instant deadline = time.get().minus(Duration.ofSeconds(podDeletionDelaySeconds));
+    final Instant deadline = time.get().minus(podDeletionDelay);
     return finishedAt.isBefore(deadline);
   }
 
   @VisibleForTesting
-  void deletePod(WorkflowInstance workflowInstance, Pod pod, String reason) {
+  boolean deletePod(WorkflowInstance workflowInstance, Pod pod, String reason) {
```
narape: This method name suggest an action: delete the pod, but now the implementation just return a boolean to signal if we should delete the pod or not. A better name for this method would be then `shouldDeletePod`.

Probably the same could be applied to the chain of methods that call `deletePod'

styx-scheduler-service/src/test/java/com/spotify/styx/ScheduledExecutionUtilTest.java
```diff
@@ -0,0 +1,75 @@
...
+      SCHEDULER.tick(1, SECONDS);
+    }
+
+    // Calculate timing statistics
+    var statistics = IntStream.range(0, executionTimes.size() - 1)
+        .map(i -> executionTimes.get(i + 1) - executionTimes.get(i))
+        .summaryStatistics();
+
+    assertThat(statistics.getAverage(), is(closeTo(10, 3)));
+    assertThat(statistics.getMin(), is(both(greaterThanOrEqualTo(5)).and(lessThanOrEqualTo(10))));
```
narape: What about making the 0.5 and 1.5 constants for lower and upper bound package private so we can use them here when computing the expected lower and upper bounds? 

```Java
    var tickInSeconds = 10;
    scheduleWithJitter(runnable, SCHEDULER, Duration.ofSeconds(tickInSeconds));
...
    assertThat(statistics.getMin(), 
             is(both(greaterThanOrEqualTo(ScheduledExecutionUtil.LOWER_BOUND_FACTOR*tickInSeconds))
                .and(lessThanOrEqualTo(tickInSeconds))));
```

Right now I had to come back to the `ScheduledExecutionUtil` because I couldn't understand why 5 and 15 was used.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -608,14 +611,21 @@ synchronized void tryPollPods() {
...
         // The pod is stale as we fetched the active states _after_ listing all pods.
-        cleanupWithoutRunState(workflowInstance.get(), pod);
+        deletePod = cleanupWithoutRunState(workflowInstance.get(), pod);
+      }
+      if (deletePod) {
+        podsToDelete.add(pod.getMetadata().getName());
       }
     }
+
+    // Do blocking pod deletion after emitting events in order to avoid run states going stale
```
narape: Sorry @danielnorberg, I couldn't understand this comment. 
Can you explain why delaying the pod deletes to the end after the emitting the events avoid run states going stale?

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -299,45 +275,53 @@ private void dequeueInstance0(final StyxConfig config, Map<String, Resource> res
...
-        blocker = gate.executionBlocker(instanceState.workflowInstance())
-            .toCompletableFuture().get(10, TimeUnit.SECONDS);
-      } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
-        throw new RuntimeException(e);
-      } catch (ExecutionException | TimeoutException e) {
-        LOG.warn("Failed to check execution blocker for {}, assuming there is no blocker",
-            instanceState.workflowInstance(), e);
-      }
+    if (executionIsBlocked(config, instance, runState)) {
```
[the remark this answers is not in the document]
danielnorberg: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: 👍

styx-scheduler-service/src/test/java/com/spotify/styx/ScheduledExecutionUtilTest.java
danielnorberg: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: There is a comment about this above, is it sufficient or do we need additional documentation?

In general, when attempting to persist a run state change we rely on the `counter` to catch when the state has already been changed e.g. by another scheduler. So having read a run state from datastore, the likelihood that the state will have been changed (and the local copy is stale) increases as time passes. Therefore it is preferable to act on the run state as soon as possibly after it has been read.

If haven't  already, we should document this mechanism in more detail, but that is outside the scope of this PR.

danielnorberg:
Rebased this on #711

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
narape: Got it

PR #703, opened 2019-03-14
auth: verify service account id token audience

danielnorberg:

## Description
Verify service account ID token audience when authenticating API requests.

## Motivation and Context
Ensure that ID tokens were intended for Styx and not some other API. This is an opt-in behavior that
service account clients can enable by specifying an audience when creating ID tokens.

Note: This only checks service account ID tokens, not user account ID tokens. That would require changes both to client-side code and require users to explicitly log on via OAuth. We might want that at some point but let's tackle that later.

## Checklist for PR author(s)

- [ ] All tests pass


honnix:
Will this require user side change once enabled?

danielnorberg:
@honnix No, because this PR only enables audience checking for service account ID tokens, and all our clients already target Styx. However, if in the future we want to also validate audience for user accounts we would have to make client changes. There is also a comment about this here https://github.com/spotify/styx/pull/703/files#diff-2804c4fb3a2336dbdfbbf77afc5232f1R160

danielnorberg:
@honnix Scratch that, seems our internal `styx-deploy` script doesn't set target audience 😩

danielnorberg:
Changed this PR to allow ID tokens with no audience (as sent by `styx-deploy`).

styx-service-common/src/main/java/com/spotify/styx/api/Authenticator.java
```diff
@@ -153,6 +157,17 @@ GoogleIdToken authenticate(String token) {
...
     }
 
+    // TODO: Also verify audience for user tokens. Currently this would require changing the auth flow in styx
+    //  clients and make users explicitly "log in" to Styx via the OAuth consent screen.
+    // Verify that this ID token was intended for Styx.
+    if (!allowedAudiences.isEmpty()
+        // TODO: Remove this null check and require tokens to have a target audience
+        && googleIdToken.getPayload().getAudience() != null) {
+      if (!googleIdToken.verifyAudience(allowedAudiences)) {
+        return null;
```
narape: Can we log something here before returning? like we do on lines 126, 136, 143,147 and 175.

We don't log any before returning on lines 152, 157, 167, 181 and 184. So it is inconsistent right now but logging here could help us to identify why we are rejecting the authentication

PR #715, opened 2019-04-01
Use guava cache

danielnorberg:

## Description
Use guava cache instead of rolling our own caching.

## Motivation and Context
* Avoid having to maintain our own implementation.
* Gain useful features, like async refresh.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/CachedSupplier.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/CachedSupplier.java
@@ -44,7 +44,7 @@ public class CachedSupplier<T> implements Supplier<T> {
 
   CachedSupplier(Try.CheckedSupplier<T> delegate, Time time, Duration timeout) {
     this.cache = CacheBuilder.newBuilder()
-        .expireAfterAccess(timeout)
+        .expireAfterWrite(timeout)
         .ticker(new Ticker() {
           @Override
           public long read() {
--- a/styx-common/src/main/java/com/spotify/styx/util/CachedSupplier.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/CachedSupplier.java
@@ -24,7 +24,11 @@ import com.google.common.base.Ticker;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
 import java.time.Duration;
+import java.util.concurrent.Executors;
 import java.util.function.Supplier;
 import javaslang.control.Try;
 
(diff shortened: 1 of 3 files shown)


styx-common/src/main/java/com/spotify/styx/util/CachedSupplier.java
```diff
@@ -20,91 +20,41 @@
...
-  }
-
-  private boolean timedOut() {
-    return time.get().toEpochMilli() - cacheTime > timeoutMillis;
-  }
-
-  @FunctionalInterface
-  public interface ThrowingSupplier<T, E extends Exception> {
-    T get() throws E;
+    return Try.of(() -> cache.get(Boolean.TRUE)).get();
```
honnix: This changes existing behaviour I think. Somewhere might get unchecked exception.

danielnorberg: Any exception from `get()` was already unchecked though.

honnix: I mean if it failed to refresh the value it fell back to old value if not null.

danielnorberg: Ah. Yeah now this PR also returns the old value if refresh fails.

styx-common/src/test/java/com/spotify/styx/util/CachedSupplierTest.java
```diff
@@ -20,99 +20,196 @@
...
-    x = 100;
-    int b = sut.get();
+  public void testCachesCalls() throws Throwable {
+    // Initial load
+    var a = sut.get();
+    assertThat(a, is(INITIAL_VALUE));
+    verify(delegate, times(1)).get();
+
+    // Make underlying supplier return new value
+    lenient().when(delegate.get()).thenReturn(NEW_VALUE);
```
honnix: Is `lenient()` needed here?

styx-common/src/test/java/com/spotify/styx/util/CachedSupplierTest.java
```diff
@@ -20,99 +20,196 @@
...
+  public void testCacheDoesNotRefreshBeforeTimeout() throws Throwable {
+    // Initial load
     int a = sut.get();
-    x = 100;
-    instant = Instant.parse("2016-10-17T15:00:09Z");
+    assertThat(a, is(INITIAL_VALUE));
+    verify(delegate, times(1)).get();
+
+    // Make supplier return new value and progress time, but not enough to trigger refresh
+    lenient().when(delegate.get()).thenReturn(NEW_VALUE);
```
honnix: And here.

styx-common/src/test/java/com/spotify/styx/util/CachedSupplierTest.java
danielnorberg: Yeah, otherwise mockito complains.

styx-common/src/test/java/com/spotify/styx/util/CachedSupplierTest.java
danielnorberg: Same. And removing these makes the tests more fragile in the face of future changes.

PR #726, opened 2019-04-04
silence state transition conflict spam + fix scheduler error handling

danielnorberg:

## Description
Some fixes for https://github.com/spotify/styx/pull/714

* Throw a new specialized Exception, `StateTransitionConflictException` on state transition failure due to transaction conflict or stale events (concurrent transition conflict). 
* Swallow this exception in periodically invoked tick-style call-sites.

* Handle processing errors in the scheduler tick.

## Motivation and Context
* There is too much conflict spam in the logs.
* Any failure during the scheduler tick is causing the tick duration to not be reported.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -46,6 +46,7 @@ import com.spotify.styx.monitoring.Stats;
 import com.spotify.styx.state.Message;
 import com.spotify.styx.state.RunState;
 import com.spotify.styx.state.StateManager;
+import com.spotify.styx.state.StateTransitionConflictException;
 import com.spotify.styx.state.Trigger;
 import com.spotify.styx.util.Debug;
 import com.spotify.styx.util.EventUtil;
@@ -628,9 +629,11 @@ class KubernetesDockerRunner implements DockerRunner {
         // TODO: spoofing counter values like this can give unexpected results, e.g. if we emit two events here the
         // first one might be discarded and the second one accepted.
         stateManager.receive(event, runState.counter() + i);
-      } catch (IsClosedException isClosedException) {
-        LOG.warn("Could not receive kubernetes event", isClosedException);
-        throw new RuntimeException(isClosedException);
+      } catch (StateTransitionConflictException e) {
+        LOG.debug("State transition conflict on kubernetes pod event: {}", event, e);
+        return;
+      } catch (IsClosedException ignore) {
+        return;
       }
     }
   }
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/state/PersistentStateManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/state/PersistentStateManager.java
@@ -134,12 +134,18 @@ public class PersistentStateManager implements StateManager {
   private void tickInstance(WorkflowInstance instance) {
     try {
       var stateOpt = storage.readActiveState(instance);
-      stateOpt.ifPresent(state -> {
-        LOG.info("Ticking instance: {}: #{} {}", instance, state.counter(), state.state());
-        outputHandler.transitionInto(state);
-      });
+      stateOpt.ifPresent(state -> tickInstance(instance, state));
     } catch (Exception e) {
       LOG.error("Error ticking instance: {}", instance, e);
+    }
+  }
+
+  private void tickInstance(WorkflowInstance instance, RunState state) {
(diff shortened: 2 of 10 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -628,9 +629,11 @@ private void emitPodEvents(Pod pod, RunState runState) {
         // TODO: spoofing counter values like this can give unexpected results, e.g. if we emit two events here the
         // first one might be discarded and the second one accepted.
         stateManager.receive(event, runState.counter() + i);
-      } catch (IsClosedException isClosedException) {
-        LOG.warn("Could not receive kubernetes event", isClosedException);
-        throw new RuntimeException(isClosedException);
+      } catch (StateTransitionConflictException e) {
+        LOG.debug("State transition conflict on kubernetes pod event: {}", event, e);
+        return;
+      } catch (IsClosedException ignore) {
```
honnix: We can use `receiveIgnoreClosed` if we don't care this exception.

styx-scheduler-service/src/main/java/com/spotify/styx/state/PersistentStateManager.java
```diff
@@ -353,7 +359,7 @@ private void verifyCounter(Event event, long expectedCounter, RunState currentRu
                              + expectedCounter  + " but current counter is "
                              + currentCounter + ". Discarding event " + event;
       log.debug(message);
-      throw new StaleEventException(message);
+      throw new StateTransitionConflictException(message);
```
honnix: Is it true this can only be caused by concurrent transition?

styx-scheduler-service/src/main/java/com/spotify/styx/state/PersistentStateManager.java
```diff
@@ -134,12 +134,18 @@ public void tick() {
...
-        outputHandler.transitionInto(state);
-      });
+      stateOpt.ifPresent(state -> tickInstance(instance, state));
     } catch (Exception e) {
-      LOG.error("Error ticking instance: {}", instance, e);
+      log.error("Error ticking instance: {}", instance, e);
+    }
+  }
+
+  private void tickInstance(WorkflowInstance instance, RunState state) {
```
honnix: `instance` is part of `RunState`, so we don't need that argument.

styx-scheduler-service/src/main/java/com/spotify/styx/state/PersistentStateManager.java
honnix: It could also be caused by retries without fetching the latest counter. But anyway we are ignoring this exception.

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: I wanted to exit the loop if we got IsClosedException, but I guess that's not really necessary.

styx-scheduler-service/src/main/java/com/spotify/styx/state/PersistentStateManager.java
danielnorberg: Yes, this could mask incorrect invocation of `receive` with an _incorrect_ counter. And we want to ignore the (hopefully) majority of cases when `receive` was invoked with a correct (although possibly stale due to concurrent transitions) counter. Here's a few ways to address that:

1. Stop using counters and wrap all pairs of read + update run-state operations into a single transaction. This seems "right" to me, and is what one would normally do when implementing transactional state processing. Maybe the reason for us using counters has become "stale" (pun intended).

2. Minimize the number of `receive` call-sites by only having them in the PSM tick. This entails making `OutputHandlers` return the events they want to emit instead of calling the state manager directly.

These are a bit outside the scope of this PR though.

styx-scheduler-service/src/main/java/com/spotify/styx/state/PersistentStateManager.java
danielnorberg: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
danielnorberg: Changing this to use `receiveIgnoreClosed` requires changing test cases as well, so opting to keep this as-is.

PR #734, opened 2019-04-09
kdr: remove redundant watch reconnect

danielnorberg:

## Description
kdr: remove redundant watch reconnect

## Motivation and Context
The fabric8 watcher handles reconnects on its own, so we were leaking
connections by creating a new watch on each disconnect.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -100,7 +100,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.TimeUnit;
 
 /**
  * A {@link DockerRunner} implementation that submits container executions to a Kubernetes cluster.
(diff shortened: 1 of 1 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -727,26 +726,9 @@ private void processPodUpdate(String podName, WorkflowInstance instance) {
...
-      }
-    }
-
-    private void scheduleReconnect() {
-      scheduledExecutor.schedule(this::reconnect, RECONNECT_DELAY_SECONDS, TimeUnit.SECONDS);
-    }
-
     @Override
     public void onClose(KubernetesClientException e) {
       LOG.warn("Watch closed", e);
```
honnix: In case of `HTTP_GONE`, fabric8 will issue a forced close and we will never get reconnected. :(

```
if (status.getCode() == HTTP_GONE) {
  webSocketRef.set(null);
  closeEvent(new KubernetesClientException(status));
  close(); // here everything is shutdown
  return;
}

@Override
  public void close() {
    logger.debug("Force closing the watch {}", this);
    closeEvent(null);
    closeWebSocket(webSocketRef.getAndSet(null));
    if (!executor.isShutdown()) {
      try {
        executor.shutdown();
        if (!executor.awaitTermination(1, TimeUnit.SECONDS)) {
          logger.warn("Executor didn't terminate in time after shutdown in close(), killing it in: {}", this);
          executor.shutdownNow();
        }
      } catch (Throwable t) {
        throw KubernetesClientException.launderThrowable(t);
      }
    }
  }
```

Nothing critical since we have a fallback polling. I guess we can reconnect ourselves in case of `HTTP_GONE` (this happened when upgrading GKE master node).

danielnorberg:
Submit a fix upstream to reconnect on http gone?

honnix:
Yeah, will first ask why. It seems to be intended behavior according to the code.

danielnorberg:
Bad assumption? Intents can also have bugs :)

honnix:
https://github.com/fabric8io/kubernetes-client/issues/1529 :(

Will try it again.

PR #736, opened 2019-04-10
Always dual write Workflows with new and old keys but only read from new

narape:

## Description
As part of the flattening of the workflows keys to reduce transaction contention of big components this PR removes the environmental variable to conditionally write to new flat key because now we dual write by default.

Also change the read logic to only read from new keys
## Checklist for PR author(s)

- [x] All tests pass


styx-service-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -657,7 +654,7 @@ public WorkflowState workflowState(WorkflowId workflowId) throws IOException {
         try {
           Thread.sleep(retryBaseDelay.toMillis());
         } catch (InterruptedException e1) {
-          throw Throwables.propagate(e1);
+          throw new RuntimeException(e1);
```
danielnorberg: 👍

PR #740, opened 2019-04-10
datastore: retry transactions

danielnorberg:

## Description
datastore: retry transactions

## Checklist for PR author(s)

- [x] All tests pass


styx-service-common/src/main/java/com/spotify/styx/storage/CheckedDatastore.java
```diff
@@ -70,4 +71,23 @@ CheckedDatastoreTransaction newTransaction() throws DatastoreIOException {
   Key allocateId(IncompleteKey newKey) throws IOException {
     return call(() -> datastore.allocateId(newKey));
   }
+
+  /**
+   * @see Datastore#runInTransaction(Datastore.TransactionCallable)
+   * @throws DatastoreIOException if the function or the underlying client throws any exception.
+   */
+  <T> T runInTransaction(Datastore.TransactionCallable<T> f) throws DatastoreIOException {
```
honnix: remove unused method.

styx-service-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -178,7 +179,7 @@
 
   public static final int ACTIVE_WORKFLOW_INSTANCE_INDEX_SHARDS = 128;
 
-  public static final int MAX_RETRIES = 100;
+  public static final int MAX_RETRIES = 6;
```
honnix: Not used constant.

styx-service-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -969,32 +934,67 @@ static StringValue jsonValue(Object o) throws JsonProcessingException {
...
     }
-    return storageTransactionFactory.apply(transaction);
+  }
+
+  private boolean isRetryableTransactionException(DatastoreException e) {
+    return
+        // code 10 ABORTED means that the transaction was aborted due to a conflict and can be retried.
+        e.getCode() == 10
+        || "ABORTED".equals(e.getReason())
+        // The datastore client has a retry bug where it incorrectly retries ABORTED lookup operations
```
honnix: add a todo to remind us that we should remove this check after datastore client fixes the bug.

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -809,82 +809,10 @@ public void runInTransactionShouldCallFunctionAndRollbackOnFailure() throws Exce
     }
 
     verify(transactionFunction).apply(storageTransaction);
-    verify(storageTransaction, never()).commit();
-    verify(storageTransaction).rollback();
   }
 
   @Test
-  public void runInTransactionShouldCallFunctionAndRollbackOnPreCommitConflict() throws Exception {
```
honnix: Make sure the rollback/commit thing are tested by some other test cases, if not let's add some.

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTransactionTest.java
```diff
@@ -104,175 +107,90 @@ public void tearDown() throws Exception {
...
+
+    var workflow = TestData.WORKFLOW_WITH_RESOURCES;
+
+    // Store workflow
+    storage.runInTransactionWithRetries(tx -> {
+      tx.store(workflow);
+      return null;
+    });
+
+    // Start a losing transaction that reads, waits for barrier and then stores the workflow
```
honnix: Comment more how they actually compete and the result is the first one will be executed twice.

styx-service-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: 👍

styx-service-common/src/main/java/com/spotify/styx/storage/CheckedDatastore.java
danielnorberg: 👍

styx-service-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
danielnorberg: 👍

styx-service-common/src/main/java/com/spotify/styx/util/Connections.java
```diff
@@ -57,15 +57,6 @@ public static InstrumentedDatastore createDatastore(Config config, Stats stats)
...
-        // Disable retries as the datastore client (at the time of writing) has a retry bug
-        // where it incorrectly retries ABORTED lookup operations in transactions without
-        // restarting the transaction. This results in 3 INVALID_ARGUMENT "transaction closed" errors
-        // that pollute our warning/error logs.
-        // We have our own retry implementation in DatastoreStorage#storeWithRetries.
-        // This has the downside that retryable transient RPC errors inside transactions will now
-        // cause transactions to be rolled back and retried, which is a more expensive operation than
-        // immediately retrying the RPC.
-        // TODO: Use the datastore client retry mechanism instead when the above bug has been fixed
         .setRetrySettings(ServiceOptions.getNoRetrySettings())
```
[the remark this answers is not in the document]
danielnorberg: 👍

styx-service-common/src/main/java/com/spotify/styx/storage/StorageTransaction.java
```diff
@@ -61,8 +61,9 @@ WorkflowId storeWorkflowWithNextNaturalTrigger(Workflow workflow, TriggerInstant
    * Deletes a workflow.
    *
    * @param workflowId The workflow to delete
+   * @return
```
honnix: add doc for what is returned. thanks for the fix!

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -410,7 +410,7 @@ private static boolean isActiveState(RunStateData runStateData) {
 
     final Backfill backfill;
     try {
-      backfill = storage.runInTransaction(tx -> {
+      backfill = storage.runInTransactionWithRetries(tx -> {
```
honnix: Not sure we want to retry on conflict here. We might end up writing stale data. If there is another transaction won, this one can give up.

styx-api-service/src/main/java/com/spotify/styx/api/workflow/WorkflowInitializer.java
```diff
@@ -50,7 +50,7 @@ public WorkflowInitializer(Storage storage, Time time) {
   public Optional<Workflow> store(Workflow workflow, Consumer<Optional<Workflow>> guard)
       throws WorkflowInitializationException {
     try {
-      return storage.runInTransaction(tx -> store(tx, workflow, guard));
+      return storage.runInTransactionWithRetries(tx -> store(tx, workflow, guard));
```
honnix: Not sure we want to retry on conflict here. We might end up writing stale data. If there is another transaction won, this one can give up.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -168,7 +168,7 @@ private void triggerAndProgress0(Backfill backfill) {
     final Instant initialNextTrigger = backfill.nextTrigger();
     while (true) {
       try {
-        if (!storage.runInTransaction(tx ->
+        if (!storage.runInTransactionWithRetries(tx ->
```
honnix: Not sure of this retry either. It should be OK to give up on this backfill and move to the next one.

styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
danielnorberg: Retrying makes sense here and will not cause the writing of "stale" data. Imagine that the "losing" request had arrived some milliseconds later and the previous "winning" transaction had already completed. Then there is no transaction conflict, and the latter request completes as well. That is fine and expected. Retrying transactions here is in effect "delaying" the losing transaction so it ends up ordered after the winning transaction. This is in practice not distinguishable from naturally arriving late.

From the user perspective, there are no ordering guarantees on concurrent requests. This is normal and to be expected in the vast majority of APIs and services. The only way for a user to guarantee an ordering between request A and B against a typical CRUD API like styx is to _wait_ for response A before issuing request B. If the user issues request A and B concurrently without waiting for either response, then the result can be that either A or B "wins". This is normal and expected for most APIs.

styx-api-service/src/main/java/com/spotify/styx/api/workflow/WorkflowInitializer.java
danielnorberg: Same

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
danielnorberg: Same. I don't believe there is much benefit in having a non-retrying flavor of `runInTransaction`.
Consider the rationale for the google provided `Datastore.runInTransaction` method retrying conflicts by default. This is a common pattern in transaction storage engine clients.

danielnorberg:
Will work a bit on test coverage.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
honnix: I see your point, and I think there is a difference between client level and application level retry because at application level we encode logic not only plain CRUD. We might need to be more cautious when doing retry.

honnix: I'm more worried about this transaction code is not idempotent, but maybe it is.

danielnorberg: After having worked with this code a bit, I now believe its behavior to be correct, although not consistent if there are concurrent updates.

I have refactored the backfill trigger manager code in PR #744 in order to (hopefully) make it easier to understand and give more consistent behavior when there are concurrent updates.

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTransactionTest.java
```diff
@@ -94,185 +114,230 @@ public static void tearDownClass() throws Exception {
...
+    executor.shutdownNow();
   }
 
   @Test
-  public void shouldCommitEmptyTransaction() throws IOException {
-    final CheckedDatastoreTransaction transaction = datastore.newTransaction();
-    DatastoreStorageTransaction storageTransaction = new DatastoreStorageTransaction(transaction);
-    storageTransaction.commit();
-    assertFalse(transaction.isActive());
+  public void shouldRetryOnConflict() throws Exception {
```
NatashaL: ⭐️

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTransactionTest.java
```diff
@@ -94,185 +114,230 @@ public static void tearDownClass() throws Exception {
...
+    });
+
+    // Start a losing transaction that reads, waits for barrier and then stores the workflow
+    var runs = new AtomicInteger();
+    var waiting = new LinkedBlockingQueue<Boolean>();
+    var proceed = new LinkedBlockingQueue<Boolean>();
+    var future = executor.submit(() -> storage.runInTransactionWithRetries(tx -> {
+      runs.incrementAndGet();
+      var wf = tx.workflow(workflow.id());
+      waiting.put(true);
```
NatashaL: 🎂

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTransactionTest.java
```diff
@@ -94,185 +114,230 @@ public static void tearDownClass() throws Exception {
...
+    var running = new CompletableFuture<String>();
+    var future = executor.submit(() -> {
+      running.complete(null);
+      DatastoreStorage.sleepMillis(Long.MAX_VALUE);
+      return "foobar";
+    });
+    running.join();
+    executor.shutdownNow();
+    var result = future.get(30, SECONDS);
+    assertThat(result, is("foobar"));
```
NatashaL: Interesting, I would have expected it to fail with an Interrupted exception

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -243,8 +248,13 @@ public static void tearDownClass() {
   @Before
   public void setUp() {
     MockitoAnnotations.initMocks(this);
-    datastore = spy(new CheckedDatastore(helper.getOptions().getService()));
-    storage = new DatastoreStorage(datastore, Duration.ZERO, DatastoreStorageTransaction::new, executor);
+    datastoreClient = helper.getOptions().toBuilder()
+        .setRetrySettings(ServiceOptions.getDefaultRetrySettings())
```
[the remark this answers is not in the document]
danielnorberg: Retries are disabled by default when using the local datastore emulator helper.

styx-service-common/src/main/java/com/spotify/styx/storage/TransactionException.java
```diff
@@ -29,44 +30,42 @@ public TransactionException(DatastoreException cause) {
           ", code=" + cause.getCode() +
           ", reason=" + cause.getReason() +
           ", isRetryable=" + cause.isRetryable()
-        , cause);
+        , Objects.requireNonNull(cause));
   }
 
-  // TODO: represent the failure cause using an enum instead
+  @Override
+  public synchronized DatastoreException getCause() {
```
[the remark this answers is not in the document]
danielnorberg: That's the signature of the method in the parent class.

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTransactionTest.java
danielnorberg: The interrupted exception is handled by the `sleepMillis` method: https://github.com/spotify/styx/pull/740/files/d929a7b0e4bee61159efdb673079dededc3cde6c#diff-489bc1d178eed213e99db37d2423a206R993

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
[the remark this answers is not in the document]
NatashaL: Interesting, I could verify that indeed. But even with using the NoRetrySettings all tests pass..

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -768,159 +796,140 @@ private Workflow workflow(WorkflowId workflowId) {
...
 
   @Test
-  public void runInTransactionShouldCallFunctionAndRollbackOnPreCommitConflict() throws Exception {
-    final DatastoreStorage storage = new DatastoreStorage(datastore, Duration.ZERO, storageTransactionFactory, executor);
-    final CheckedDatastoreTransaction transaction = datastore.newTransaction();
-    final DatastoreStorageTransaction storageTransaction = spy(new DatastoreStorageTransaction(transaction));
-    when(storageTransactionFactory.apply(any())).thenReturn(storageTransaction);
-
-    final Exception expectedException = new DatastoreException(10, "", "");
+  public void runInTransactionWithRetriesShouldRetry() throws Exception {
```
NatashaL: It doesn't look like we actually verify that a retry happened, or am I missing something?

NatashaL: I would guess this test is causing my confusion. The name looks like it would use the retry settings, but the implementation doesn't.

danielnorberg: Good catch. Currently the retry tests were in `DatastoreStorageTransactionTest.java`. Moving  them to `DatastoreStorageTest.java`.

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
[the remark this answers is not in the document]
danielnorberg: Moved the retry tests to this class, so now NoRetrySettings won't pass.

PR #744, opened 2019-04-19
refactor backfill triggering

danielnorberg:

## Description
Refactor backfill trigger manager to:
* Make code easier to understand
* Give more consistent behavior when there are concurrent updates

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
@@ -167,7 +167,7 @@ class BackfillTriggerManager {
           triggerNextInstanceAndProgress(tx, backfill.id(), workflow, activeInstances));
 
       if (!shouldContinue) {
-        return;
+        break;
       }
     }
   }
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
@@ -50,6 +50,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Consumer;
 import javaslang.control.Try;
 import org.slf4j.Logger;
@@ -147,6 +148,13 @@ class BackfillTriggerManager {
     // Note: getActiveStatesByTriggerId is eventually consistent and may not return all triggered instances.
     var alreadyActiveInstances = stateManager.getActiveStatesByTriggerId(backfill.id()).size();
 
+    // Look up the workflow. Halts the backfill if the workflow does not exist.
+    var workflowOpt = readBackfillWorkflowOrHalt(backfill.id());
+    if (workflowOpt.isEmpty()) {
+      return;
+    }
+    var workflow = workflowOpt.orElseThrow();
+
     // Trigger a limited number of instances. Limit here to avoid starvation (if this is a big backfill)
     // and to limit the risk of alreadyActiveInstances going stale (because of concurrent triggering) while
     // triggering instances.
@@ -156,7 +164,7 @@ class BackfillTriggerManager {
 
       // Attempt to trigger an instance
       var shouldContinue = storage.runInTransactionWithRetries(tx ->
-          triggerNextInstanceAndProgress(tx, backfill.id(), activeInstances));
+          triggerNextInstanceAndProgress(tx, backfill.id(), workflow, activeInstances));
 
       if (!shouldContinue) {
         break;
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -135,85 +139,95 @@ void tick0() {
   private void triggerAndProgress(Backfill backfill) {
     // Do not include all backfill spans in parent tick span to avoid it growing too big
     tracer.spanBuilderWithExplicitParent("Styx.BackfillTriggerManager.triggerAndProgress", null)
-        .startSpanAndRun(() -> triggerAndProgress0(backfill));
+        .startSpanAndRun(() -> Try.run(() -> triggerAndProgress0(backfill)));
```
NatashaL: It looks to me like Try.run will swallow unchecked exceptions without logging them. How about just using `guard` like above, which will log any uncaught exceptions?

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -135,85 +139,95 @@ void tick0() {
...
+
+    // Look up the workflow. Halts the backfill if the workflow does not exist.
+    var workflowOpt = readBackfillWorkflowOrHalt(backfill.id());
+    if (workflowOpt .isEmpty()) {
       return;
     }
+    var workflow = workflowOpt.orElseThrow();
 
-    final Workflow workflow = workflowOpt.get();
+    for (int i = 0; i < MAX_INSTANCES_TICK_TRIGGER; i++) {
```
NatashaL: So you're capping the number of instances to run per backfill per tick. Is that to make sure that bigger backfills don't starve others? Any other reason?

NatashaL:
Also, really nice comments describing the code! Makes it much easier to read and understand the reasoning

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
```diff
@@ -135,85 +139,95 @@ void tick0() {
...
-    if (!workflowOpt.isPresent()) {
-      LOG.debug("workflow not found for backfill {}, halt it.", backfill);
-      storeBackfill(backfill.builder().halted(true).build());
+    // Check how many instances for this backfill are currently active
+    // Note: getActiveStatesByTriggerId is eventually consistent and may not return all triggered instances.
+    var alreadyActiveInstances = stateManager.getActiveStatesByTriggerId(backfill.id()).size();
+
+    // Look up the workflow. Halts the backfill if the workflow does not exist.
+    var workflowOpt = readBackfillWorkflowOrHalt(backfill.id());
+    if (workflowOpt .isEmpty()) {
```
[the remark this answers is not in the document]
danielnorberg: 😅

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
danielnorberg: `guard` doesn't allow checked exceptions

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
danielnorberg: Non-starvation is one reason. Another is that the `alreadyActiveInstances` might get stale.

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
danielnorberg: Now propagating the exception though

styx-scheduler-service/src/main/java/com/spotify/styx/BackfillTriggerManager.java
danielnorberg: Added a comment.

NatashaL:
I tried but couldn't spot why the test fails. @danielnorberg please take a look before merging. Other than that, it looks good 👍

danielnorberg:
The test fails because [reading the workflow in the backfill transaction](https://github.com/spotify/styx/pull/744/files#diff-7724660cca477426c555a62afcc34313R185) conflicts with the [instance triggering transaction](https://github.com/spotify/styx/pull/744/files#diff-7724660cca477426c555a62afcc34313R221). 

There is a [TODO about doing triggering in the backfill transaction](https://github.com/spotify/styx/pull/744/files#diff-7724660cca477426c555a62afcc34313R213), but I'm not sure I want to undertake that refactoring in this PR. For now the simplest thing to do is to just revert https://github.com/spotify/styx/pull/744/commits/ec2cd5558f1dfd54e7f58995262012afa7ad696a which moved the workflow reading into the backfill transaction.

PR #745, opened 2019-04-19
create backfills in transaction

danielnorberg:

## Description
create backfills in transaction

## Motivation and Context
Guarantees that the workflow exists when the backfill is created.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
@@ -60,6 +60,7 @@ import com.spotify.styx.serialization.Json;
 import com.spotify.styx.state.RunState;
 import com.spotify.styx.state.StateData;
 import com.spotify.styx.storage.Storage;
+import com.spotify.styx.storage.StorageTransaction;
 import com.spotify.styx.util.RandomGenerator;
 import com.spotify.styx.util.ReplayEvents;
 import com.spotify.styx.util.ResourceNotFoundException;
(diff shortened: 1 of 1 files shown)


styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -328,34 +328,37 @@ private static boolean isActiveState(RunStateData runStateData) {
...
-    if (error.isPresent()) {
-      return Response.forStatus(Status.BAD_REQUEST.withReasonPhrase(error.get()));
+    // Validate backfill & workflow
+    var validationError = validate(rc, input, workflow);
+    if (validationError.isPresent()) {
+      return Response.forStatus(Status.BAD_REQUEST.withReasonPhrase(validationError.get()));
     }
 
+    // Forbid backfill creation if there are already active instances in the backfill range
+    // TODO: Does this make sense?
```
zatine: To me it makes sense, otherwise users can't know which run will be the final one and might get unexpected results. Could we elaborate on why it wouldn't make sense?

honnix: We are already doing this: https://github.com/spotify/styx/pull/745/files#diff-0cec30db46b18b2c0132f746e3bad5b0R371

Or this refers to something else?

honnix: NVM, this is a question, not a real todo. And yes, to me it makes sense as well.

PR #746, opened 2019-04-24
only allow usage of whitelisted secret

honnix:

## Description

Only allow usage of whitelisted secrets, otherwise workflow creation/update/execution will fail.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/WorkflowValidator.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/WorkflowValidator.java
@@ -29,6 +29,8 @@ import java.time.ZonedDateTime;
 import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
+import java.util.Set;
 
 public class WorkflowValidator {
 
@@ -46,12 +48,15 @@ public class WorkflowValidator {
 
   private final DockerImageValidator dockerImageValidator;
   private final Duration maybeMaxRunningTimeout;
+  private final Set<String> secretWhitelist;
 
-  private WorkflowValidator(DockerImageValidator dockerImageValidator, Duration maybeMaxRunningTimeout) {
+  private WorkflowValidator(DockerImageValidator dockerImageValidator, Duration maybeMaxRunningTimeout,
+                            Set<String> secretWhitelist) {
     Preconditions.checkArgument(maybeMaxRunningTimeout == null || !maybeMaxRunningTimeout.isNegative(),
         "Max Running timeout should be positive");
-    this.dockerImageValidator = dockerImageValidator;
+    this.dockerImageValidator = Objects.requireNonNull(dockerImageValidator);
     this.maybeMaxRunningTimeout = maybeMaxRunningTimeout;
+    this.secretWhitelist = secretWhitelist;
   }
 
   public static Builder newBuilder(DockerImageValidator dockerImageValidator) {
@@ -133,6 +138,12 @@ public class WorkflowValidator {
       }
     });
 
+    cfg.secret().ifPresent(secret -> {
+      if (secretWhitelist != null && !secretWhitelist.contains(secret.name())) {
+        e.add("secret " + secret.name() + " is not whitelisted");
+      }
+    });
+
     return e;
   }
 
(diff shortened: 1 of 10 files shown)


styx-common/src/main/java/com/spotify/styx/util/WorkflowValidator.java
```diff
@@ -46,12 +48,15 @@
...
   private final Duration maybeMaxRunningTimeout;
+  private final Set<String> secretWhitelist;
 
-  private WorkflowValidator(DockerImageValidator dockerImageValidator, Duration maybeMaxRunningTimeout) {
+  private WorkflowValidator(DockerImageValidator dockerImageValidator, Duration maybeMaxRunningTimeout,
+                            Set<String> secretWhitelist) {
     Preconditions.checkArgument(maybeMaxRunningTimeout == null || !maybeMaxRunningTimeout.isNegative(),
         "Max Running timeout should be positive");
-    this.dockerImageValidator = dockerImageValidator;
+    this.dockerImageValidator = Objects.requireNonNull(dockerImageValidator);
```
narape: Can you add `requireNonNull` also for `secretWhitelist` to make it consistent and more error proof?

styx-common/src/main/java/com/spotify/styx/util/WorkflowValidator.java
```diff
@@ -133,6 +138,12 @@ public static Builder newBuilder(DockerImageValidator dockerImageValidator) {
       }
     });
 
+    cfg.secret().ifPresent(secret -> {
+      if (secretWhitelist != null && !secretWhitelist.contains(secret.name())) {
```
narape: I would rather initialize `secretWhitelist` to `Set.of()` (the empty set) in the `Builder` class so we can simplify this line to be just 

```java
if (!secretWhitelist.contains(secret.name())) {
```

styx-common/src/test/java/com/spotify/styx/util/WorkflowValidatorTest.java
```diff
@@ -247,6 +248,30 @@ public void shouldNotAllowComponentWithHash(String component) {
...
+    WorkflowValidator sut = WorkflowValidator.newBuilder(dockerImageValidator)
+        .withSecretWhitelist(Set.of(FULL_WORKFLOW_CONFIGURATION.secret().get().name()))
+        .build();
+
+    final List<String> errors = sut.validateWorkflow(
+        Workflow.create("test", FULL_WORKFLOW_CONFIGURATION));
+
+    assertThat(errors, empty());
+  }
+
```
narape: Can you add also a test named like `shouldPassUsageForEmptyWhitelist` to verify that we wont fail if there is no white list?

styx-common/src/main/java/com/spotify/styx/util/WorkflowValidator.java
honnix: Unfortunately this validator is used by CLIMain.java where we can not apply any whitelist check. So `null` means skipping whitelist check.

styx-common/src/main/java/com/spotify/styx/util/WorkflowValidator.java
honnix: Unfortunately this validator is used by CLIMain.java where we can not apply any whitelist check. So `null` means skipping whitelist check.

styx-common/src/main/java/com/spotify/styx/util/WorkflowValidator.java
honnix: Unless we say `Set.of("*")` is a wildcard.

styx-common/src/test/java/com/spotify/styx/util/WorkflowValidatorTest.java
honnix: I think this is covered by all the other test cases where we use `sut`.

styx-common/src/main/java/com/spotify/styx/util/WorkflowValidator.java
narape: Yes, I see

styx-common/src/main/java/com/spotify/styx/util/WorkflowValidator.java
narape: Wildcard will make it more messy. As we discussed earlier the problem with the nulls are that we are using the same class in the cli and the service. What we need is actually split this class and make the validators composable but that is for another PR

styx-common/src/test/java/com/spotify/styx/util/WorkflowValidatorTest.java
narape: Agree

PR #747, opened 2019-04-25
refactor workflow validator

honnix:

## Description

Split `WorkflowValidator` into basic and extra part.

Basic can be used by both client and server while extra can only be used by server.

## Motivation and Context

Action based on `null` check is hard to follow and is error prone.

## Checklist for PR author(s)

- [ ] All tests pass


styx-common/src/main/java/com/spotify/styx/util/WorkflowValidator.java
```diff
@@ -20,168 +20,24 @@
...
-        e.add("secret " + secret.name() + " is not whitelisted");
-      }
-    });
-
-    return e;
-  }
-
-  private <T extends Comparable<T>> void lowerLimit(List<String> errors, T value, T limit, String message) {
-    limit(errors,value.compareTo(limit) < 0, value, limit, message);
+  default <T extends Comparable<T>> void lowerLimit(List<String> errors, T value, T limit, String message) {
```
narape: I see this methods as auxiliary methods and not part of the interface. So it would be better to have it as statics methods instead of default methods

styx-service-common/src/test/java/com/spotify/styx/util/ExtraWorkflowValidatorTest.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.List;
+import java.util.Set;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ExtraWorkflowValidatorTest {
```
narape: Add a test for making sure that the we are propagating a call to the delegate. For example that the validate method was called and that we propagate any error returned by the delegate

honnix:
@narape PTAL again. I have rebased from master.

PR #773, opened 2019-07-02
Add more error description

NatashaL:

## Motivation and Context

Errors like these are confusing for users:
```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `com.spotify.styx.model.WorkflowConfigurationBuilder$Value`, problem: schedule

and:
Caused by: java.lang.NullPointerException: schedule
```
 

Unit test

## Checklist for PR author(s)

- [x] All tests pass


styx-cli-unshaded/src/test/resources/com/spotify/styx/cli/bad-content.yaml
```diff
@@ -0,0 +1 @@
+null
```
tnsetting: Maybe we can add some valid configurations then something invalid.

PR #774, opened 2019-07-02
Get active states for multi components endpoint

tnsetting:

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/doc/api.apib
+++ b/doc/api.apib
@@ -678,7 +678,7 @@ Workflow Instance belonging to a Backfill entity.
 
 Resources related to inspect detailed status of Workflow Instances.
 
-## Active States [/{version}/status/activeStates{?component,workflow}]
+## Active States [/{version}/status/activeStates{?component,workflow,components}]
 
 + Parameters
     + version: `v3` (enum[string]) - API version
@@ -686,6 +686,8 @@ Resources related to inspect detailed status of Workflow Instances.
             + `v3`
     + component: `styx-canary` (optional, string) - Component ID
     + workflow: `StyxCanary`  (optional, string) - Workflow ID - requires Component ID to also be specified
+    + components: `styx-canary,styx-canary1,styx-canary3` (optional, string) - Component IDs, IDs should be comma
+    separated.
 
 ### Get Active States [GET]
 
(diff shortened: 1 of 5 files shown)


doc/api.apib
```diff
@@ -678,14 +678,16 @@ Workflow Instance belonging to a Backfill entity.
...
+## Active States [/{version}/status/activeStates{?component,workflow,components}]
 
 + Parameters
     + version: `v3` (enum[string]) - API version
         + Members
             + `v3`
     + component: `styx-canary` (optional, string) - Component ID
     + workflow: `StyxCanary`  (optional, string) - Workflow ID - requires Component ID to also be specified
+    + components: `styx-canary,styx-canary1,styx-canary3` (optional, string) - Component IDs, IDs should be comma
+    separated.
```
NatashaL: Let's also make it clear that if `components` is provided, the other two parameters will be ignored

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -139,6 +144,29 @@ private static String arg(String name, RequestContext rc) {
...
+
+    final List<Optional<IOException>> exceptions = components.parallelStream().map( componentId -> {
+      Optional<IOException> exception = Optional.empty();
+      try{
+        final Map<WorkflowInstance, RunState> stateMap = storage.readActiveStates(componentId);
+        for (WorkflowInstance instance : stateMap.keySet()) {
+          mapBuilder.put(instance, stateMap.get(instance));
+        }
+      } catch (IOException e) {
+        exception = Optional.of(e);
```
NatashaL: Instead of using a variable, we can just return `Optional.of(e)` here and return `Optional.empty()` below

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -139,6 +144,29 @@ private static String arg(String name, RequestContext rc) {
     return Response.forPayload(RunStateDataPayload.create(runStates));
   }
 
+  private Map<WorkflowInstance, RunState> getActiveStates(String componentsStr) throws IOException {
+    final List<String> components = Arrays.asList(componentsStr.split(","));
+    final ImmutableMap.Builder<WorkflowInstance, RunState> mapBuilder = ImmutableMap.builder();
+
+    final List<Optional<IOException>> exceptions = components.parallelStream().map( componentId -> {
```
honnix: I think we have some problems here:
* `parallelStream` for IO blocking operation is not good because it uses common FJP, [this](https://github.com/spotify/styx/blob/master/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java#L188) is a better way
* The map is not thread safe and we have multiple threads adding things to it

PR #781, opened 2019-07-25
No common FJP for BigtableStorage

honnix:

## Description

Use dedicated FJP for IO operation.

~Depends on #780~

## Motivation and Context

Using common FJP for IO operation is not a good practice. This change should also improve performance.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-service-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
+++ b/styx-service-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
@@ -61,6 +61,7 @@ public class AggregateStorage implements Storage {
 
   @Override
   public void close() throws IOException {
+    bigtableStorage.close();
     datastoreStorage.close();
   }
 
--- a/styx-service-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
+++ b/styx-service-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
@@ -117,6 +117,11 @@ public class BigtableStorage implements Closeable {
         .build();
   }
 
+  @Override
+  public void close() throws IOException {
+    closer.close();
+  }
+
   SortedSet<SequenceEvent> readEvents(WorkflowInstance workflowInstance) throws IOException {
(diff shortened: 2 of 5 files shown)


styx-service-common/src/main/java/com/spotify/styx/storage/BigtableStorage.java
```diff
@@ -220,16 +245,19 @@ WorkflowInstanceExecutionData executionData(WorkflowInstance workflowInstance) t
 
   private List<WorkflowInstanceExecutionData> executionData(
       Set<WorkflowInstance> workflowInstancesSet) {
-    return workflowInstancesSet.parallelStream()
-        .map(workflowInstance -> {
+    return workflowInstancesSet.stream()
+        .map(workflowInstance -> asyncIO(() -> {
```
narape: we don't need to catch the `IOException` because the lambda would be mapped to a `IOOperation` and those can throw `IOException`

styx-service-common/src/main/java/com/spotify/styx/storage/AggregateStorage.java
```diff
@@ -61,6 +61,7 @@ public AggregateStorage(Connection connection, Datastore datastore) {
 
   @Override
   public void close() throws IOException {
+    bigtableStorage.close();
```
narape: In `AggregateStorageTest` there is a test for `shouldCloseDatastore` but no for BigTable

styx-service-common/src/test/java/com/spotify/styx/storage/AggregateStorageTest.java
```diff
@@ -72,6 +72,12 @@ public void shouldCloseDatastore() throws IOException {
     verify(datastore).close();
   }
 
+  @Test
+  public void shouldCloseBigTable() throws IOException {
```
honnix: How about we merge those two test cases?

honnix:
Let's add PR to release note before merging it.

PR #789, opened 2019-08-08
Fix parsing of auth tocken with google cloud compute metadata

narape:

## Description
JWT with extra google cloud compute metadata triggers a parsing bug  in the google.http.client.lib  1.29.1.

```yaml
{
  azp: 103411466401044735393,
  email: some.email@project.iam.gserviceaccount.com,
  email_verified: true,
  exp: 1565270045,
  iat: 1565266445,
  iss: https://accounts.google.com,
  sub: 103411466401044735393,
  google: {
    compute_engine: {
      instance_creation_timestamp: 1556025719,
      instance_id: 5853837336805153889,
      instance_name: some-host,
      project_id: some-project,
      project_number: 213281561831,
      zone: europe-west1-d
    }
  }
}
```

In this example everything under `google` key is the "extra" metadata and causes a problem parsing the project number as a `int` (overflows).

Here we create a Unit Test to reproduce the problem and bump the library that reverts the problematic commit that caused the problem.

See https://github.com/googleapis/google-http-java-client/releases/tag/v1.29.2
and https://github.com/googleapis/google-http-java-client/pull/662
## Motivation and Context
See description

I have included unit tests.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-service-common/src/test/java/com/spotify/styx/api/GoogleIdTokenVerifierTest.java
+++ b/styx-service-common/src/test/java/com/spotify/styx/api/GoogleIdTokenVerifierTest.java
@@ -101,10 +101,10 @@ public class GoogleIdTokenVerifierTest {
     setField(keysManager, "expirationTimeMilliseconds", Long.MAX_VALUE);
   }
 
-  private void setField(GooglePublicKeysManager keysManager, String name, Object publicKey1)
+  private void setField(GooglePublicKeysManager keysManager, String name, Object value)
       throws NoSuchFieldException, IllegalAccessException {
     final var field = GooglePublicKeysManager.class.getDeclaredField(name);
     field.setAccessible(true);
-    field.set(keysManager, publicKey1);
+    field.set(keysManager, value);
   }
 }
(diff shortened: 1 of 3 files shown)


styx-service-common/src/test/java/com/spotify/styx/api/GoogleIdTokenVerifierTest.java
```diff
@@ -0,0 +1,110 @@
...
+   * Force use custom public key to avoid to request keys from Certificate Authorities.
+   */
+  private void stubPublicKey(GooglePublicKeysManager keysManager, PublicKey publicKey)
+      throws NoSuchFieldException, IllegalAccessException {
+    // Reflection is used because final methods make it impossible to mock
+    setField(keysManager, "publicKeys", List.of(publicKey));
+    setField(keysManager, "expirationTimeMilliseconds", Long.MAX_VALUE);
+  }
+
+  private void setField(GooglePublicKeysManager keysManager, String name, Object publicKey1)
```
tnsetting: publicKey1 does not really represent what it is as it can be public key or Long integer value. Maybe change it just change to fieldValue?

PR #794, opened 2019-09-27
Use a new thread pool for get active states

tnsetting:

This PR addresses issues connecting to PR: https://github.com/spotify/styx/pull/774
Now the getActiveStates endpoint will use a separate thread pool instead of the main FJP and it uses the `concurrentHashMap` instead to prevent race condition.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
@@ -21,11 +21,12 @@
 package com.spotify.styx.api;
 
 import static com.spotify.styx.api.Api.Version.V3;
+import static com.spotify.styx.util.CloserUtil.register;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
 
 import com.google.api.client.util.Lists;
-import com.google.common.collect.ImmutableMap;
+import com.google.common.io.Closer;
 import com.spotify.apollo.RequestContext;
 import com.spotify.apollo.Response;
 import com.spotify.apollo.Status;
@@ -49,6 +50,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ForkJoinTask;
 import java.util.stream.Stream;
 import okio.ByteString;
 import org.slf4j.Logger;
@@ -60,15 +64,19 @@ import org.slf4j.LoggerFactory;
 public class StatusResource {
 
   private static final Logger log = LoggerFactory.getLogger(StatusResource.class);
+  private static final int CONCURRENCY = 32;
 
   static final String BASE = "/status";
 
   private final Storage storage;
   private final ServiceAccountUsageAuthorizer accountUsageAuthorizer;
+  private final ForkJoinPool forkJoinPool;
+  private final Closer closer = Closer.create();
 
   public StatusResource(Storage storage, ServiceAccountUsageAuthorizer accountUsageAuthorizer) {
     this.storage = requireNonNull(storage);
     this.accountUsageAuthorizer = requireNonNull(accountUsageAuthorizer);
+    this.forkJoinPool = register(closer, new ForkJoinPool(CONCURRENCY), "status-resource");
   }
 
   public Stream<Route<AsyncHandler<Response<ByteString>>>> routes() {
@@ -129,7 +137,7 @@ public class StatusResource {
     final List<RunStateData> runStates = Lists.newArrayList();
     try {
       activeStates = componentsOpt.isPresent() ? getActiveStates(componentsOpt.get()):
-                       getActiveStates(componentOpt, workflowOpt);
+                     getActiveStates(componentOpt, workflowOpt);
 
     } catch (InvalidParametersException e) {
       return Response.forStatus(Status.BAD_REQUEST.withReasonPhrase(e.getMessage()));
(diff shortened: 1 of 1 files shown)


styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -129,7 +137,7 @@ private static String arg(String name, RequestContext rc) {
     final List<RunStateData> runStates = Lists.newArrayList();
     try {
       activeStates = componentsOpt.isPresent() ? getActiveStates(componentsOpt.get()):
-                       getActiveStates(componentOpt, workflowOpt);
+                     getActiveStates(componentOpt, workflowOpt);
```
honnix: In Styx source code we usually format ternary expression as:
```
componentsOpt.isPresent()
? getActiveStates(componentsOpt.get())
: getActiveStates(componentOpt, workflowOpt)
```

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -146,25 +154,36 @@ private static String arg(String name, RequestContext rc) {
...
-      return exception;
-      }).filter(Optional::isPresent).collect(toList());
+    final ConcurrentHashMap<WorkflowInstance, RunState> concurrentHashMap = new ConcurrentHashMap<>();
+
+    final List<Optional<IOException>> exceptions =
+        components.stream()
+            .map(
+                componentId ->
+                    forkJoinPool.submit(
+                        () -> {
```
honnix: If we move this line up, it might be easier to read.

honnix: We can refactor this a bit by having a method.

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -146,25 +154,36 @@ private static String arg(String name, RequestContext rc) {
...
-        final Map<WorkflowInstance, RunState> stateMap = storage.readActiveStates(componentId);
-        for (WorkflowInstance instance : stateMap.keySet()) {
-          mapBuilder.put(instance, stateMap.get(instance));
-        }
-      } catch (IOException e) {
-        exception = Optional.of(e);
-      }
-      return exception;
-      }).filter(Optional::isPresent).collect(toList());
+    final ConcurrentHashMap<WorkflowInstance, RunState> concurrentHashMap = new ConcurrentHashMap<>();
```
honnix: Let's use `ConcurrentMap` as type declaration, or `var` everything.

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -146,25 +154,36 @@ private static String arg(String name, RequestContext rc) {
...
+                          } catch (IOException e) {
+                            exception = Optional.of(e);
+                          }
+                          return exception;
+                        }))
+            .collect(toList())
+            .stream()
+            .map(ForkJoinTask::join)
+            .filter(Optional::isPresent)
+            .collect(toList());
```
honnix: We should be able to rewrite this logic to collect a map instead of a list of exceptions and then return the first one if not empty. So we can avoid creating a concurrent map.

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -146,25 +154,36 @@ private static String arg(String name, RequestContext rc) {
...
+                          try {
+                            final Map<WorkflowInstance, RunState> stateMap =
+                                storage.readActiveStates(componentId);
+                            for (WorkflowInstance instance : stateMap.keySet()) {
+                              concurrentHashMap.put(instance, stateMap.get(instance));
+                            }
+                          } catch (IOException e) {
+                            exception = Optional.of(e);
+                          }
+                          return exception;
```
honnix: `return Option.empty();` here so we can avoid having variable.

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -146,25 +154,36 @@ private static String arg(String name, RequestContext rc) {
...
+                        () -> {
+                          Optional<IOException> exception = Optional.empty();
+                          try {
+                            final Map<WorkflowInstance, RunState> stateMap =
+                                storage.readActiveStates(componentId);
+                            for (WorkflowInstance instance : stateMap.keySet()) {
+                              concurrentHashMap.put(instance, stateMap.get(instance));
+                            }
+                          } catch (IOException e) {
+                            exception = Optional.of(e);
```
honnix: `return Optional.of(e)`.

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
honnix: e.g. https://stackoverflow.com/questions/42490314/convert-list-of-maps-to-single-map-via-streams

To handle checked exception, we could use `Try` from javaslang, e.g. https://github.com/spotify/styx/blob/master/styx-scheduler-service/src/main/java/com/spotify/styx/state/PersistentStateManager.java#L122

honnix: Alternatively we could turn this thing around. Instead of building a map, we could build a list of exceptions and later on check whether the list is empty or not.

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -144,27 +154,28 @@ private static String arg(String name, RequestContext rc) {
...
-      }).filter(Optional::isPresent).collect(toList());
-    if (!exceptions.isEmpty()) {
-      throw exceptions.get(0).orElseThrow();
-    }
+  private Map<WorkflowInstance, RunState> getActiveStates(String componentsStr) {
+    var components = Arrays.asList(componentsStr.split(","));
+    var activeStatesOrExceptions =
+        components.stream()
+            .map( componentId -> forkJoinPool.submit(() -> {
+               return Try.of( () -> { return storage.readActiveStates(componentId); });
```
honnix: `() -> { return storage.readActiveStates(componentId);` can this be simplified to `() -> storage.readActiveStates(componentId)`

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -144,27 +154,28 @@ private static String arg(String name, RequestContext rc) {
...
-        exception = Optional.of(e);
-      }
-      return exception;
-      }).filter(Optional::isPresent).collect(toList());
-    if (!exceptions.isEmpty()) {
-      throw exceptions.get(0).orElseThrow();
-    }
+  private Map<WorkflowInstance, RunState> getActiveStates(String componentsStr) {
+    var components = Arrays.asList(componentsStr.split(","));
+    var activeStatesOrExceptions =
```
honnix: Not problem of this PR, but we recently did a change to simplify FJP usage: https://github.com/spotify/styx/pull/795/files#diff-0cec30db46b18b2c0132f746e3bad5b0R453

We can do the same here in this PR.

styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -60,15 +65,19 @@
 public class StatusResource {
```
honnix: Here we forgot to close the closer... which is pretty bad.

PR #800, opened 2019-10-07
handle duplicated components

honnix:

## Description

In case of duplicated component, existing code would fail to merge maps due to key conflict.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
@@ -47,6 +47,7 @@ import com.spotify.styx.storage.Storage;
 import java.io.Closeable;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


styx-api-service/src/main/java/com/spotify/styx/api/StatusResource.java
```diff
@@ -161,7 +162,7 @@ private static String arg(String name, RequestContext rc) {
   }
 
   private Map<WorkflowInstance, RunState> getActiveStates(String componentsStr) throws IOException {
-    var components = Arrays.asList(componentsStr.split(","));
+    var components = new HashSet<>(Arrays.asList(componentsStr.split(",")));
```
narape: A better way to do it is:

```java
var components = Set.of(componentsStr.split(","));
 ```

honnix: No this won't work unfortunately because `Set.of` fails on duplicates.

PR #803, opened 2019-10-18
halt workflow instance if disabled and naturally triggered

honnix:

## Description

If the workflow has been disabled between retry and it is naturally triggered, we halt it.

## Motivation and Context

The hypothesis is in this case users would most likely want the instance to stop.

**Note, this is a breaking change.**

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
@@ -116,7 +116,7 @@ public class ExecutionDescriptionHandler implements OutputHandler {
     );
 
     var isWorkflowEnabled = workflowWithState.state().enabled().orElse(false);
-    var isNaturalTrigger = data.trigger().map(t -> t.equals(Trigger.natural()).orElse(false);
+    var isNaturalTrigger = data.trigger().map(t -> t.equals(Trigger.natural())).orElse(false);
 
     if (!isWorkflowEnabled && isNaturalTrigger) {
       throw new MissingRequiredPropertyException(format("%s is disabled, halting %s", workflowId, workflowInstance));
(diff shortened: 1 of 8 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
```diff
@@ -104,19 +103,26 @@ public void transitionInto(RunState state) {
...
 
-    final Collection<String> errors = validator.validateWorkflow(workflow);
+    if (data.trigger().isPresent()) {
+      if (data.trigger().orElseThrow().equals(Trigger.natural()) &&
+          !workflowWithState.state().enabled().orElse(false)) {
+        throw new MissingRequiredPropertyException(format("%s is disabled, halting %s", workflowId, workflowInstance));
+      }
+    }
+
+    final Collection<String> errors = validator.validateWorkflow(workflowWithState.workflow());
```
narape: You have `var workflow = workflowWithState.workflow();` on line 111 already

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/ExecutionDescriptionHandler.java
```diff
@@ -104,19 +103,26 @@ public void transitionInto(RunState state) {
...
+
+    var dockerImage = workflow.configuration().dockerImage().orElseThrow(
+        () -> new MissingRequiredPropertyException(
+            format("%s has no docker image, halting %s", workflowId, workflowInstance))
     );
 
-    final Collection<String> errors = validator.validateWorkflow(workflow);
+    if (data.trigger().isPresent()) {
+      if (data.trigger().orElseThrow().equals(Trigger.natural()) &&
+          !workflowWithState.state().enabled().orElse(false)) {
```
narape: This is hard to follow. What about rewriting it in this way?

```java
var isWorkflowEnabled = workflowWithState.state().enabled().orElse(false);
var isNaturalTrigger = data.trigger().map(t -> t.equals(Trigger.natural()).orElse(false);

if (!isWorkflowEnabled && isNaturalTrigger) {
```

honnix:
Will ship this change in a separate release.

PR #810, opened 2019-11-20
Rewatch

honnix:

## Description

This reverts https://github.com/spotify/styx/pull/734

There are cases that fabric8 won't reconnect on itself, e.g. in case of `HTTP_GONE`.

There was [an attempted fix](https://github.com/fabric8io/kubernetes-client/pull/1800) in upstream, but eventually got rolled back because it broke the contract.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -157,6 +157,7 @@ class KubernetesDockerRunner implements DockerRunner {
   private final Duration podDeletionDelay;
   private final Time time;
   private final ExecutorService executor;
+  private Watch watch;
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -753,9 +755,33 @@ private void processPodUpdate(String podName, WorkflowInstance instance) {
       emitPodEvents(pod, runState.get());
     }
 
+    private void reconnect() {
+      LOG.info("Re-establishing pod watcher");
+
+      closeWatch();
```
honnix: This is not necessarily needed because fabric8 force closes it anyway. But to be on the safe side, we do it again and it is idempotent.

PR #783, opened 2019-07-31
flexible retry

honnix:
# Hey, I just made a Pull Request!

## Description

Support using a boolean expression to decide whether retry should be done or not.

e.g `exitCode == 1 && (tries < 3 || consecutiveFailures < 4) && triggerType == "natural"`

`50` as a special exit code still has higher priority than retry condition. Max number of retries are not affected by this feature.

## Motivation and Context

Users will be able to control when to stop retrying in a flexible way.

## Checklist for PR author(s)

- [x] Changes are covered by unit test
- [x] All tests pass
- [x] Code coverage check passes
- [x] Error handling is tested
- [x] Errors are handled at the appropriate layer
- [x] Errors that cannot be handled where they occur are propagated
- [ ] (optional) Changes are covered by system test
- [ ] Relevant documentation updated
- [x] This PR has NO breaking change to public API
- [ ] This PR has breaking change to public API and it is documented

## Checklist for PR reviewer(s)

- [ ] This PR has been incorporated in release note for the coming version
- [ ] Risky changes introduced by this PR have been all considered

PR #812, opened 2019-12-27
flexible retry alt

honnix:

## Description

Alternative of #783 

A [SpEL](https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#expressions) boolean expression.
If the expression evaluates to `false`, Styx will stop retrying and halt the workflow instance immediately. This
configuration has no impact on possible max number of tries, meaning it can only be used to halt workflow instance
earlier.

The following variables will be injected by Styx so that they can be used in the expression:
* \#exitCode: the exit code from the last execution
* \#tries: total number of tries, which equals to `1` when the first time a workflow instance gets executed and `2` when
  the first retry is issued
* \#consecutiveFailures: total number of consecutive failures that is not `missing dependency`
* \#triggerType: `natural`, `backfill` or `ad-hoc`

## Motivation and Context

Users want to have flexible retry condition. This PR exposes parameters impacting retry and gives users mechanism to have their own retry policy.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
@@ -32,6 +32,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.regex.Pattern;
+import org.springframework.expression.ParseException;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
 
 public class BasicWorkflowValidator implements WorkflowValidator {
 
@@ -43,6 +45,7 @@ public class BasicWorkflowValidator implements WorkflowValidator {
   private static final int MAX_SECRET_NAME_LENGTH = 253;
   private static final int MAX_SECRET_MOUNT_PATH_LENGTH = 1024;
   private static final int MAX_SERVICE_ACCOUNT_LENGTH = 256;
+  private static final int MAX_RETRY_CONDITION_LENGTH = 256;
   private static final int MAX_ENV_VARS = 128;
   private static final int MAX_ENV_SIZE = 16 * 1024;
   private static final Pattern VALID_EMAIL_ADDRESS_REGEX =
@@ -95,6 +98,8 @@ public class BasicWorkflowValidator implements WorkflowValidator {
     upperLimit(e, cfg.env().entrySet().stream()
             .mapToInt(entry -> entry.getKey().length() + entry.getValue().length()).sum(),
         MAX_ENV_SIZE, "env too big");
+    upperLimit(e, cfg.retryCondition().map(String::length).orElse(0),
+        MAX_RETRY_CONDITION_LENGTH, "retry condition too long");
 
     cfg.dockerImage().ifPresent(image ->
         dockerImageValidator.validateImageReference(image).stream()
@@ -132,11 +137,19 @@ public class BasicWorkflowValidator implements WorkflowValidator {
       }
     });
 
+    cfg.retryCondition().ifPresent(retryCondition -> {
+      try {
+        new SpelExpressionParser().parseRaw(retryCondition);
+      } catch (ParseException ex) {
+        e.add(format("invalid retry condition: %s", ex.getMessage()));
+      }
+    });
+
     return e;
   }
 
   private static boolean validateServiceAccount(String serviceAccount) {
-    var matcher = VALID_EMAIL_ADDRESS_REGEX .matcher(serviceAccount);
+    var matcher = VALID_EMAIL_ADDRESS_REGEX.matcher(serviceAccount);
     return matcher.matches();
   }
 }
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
@@ -136,7 +136,7 @@ public class TerminationHandler implements OutputHandler {
 
     var workflow = workflows.get().get(state.workflowInstance().workflowId());
     if (workflow == null) {
-      LOG.debug("Workflow {} does not exist possibly due to stale caching",
+      LOG.debug("Workflow {} does not exist possibly due to stale workflow cache",
           state.workflowInstance().workflowId().toKey());
       return false;
     }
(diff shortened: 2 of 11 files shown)


styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
```diff
@@ -134,6 +137,15 @@ public BasicWorkflowValidator(DockerImageValidator dockerImageValidator) {
       }
     });
 
+    cfg.retryCondition().ifPresent(retryCondition -> {
+      upperLimit(e, retryCondition.length(), MAX_RETRY_CONDITION_LENGTH, "retry condition too long");
```
NatashaL: If this `upperLimit` check generates an error, the rest of the body will run and the expression will still be parsed. I think we should skip parsing it in that case. Since it might be any arbitrary length I think we should avoid doing the effort of parsing it.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
```diff
@@ -96,7 +130,46 @@ private static boolean isMissingDependency(Optional<Integer> exitCode) {
...
+    try {
+      expression = retryConditionExpressionCache.get(retryCondition,
+          () -> expressionParser.parseExpression(retryCondition));
+    } catch (ExecutionException | UncheckedExecutionException e) {
+      LOG.debug("Failed to parse retry condition `{}`", retryCondition, e.getCause());
+      return false;
+    }
+
+    try {
+      return TRUE.equals(expression.getValue(context, Boolean.class));
```
NatashaL: `TRUE.equals` shouldn't be necessary, a Boolean.class type can be interpreted as a simple boolean already.

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/TerminationHandlerTest.java
```diff
@@ -126,13 +152,114 @@ public void shouldScheduleRetryOf10MinutesOnMissingDependencies() {
...
+  public void shouldStopOnRetryConditionNotMet() {
+    var data = data(1, 1.0, Optional.of(2));
+    var nonZeroTerm = RunState.create(WORKFLOW_INSTANCE, TERMINATED, data, NOW, COUNTER);
+    when(workflows.get()).thenReturn(Map.of(WORKFLOW_WITH_RETRY_CONDITION.id(), WORKFLOW_WITH_RETRY_CONDITION));
+    terminationHandler.transitionInto(nonZeroTerm);
     verify(stateManager).receiveIgnoreClosed(Event.stop(WORKFLOW_INSTANCE), COUNTER);
   }
 
+  @Test
+  public void shouldEvaluateToTrue() {
```
NatashaL: What is this testing? all threads seem to be identical.

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
honnix: This keeps the same behaviour as any other validation. We always try to find as many problems as possible.

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
honnix: Since `retryConditionMet` returns a raw value, unboxing a `Nullable Boolean` may result NPE. Return value of `expression.getValue` is `Nullable`.

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/TerminationHandlerTest.java
honnix: This is to ensure thread-safety of the implementation because `expressionParser` is shared by threads.

honnix: It also serves as performance test. Maybe I should put a timeout.

NatashaL: Ok, let's reflect that in the test name then?

honnix: Yeah, absolutely.

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
NatashaL: I understand, however I see this piece of config to be of different nature. Some checks make sense to have a level of priority of others, such as this max length. I see the max length restriction to have the purpose of saving Styx cpu cycles and avoiding too long parsing times and users going crazy with their conditions. If the check fails, but then we still parse the expression - we've lost the point of the restriction.

honnix: I see this as what a normal compiler would always do: parsing as much as possible and finding all problems. And this is the same as what we have been doing to service account validation: first check length then check regex.  Parsing the expression is relatively fast (not much slower than regex validation). This piece of code is also used by styx cli so it is not needed to save CPU cycles there.

honnix: To me another point to think here is: when later on we need to validate another thing, should we stop earlier or not? Once opening an example, every time we need to validate different aspects we will be trapped into this kind of struggling.

NatashaL: Alright, makes sense

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/TerminationHandler.java
```diff
@@ -136,8 +136,9 @@ private boolean shouldFailFast(RunState state, Optional<Integer> exitCode) {
 
     var workflow = workflows.get().get(state.workflowInstance().workflowId());
     if (workflow == null) {
-      LOG.info("Workflow {} does not exist", state.workflowInstance().workflowId().toKey());
-      return true;
+      LOG.debug("Workflow {} does not exist possibly due to stale caching",
+          state.workflowInstance().workflowId().toKey());
+      return false;
```
honnix: @NatashaL I had second thought of this. Not entirely impossible that we have stale cache when reading the workflow, so . here we just let it retry normally and it's up to `ExecutionDescriptionHandler` to decide later.

PR #819, opened 2020-01-17
Retry decrement counter 

tnsetting:

## Description
For some reason, Styx lost track of shard counter. Shards all have counter 0 and when Styx tries to transition an active state, it can't because there is nothing to decrease. Therefore we will retry the transaction if it is not able to decrement the counter.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-service-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
+++ b/styx-service-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
@@ -41,6 +41,7 @@ import static java.util.stream.Collectors.toSet;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.google.api.client.util.ExponentialBackOff;
+import com.google.api.gax.retrying.RetrySettings;
 import com.google.cloud.Timestamp;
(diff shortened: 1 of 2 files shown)


styx-service-common/src/main/java/com/spotify/styx/storage/DatastoreStorage.java
```diff
@@ -955,20 +957,30 @@ private static StringValue jsonValue(Object o) throws JsonProcessingException {
         if (!isRetryableTransactionException(e.getCause())) {
           throw e;
         }
-        if (retrySettings.getMaxAttempts() > 0 && attempt >= retrySettings.getMaxAttempts()) {
+        checkAttemptsAndSleep(retrySettings, backoff, attempt, e);
+      } catch (CounterCapacityException e) {
```
honnix: This method is generic handling of running transaction with retries, dealing with this specific exception by checking the error message seems like code smell. At least this should be addressed in `isRetryableTransactionException`. Or be handling at the calling site.

honnix:
> Shards all have counter 0 and when Styx tries to transition an active state, it can't because there is nothing to decrease. Therefore we will retry the transaction if it is not able to decrement the counter.

How would retry work if everything is still 0? When you say `shards all have counter 0`, does that refer to the value in Datastore?

PR #820, opened 2020-02-07
upgrade crm lib revision

honnix:

## Description

Upgrade crm lib revision to support requestedPolicyVersion 3.

## Motivation and Context

Version 3 is required to set when requesting IAM policy with `condition` specified: https://cloud.google.com/iam/docs/policies#specifying-version

Otherwise this request simply fails and that results unexpected authorization failure.

I have updated unit tests.

Also tested in internal repo against real IAM.

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-service-common/src/test/java/com/spotify/styx/api/ServiceAccountUsageAuthorizerTest.java
+++ b/styx-service-common/src/test/java/com/spotify/styx/api/ServiceAccountUsageAuthorizerTest.java
@@ -50,6 +50,8 @@ import com.google.api.client.http.HttpResponseException;
 import com.google.api.services.admin.directory.Directory;
 import com.google.api.services.admin.directory.model.MembersHasMember;
 import com.google.api.services.cloudresourcemanager.CloudResourceManager;
+import com.google.api.services.cloudresourcemanager.model.GetIamPolicyRequest;
+import com.google.api.services.cloudresourcemanager.model.GetPolicyOptions;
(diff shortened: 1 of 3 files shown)


styx-service-common/src/test/java/com/spotify/styx/api/ServiceAccountUsageAuthorizerTest.java
```diff
@@ -159,7 +165,8 @@ public void setUp() throws IOException {
     when(authorizationPolicy.shouldEnforceAuthorization(any(), any(), any())).thenReturn(true);
     when(idToken.getPayload()).thenReturn(idTokenPayload);
     when(idTokenPayload.getEmail()).thenReturn(PRINCIPAL_EMAIL);
-    when((Object) crm.projects().getIamPolicy(any(), any()).execute()).thenReturn(projectPolicy);
+    when((Object) getIamPolicy.execute()).thenReturn(projectPolicy);
+    when((Object) crm.projects().getIamPolicy(any(), eq(GET_IAM_POLICY_REQUEST))).thenReturn(getIamPolicy);
```
honnix: Might be a limitation of mockito deep stub, but `when((Object) crm.projects().getIamPolicy(any(), eq(GET_IAM_POLICY_REQUEST)).execute()).thenReturn(projectPolicy);` doesn't work if later on we mock it again to get a different return value.

PR #830, opened 2020-08-05
Add FlyteExecConf and DockerExecConf to WorkflowConfiguration

narape:

## Description
This add some extra fields to `WorkflowConfiguration` to support Flyte execution configuration 
This PR docker and Flyte are not in the "same level" but the refactoring of Docker parameters will be dealt on a different PR.

```
Workflow: {
    componentId: String,
    workflowId: String,
    configuration: WorkflowConfiguration {
        workflowId: String,
        schedule: Schedule {
            expression: String
        }
        offset: Optional<String>,
        commitSha: Optional<String>,
        serviceAccount: Optional<String>,
        resources: List<String>,
        env: Map<String, String>,
        runningTimeout: Optional<Duration>,
        retryCondition: Optional<String>,
        dockerImage: Optional<String>,      <------- docker conf, we will move it later
        dockerArgs: Optional<List<String>>, <--+
        dockerTerminationLogging: boolean,  <--+
        flyteExecConf: Optional<FlyteExecConf> { <------ new flyte conf
            referenceId: FlyteIdentifier {
                resourceType: String,
                project: String,
                domain: String,
                name: String,    
                version: String,
            },
            inputFieldNames: Map<String, String>
        }
    }
}
```
## Motivation and Context
This is the start of the transition of making Styx able to execute launch plans on Flyte. 
To do so we need to modify our WorkflowConfiguration to specify this info.
Because this is so central to styx a lot of tests are needed to make sure that we are able to parse docker and flyte config style

I have included unit tests **and we will run a load test in staging**
## Checklist for PR author(s)

- [x] All tests pass


styx-common/src/test/java/com/spotify/styx/util/BasicWorkflowValidatorTest.java
```diff
@@ -267,7 +275,119 @@ public void shouldAcceptValidServiceAccount(String serviceAccount) {
...
+      identifierMutator.accept(referenceIdBuilder);
+      return confBuilder.flyteExecConf(
+          FlyteExecConfBuilder.from(flyteExecConf)
+              .referenceId(referenceIdBuilder.build())
+              .build())
+          .build();
+    }
+  }
+
+    private String limit(String msg, Object value, Object limit) {
```
sonjaer, line 390: I think this has wrong indentation

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -261,17 +289,24 @@ public void tearDown() throws Exception {
...
+  private static WorkflowConfiguration[] configurations() {
+    return new WorkflowConfiguration[] {
+        FULL_WORKFLOW_CONFIGURATION, DOCKER_EXEC_WORKFLOW_CONFIGURATION, FLYTE_EXEC_WORKFLOW_CONFIGURATION
+    };
+  }
+
   @Test
-  public void shouldDeleteWorkflows() throws Exception {
+  @Parameters(method = "configurations")
+  public void shouldDeleteWorkflows(WorkflowConfiguration configuration) throws Exception {
```
sonjaer, line 309: Not exactly sure what the `@Parameters` does, but the parameter to this function doesn't seem to be used inside it

narape: Done

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -222,10 +225,35 @@
...
   private static final WorkflowId WORKFLOW_ID = WorkflowId.create("dockerComp", "dockerEndpoint");
 
+  private static final String DOCKER_IMAGE = "gcr.io/foo/bar";
   private static final WorkflowConfiguration WORKFLOW_CONFIGURATION =
       WorkflowConfiguration.builder()
           .id(WORKFLOW_ID.id())
           .schedule(DAYS)
+          .dockerImage(DOCKER_IMAGE)
+          .build();
+  private static final WorkflowConfiguration DOCKER_EXEC_WORKFLOW_CONFIGURATION =
```
sonjaer, line 235: Is there a reason we don't use the TestData.java declaration of these variables?

sonjaer:
Added changes to:
* Ignore flyte workflows from the TriggeringManager
* Disabling users posting backfills for flyte workflows
* Disabling users doing adhoc triggering of flyte workflows
and tested this. Not sure if it is possible to test more end to end that a flyte workflow can't currently be scheduled, as my knowledge on Styx isn't 100%.

styx-common/src/main/java/com/spotify/styx/model/WorkflowConfiguration.java
```diff
@@ -42,10 +42,10 @@
 
   Optional<String> offset();
 
-  Optional<String> dockerImage();
-
   Optional<String> commitSha();
 
+  Optional<String> dockerImage();
```
ckiosidis, line 47: What is the strategy for deprecating these?

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
```diff
@@ -66,15 +71,13 @@ public BasicWorkflowValidator(DockerImageValidator dockerImageValidator) {
     final List<String> e = new ArrayList<>();
 
     var componentId = workflowId.componentId();
-    if (componentId.isEmpty()) {
-      e.add("component id cannot be empty");
-    } else if (componentId.contains("#")) {
+
+    empty(e, "component id", workflowId.componentId());
```
ckiosidis, line 75: Should we rename this to something more descriptive? like checkIfEmpty or something

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
```diff
@@ -144,11 +137,97 @@ public BasicWorkflowValidator(DockerImageValidator dockerImageValidator) {
...
+    }
+  }
+
+  private void empty(List<String> errors, String fieldName, String value) {
+    if (value.isEmpty()) {
+      errors.add(fieldName + " cannot be empty");
+    }
+  }
+
+  private <T> void same(List<String> errors, String fieldName, T valueA, T valueB) {
```
ckiosidis, line 228: should we rename this one as well?

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
```diff
@@ -144,11 +137,97 @@ public BasicWorkflowValidator(DockerImageValidator dockerImageValidator) {
...
+  }
+
+  @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+  private <T> void validateDockerParam(List<String> errors, String paramName, Optional<T> valueA, Optional<T> valueB,
+                                       BiConsumer<List<String>, T> validateFunction) {
+    valueA.ifPresent(a -> validateFunction.accept(errors, a));
+    valueB.ifPresent(b -> validateFunction.accept(errors, b));
+
+    valueA.ifPresent(a ->
+        valueB.ifPresent(b -> same(errors, paramName, a, b))
```
ckiosidis, line 183: Do we expect that both of these fields will be present and the same?

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
```diff
@@ -144,11 +137,97 @@ public BasicWorkflowValidator(DockerImageValidator dockerImageValidator) {
...
   }
 
   private static boolean validateServiceAccount(String serviceAccount) {
     var matcher = VALID_EMAIL_ADDRESS_REGEX.matcher(serviceAccount);
     return matcher.matches();
   }
+
+  private void validateDockerExecConf(List<String> errors, WorkflowConfiguration config) {
+    validateDockerParam(errors, "dockerImage", config.dockerImage(),
+        config.dockerExecConf().flatMap(DockerExecConf::dockerImage),
```
ckiosidis, line 154: Do we expect that both of these dockerImages should be present and the same?

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
```diff
@@ -144,11 +137,97 @@ public BasicWorkflowValidator(DockerImageValidator dockerImageValidator) {
...
+    //              .dockerImage("busybox")
+    //              .dockerArgs(List.of("x", "y"))
+    //              .dockerTerminationLogging(true)
+    //              .build())
+    //          .build();
+    // will fail as default at upper level is false but inside dockerExecConf is set to true
+  }
+
+  @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+  private <T> void validateDockerParam(List<String> errors, String paramName, Optional<T> valueA, Optional<T> valueB,
```
ckiosidis, line 177: I kind of feel that this abstraction creates unnecessary complication to the code

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
ckiosidis: or shouldNotBeEmpty

styx-common/src/main/java/com/spotify/styx/testdata/TestData.java
```diff
@@ -140,6 +143,37 @@
           .retryCondition("#exitCode == 1 && (#tries < 3 || #consecutiveFailures < 4) && #triggerType == \"natural\"")
           .build();
 
+  public static final WorkflowConfiguration DOCKER_EXEC_WORKFLOW_CONFIGURATION =
+      WorkflowConfiguration.builder()
+          .id("styx.TestEndpoint")
+          .commitSha(VALID_SHA)
+          .schedule(DAYS)
+          .serviceAccount("foo@bar.baz.quux")
+          .dockerExecConf(new DockerExecConfBuilder()
```
ckiosidis, line 152: Why do we have a 
WorkflowConfiguration.builder(), but then a `new DockerExecConfBuilder()` 

Should we be consistent with what we use?

styx-common/src/main/java/com/spotify/styx/testdata/TestData.java
```diff
@@ -140,6 +143,37 @@
...
+              .build())
+          .build();
+
+  public static final WorkflowConfiguration FLYTE_WORKFLOW_CONFIGURATION =
+      WorkflowConfiguration.builder()
+          .id("styx.TestEndpoint")
+          .commitSha(VALID_SHA)
+          .schedule(DAYS)
+          .serviceAccount("foo@bar.baz.quux")
+          .flyteExecConf(new FlyteExecConfBuilder()
```
ckiosidis, line 165: Let's make the Builder usage consistent with the WorkflowConfiguration.builder()

styx-common/src/main/java/com/spotify/styx/testdata/TestData.java
```diff
@@ -140,6 +143,37 @@
...
+              .referenceId(new FlyteIdentifierBuilder()
```
ckiosidis, line 166: Let's make the Builder usage consistent with the WorkflowConfiguration.builder()

styx-common/src/test/java/com/spotify/styx/model/WorkflowConfigurationTest.java
```diff
@@ -76,4 +86,116 @@ public void shouldReturnDefaultOffset() {
...
+  }
+
+  public static class ParseJsonArgsProvider {
+
+    private ParseJsonArgsProvider(){
+      // Prevents instantiation
+    }
+
+    @SuppressWarnings("unused")
+    public static Object[] provideDockerCentredConf() {
```
ckiosidis, line 107: do we mean dockerExecConf ?

styx-common/src/test/java/com/spotify/styx/model/WorkflowConfigurationTest.java
```diff
@@ -76,4 +86,116 @@ public void shouldReturnDefaultOffset() {
...
+             + "\"service_account\":\"foo@project.com\","
+             + "\"resources\":[\"resource\"],"
+             + "\"env\":{\"STYX_FOO\":\"bar\"},"
+             + "\"running_timeout\":10800.000000000,"
+             + "\"retry_condition\":\"true\","
+             + execConf
+             + "}";
+    }
+  }
 }
```
ckiosidis, line 201: missing new line

styx-common/src/test/java/com/spotify/styx/util/BasicWorkflowValidatorTest.java
```diff
@@ -267,6 +275,118 @@ public void shouldAcceptValidServiceAccount(String serviceAccount) {
...
+        sut.validateWorkflow(Workflow.create("test", DOCKER_EXEC_WORKFLOW_CONFIGURATION)),
+        empty()
+    );
+  }
+
+  @Test
+  public void shouldRejectInvalidImageInDockerExecConf() {
+    when(dockerImageValidator.validateImageReference(anyString())).thenReturn(Set.of("foo", "bar"));
+
+    var errors = sut.validateWorkflow(Workflow.create("test", DOCKER_EXEC_WORKFLOW_CONFIGURATION));
```
ckiosidis, line 290: Should we instead create an INVALID_DOCKER_EXEC_WORKFLOW_CONFIGURATION, and make this test a bit more explicit?

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -152,7 +152,8 @@ private void tryTriggering(Workflow workflow,
   private void tryTriggering0(Workflow workflow,
                                  TriggerInstantSpec instantSpec,
                                  Set<WorkflowId> enabledWorkflows) {
-    if (enabledWorkflows.contains(workflow.id())) {
+    // Ignore flyte workflows
+    if (enabledWorkflows.contains(workflow.id()) && workflow.configuration().flyteExecConf().isEmpty()) {
```
ckiosidis, line 156: I think it makes sense to remove the flyteExecConf from this if check, and instead do an early return with it's own debug log

styx-test/src/test/java/com/spotify/styx/storage/DatastoreEmulatorTest.java
```diff
@@ -66,11 +64,12 @@ public void testEmulator(Datastore client) {
...
     var emulator = new DatastoreEmulator();
-    exception.expect(AssertionError.class);
-    exception.expectMessage(Matchers.startsWith("Not using gcloud sdk datastore emulator"));
-    emulator.before();
+
+    var exception = Assert.assertThrows(AssertionError.class, emulator::before);
+
+    assertThat(exception.getMessage(), Matchers.startsWith("Not using gcloud sdk datastore emulator"));
   }
 }
```
ckiosidis, line 75: new line

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -65,7 +66,8 @@
  */
 class TriggerManager implements Closeable {
 
-  private static final Logger LOG = LoggerFactory.getLogger(TriggerManager.class);
+  private static final Logger DEFAULT_LOG = LoggerFactory.getLogger(TriggerManager.class);
+  private final Logger log;
```
ckiosidis, line 70: we don't need a different logger here. Let's reuse the first one.

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
```diff
@@ -152,19 +163,23 @@ private void tryTriggering(Workflow workflow,
...
                                  TriggerInstantSpec instantSpec,
                                  Set<WorkflowId> enabledWorkflows) {
-    // Ignore flyte workflows
-    if (enabledWorkflows.contains(workflow.id()) && workflow.configuration().flyteExecConf().isEmpty()) {
+
+    boolean isFlyteWorkflow = workflow.configuration().flyteExecConf().isPresent();
+
+    if (isFlyteWorkflow) {
+      log.info("Skip triggering flyte workflow");
+    } else if (enabledWorkflows.contains(workflow.id())) {
```
ckiosidis, line 171: It's better here to do
```
    if (isFlyteWorkflow) {
      log.info("Skip triggering flyte workflow");
      return;
    }

    if (enabledWorkflows.contains(workflow.id())) {...
```

this way we separate from the other (unrelated logic), and we return early and safe

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
sonjaer: I added this to be able to pass a mocked logger from the test class to verify that the logging happened

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
sonjaer: I did it this way since I want this code to execute: https://github.com/spotify/styx/pull/830/commits/5fbba776fb6f5edc968d52fecc7d8001ec8edf44#diff-8681f2f92e0c50ca9a99b3fc6ee3e4a2R196 So that the next trigger date updates

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
narape: I agreed that those are better names but I was being consistent with `limit`. We can make this change and limit in a following PR

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
narape: Yes it is confusing. I will change it to be a clear cut old style or clear cut new style in flyte or docker

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
narape: Usually verifying logging is not needed. An exception case would be that logging is the interface between systems, like our audit logging

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
narape: No really, I will move code overthere

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
sonjaer: Makes sense, I can revert that last commit

styx-scheduler-service/src/main/java/com/spotify/styx/TriggerManager.java
ckiosidis: Do we want this to happen? and run the `updateNextNaturalTrigger` regardless if it will fail next time?

sonjaer: I think so. I assume we want to handle it in the same way as disabled workflows. Otherwise, I think that once we enable the ability to run flyte workflows that each workflow would backfill from the date it was created until we enable the functionality. Not that this would maybe be a huge problem, but seems better to not cause them to backfill.

sonjaer: This is my understanding of what happens. But would be great with a second pair of eyes.

styx-common/src/main/java/com/spotify/styx/model/WorkflowConfiguration.java
narape: Step 1: Change model to allow both. This PR
Step 2: Remove old methods, keep tests that added in this PR to make sure that we still can read old json and read old entities in Datastore but converting old docker params to new DockerExecConf.

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
```diff
@@ -149,45 +147,33 @@ private static boolean validateServiceAccount(String serviceAccount) {
...
-    //              .dockerTerminationLogging(true)
-    //              .build())
-    //          .build();
-    // will fail as default at upper level is false but inside dockerExecConf is set to true
+  private void validateConflictingExecConf(List<String> e, WorkflowConfiguration cfg) {
+    var hasOldStyleDockerConf = cfg.dockerImage().isPresent() || cfg.dockerArgs().isPresent();
+    var hasNewStyleDockerConf = cfg.dockerExecConf().isPresent();
+
+    if (hasNewStyleDockerConf && hasOldStyleDockerConf) {
+      e.add("configuration cannot docker parameters in new and old style");
```
sonjaer, line 155: Nothing big, but missing `specify both`

styx-scheduler-service/src/test/java/com/spotify/styx/TriggerManagerTest.java
```diff
@@ -54,23 +54,21 @@
...
 public class TriggerManagerTest {
 
   private static final Trigger NATURAL_TRIGGER = Trigger.natural();
 
-  private static Workflow WORKFLOW_DAILY =
+  private static final Workflow WORKFLOW_DAILY =
       Workflow.create("comp", FULL_WORKFLOW_CONFIGURATION);
 
-  private static Workflow FLYTE_WORKFLOW =
+  private static final Workflow FLYTE_WORKFLOW =
```
sonjaer, line 66: This could be removed

styx-common/src/test/java/com/spotify/styx/util/BasicWorkflowValidatorTest.java
```diff
@@ -88,6 +103,8 @@
   public void setUp() throws Exception {
     MockitoAnnotations.initMocks(this);
     when(dockerImageValidator.validateImageReference(anyString())).thenReturn(Set.of());
+    when(dockerImageValidator.validateImageReference(eq(NOT_VALID_IMAGE)))
```
sonjaer, line 106: Nothing important, but do we need to mock this? Is that good practice? Or is there another way to do it?

styx-common/src/main/java/com/spotify/styx/model/WorkflowConfiguration.java
ckiosidis: 👍

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
ckiosidis: cool 👍

styx-common/src/test/java/com/spotify/styx/util/BasicWorkflowValidatorTest.java
narape: Well I guess it depends to whom you ask. 
* Some people are against mocking because they say you should create interface for everything and then implement simple test classes implementation of those interfaces. 
* Other people are against creating "unnecessary" interfaces because that is what mocking libraries are for.  

I see that both are possible approaches but I find Mocking libraries are quite handy and I don't find the downsides of mocking that convincing.

To the question of do we need to mock this? I say that we want verify that we keep calling the validator and react to it as we should, if I comment by mistake the if block that calls the validator then I want a test to fail. So we need a test for that and we want to control when to fail and when not to fail. We could achieve this by mocking (as done in the PR) or we could have introduced a new interface and make the validator implement that interface and create a test class that implements that interface but it is certainly more work that just adding two lines of mocking.

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
```diff
@@ -149,45 +147,33 @@ private static boolean validateServiceAccount(String serviceAccount) {
...
+    }
+    if (cfg.flyteExecConf().isPresent() && (hasNewStyleDockerConf || hasOldStyleDockerConf)) {
+      e.add("configuration cannot specify both docker and flyte parameters");
```
ckiosidis, line 158: shouldn't it be allowed to have flyteExecConf and dockerExecConf ?

narape: I guess you are asking because for migrating workflows we said that we will switch workflows from Docker to Flyte. But this PR is not that. My intention with this PR is to have either docker or flyte but not both. Future PRs we will look on how to do the switch and I didn't want to mix concerns

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
```diff
@@ -1131,8 +1105,13 @@ public void runInTransactionWithRetriesShouldPropagateDatastoreIOException() thr
...
     doThrow(cause).when(datastore).newTransaction();
-    exception.expect(is(cause));
-    storage.runInTransactionWithRetries(transactionFunction);
+
+    var exception = assertThrows(
+        DatastoreIOException.class,
+        () -> storage.runInTransactionWithRetries(transactionFunction)
+    );
+
+    assertThat(exception, is(cause));
```
ckiosidis, line 1114: these tests look a bit too rigid, but good refactoring

styx-scheduler-service/src/main/java/com/spotify/styx/StateInitializingTrigger.java
```diff
@@ -50,13 +51,18 @@
   @Override
   public void event(Workflow workflow, Trigger trigger, Instant instant,
                     TriggerParameters parameters) {
-    if (workflow.configuration().dockerImage().isEmpty()) {
+    final var configuration = workflow.configuration();
+    if (configuration.flyteExecConf().isPresent()) {
+      LOG.info("{} is a Flyte Workflow, skipping as not supported at the moment", workflow.id());
+      return;
+    } else if (configuration.dockerImage().isEmpty()
```
ckiosidis, line 58: The flyteExecConf and dockerImage look unrelated at first.
Can we add a comment here describing the relation and why we need them in a else if block, instead of two separate early returns?
It should help us moving on when working with the code.

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
narape: This was more a "Boy Scout rule" change as the way for asserting for exception using `@Rule` is now deprecated in favor of `assertThrows`. Functionally this test is as strict as the old version, just using a not deprecated API

styx-scheduler-service/src/test/java/com/spotify/styx/StateInitializingTriggerTest.java
```diff
@@ -123,19 +140,37 @@ public void shouldInjectTriggerExecutionEventWithBackfillTrigger() throws Except
   }
 
   @Test
-  public void shouldDoNothingIfDockerImageMissing() {
-    final WorkflowConfiguration configuration =
+  @Parameters(method = "noDockerImageConfigurations")
+  public void shouldDoNothingIfDockerImageMissing(WorkflowConfiguration configuration) {
```
ckiosidis, line 144: What is the benefit of having this method Parameterized instead of having a static generation of the configuration, or defining it in the beginning of the method?

styx-scheduler-service/src/test/java/com/spotify/styx/StateInitializingTriggerTest.java
```diff
@@ -150,12 +185,26 @@ public void shouldCreateWorkflowInstanceParameter() throws Exception {
...
     return WorkflowConfiguration.builder()
         .id("styx.TestEndpoint")
         .schedule(schedule)
         .dockerImage("busybox")
         .dockerArgs(ImmutableList.copyOf(args))
         .build();
   }
+
+  @Test
+  public void shouldWrapIsClosedExceptionIntoRuntime() throws Exception {
```
ckiosidis, line 198: what are we testing in this test? is it something we added in this commit?

styx-scheduler-service/src/main/java/com/spotify/styx/StateInitializingTrigger.java
narape: Aha, yes good point. My tests didn't pass until I did the else if but I didn't document my finding

styx-common/src/test/java/com/spotify/styx/util/BasicWorkflowValidatorTest.java
ckiosidis: Mocking here is ok if we have others tests where we test the behaviour of the `dockerImageValidator.validateImageReference`.
Since this test only makes sure that the method is called, and not that the validation logic behaves properly.
Here are the other tests: https://github.com/spotify/styx/blob/master/styx-common/src/test/java/com/spotify/styx/util/DockerImageValidatorTest.java

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
ckiosidis: right i hear you.
So what am I missing here?
We don't care about the migration/switch here.
So we should have 2 valid cases:
- either it's the old workflow configuration.
- or it's Flyte
It seems to me that the if-statement will trigger whenever Flyte is available.

styx-service-common/src/test/java/com/spotify/styx/storage/DatastoreStorageTest.java
ckiosidis: yes I agree. I mean that the test itself, not the refactoring

styx-scheduler-service/src/test/java/com/spotify/styx/StateInitializingTriggerTest.java
sonjaer: The test is running for multiple configurations

styx-common/src/test/java/com/spotify/styx/util/BasicWorkflowValidatorTest.java
sonjaer: Ok thank you 👍  Was mostly asking out of curiosity.

styx-scheduler-service/src/test/java/com/spotify/styx/StateInitializingTriggerTest.java
narape: No it was a hole is our test coverage

styx-scheduler-service/src/test/java/com/spotify/styx/StateInitializingTriggerTest.java
ckiosidis: Ah right. Didn't notice it was 2 configurations. cool!

styx-common/src/main/java/com/spotify/styx/util/BasicWorkflowValidator.java
ckiosidis: Hm right. So it should be either the FlyteExecConf or the docker configuration. 👍

ckiosidis: Can we add a comment explaining that we dont need to store docker configuration with flyte, because it's stored on flyte itself?

PR #842, opened 2020-09-03
Create flyte execution 

sonjaer:
# Hey, I just made a Pull Request!

## Checklist for PR author(s)

- [ ] Changes are covered by unit test
- [ ] All tests pass
- [ ] Code coverage check passes
- [ ] Error handling is tested
- [ ] Errors are handled at the appropriate layer
- [ ] Errors that cannot be handled where they occur are propagated
- [ ] (optional) Changes are covered by system test
- [ ] Relevant documentation updated
- [ ] This PR has NO breaking change to public API
- [ ] This PR has breaking change to public API and it is documented

## Checklist for PR reviewer(s)

- [ ] This PR has been incorporated in release note for the coming version
- [ ] Risky changes introduced by this PR have been all considered

PR #843, opened 2020-09-07
Transition from running

ckiosidis:

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -94,7 +94,7 @@ public class FlyteAdminClientRunner implements FlyteRunner {
         case ABORTED:
         case TIMED_OUT:
           final String flyteCode = execution.getClosure().getError().getCode();
-          final int styxCode = flyteCodeToStyx(flyteCode);
+          final int styxCode = flyteErrorCodeToStyx(flyteCode);
           stateManager.receive(Event.terminate(runState.workflowInstance(), Optional.of(styxCode)));
           break;
       }
@@ -103,8 +103,8 @@ public class FlyteAdminClientRunner implements FlyteRunner {
     }
   }
 
-  private int flyteCodeToStyx(final String flyteCode) {
-      switch (flyteCode) {
+  private int flyteErrorCodeToStyx(final String errorCode) {
+      switch (errorCode) {
         case "USER:NotReady":
           return 20;
         case "USER:NotRetryable":
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -25,6 +25,7 @@ import static com.spotify.styx.state.RunState.SUCCESS_EXIT_CODE;
 import static com.spotify.styx.state.RunState.UNKNOWN_ERROR_EXIT_CODE;
 import static com.spotify.styx.state.RunState.UNRECOVERABLE_FAILURE_EXIT_CODE;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.spotify.styx.flyte.client.FlyteAdminClient;
 import com.spotify.styx.model.Event;
 import com.spotify.styx.model.FlyteExecConf;
@@ -103,7 +104,8 @@ public class FlyteAdminClientRunner implements FlyteRunner {
 
   }
 
-  private void emitFlyteEvents(ExecutionOuterClass.Execution execution, RunState runState) {
+  @VisibleForTesting
+  void emitFlyteEvents(ExecutionOuterClass.Execution execution, RunState runState) {
     final Execution.WorkflowExecution.Phase phase = execution.getClosure().getPhase();
     final FlytePhase flytePhase = FlytePhase.fromProto(phase);
     try {
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -30,6 +30,7 @@ import com.spotify.styx.model.Event;
 import com.spotify.styx.model.FlyteExecConf;
 import com.spotify.styx.state.RunState;
 import com.spotify.styx.state.StateManager;
+import com.spotify.styx.state.handlers.DockerRunnerHandler;
 import com.spotify.styx.util.IsClosedException;
 import flyteidl.admin.ExecutionOuterClass;
 import flyteidl.core.Execution;
@@ -38,9 +39,12 @@ import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
 import java.util.Objects;
 import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class FlyteAdminClientRunner implements FlyteRunner {
 
+  private static final Logger LOG = LoggerFactory.getLogger(FlyteAdminClientRunner.class);
   private final FlyteAdminClient flyteAdminClient;
   private final StateManager stateManager;
 
@@ -115,6 +119,7 @@ public class FlyteAdminClientRunner implements FlyteRunner {
           break;
       }
     } catch (IsClosedException e) {
+      LOG.warn("Could not emit 'terminate' event", e);
       return;
     }
   }
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -20,6 +20,10 @@
 
 package com.spotify.styx.flyte;
 
+import static com.spotify.styx.state.RunState.MISSING_DEPS_EXIT_CODE;
+import static com.spotify.styx.state.RunState.UNKNOWN_ERROR_EXIT_CODE;
+import static com.spotify.styx.state.RunState.UNRECOVERABLE_FAILURE_EXIT_CODE;
+
 import com.spotify.styx.flyte.client.FlyteAdminClient;
 import com.spotify.styx.model.Event;
 import com.spotify.styx.model.FlyteExecConf;
@@ -106,11 +110,11 @@ public class FlyteAdminClientRunner implements FlyteRunner {
   private int flyteErrorCodeToStyx(final String errorCode) {
       switch (errorCode) {
         case "USER:NotReady":
-          return 20;
+          return MISSING_DEPS_EXIT_CODE;
         case "USER:NotRetryable":
-          return 50;
+          return UNRECOVERABLE_FAILURE_EXIT_CODE;
         default:
-          return 1;
+          return UNKNOWN_ERROR_EXIT_CODE;
     }
   }
 
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -54,6 +54,7 @@ public class FlyteAdminClientRunner implements FlyteRunner {
     this.stateManager = stateManager;
   }
 
+  @Override
   public FlyteExecutionId createExecution(final String name, final FlyteExecConf flyteExecConf)
       throws CreateExecutionException {
     final var flyteIdentifier = flyteExecConf.referenceId();
@@ -82,6 +83,7 @@ public class FlyteAdminClientRunner implements FlyteRunner {
     }
   }
 
+  @Override
   public void poll(FlyteExecutionId flyteExecutionId, RunState runState)
       throws PollingException {
     Objects.requireNonNull(flyteExecutionId);
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -30,7 +30,6 @@ import com.spotify.styx.model.Event;
 import com.spotify.styx.model.FlyteExecConf;
 import com.spotify.styx.state.RunState;
 import com.spotify.styx.state.StateManager;
-import com.spotify.styx.state.handlers.DockerRunnerHandler;
 import com.spotify.styx.util.IsClosedException;
 import flyteidl.admin.ExecutionOuterClass;
 import flyteidl.core.Execution;
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -105,25 +105,21 @@ public class FlyteAdminClientRunner implements FlyteRunner {
   }
 
   @VisibleForTesting
-  void emitFlyteEvents(ExecutionOuterClass.Execution execution, RunState runState) {
+  void emitFlyteEvents(ExecutionOuterClass.Execution execution, RunState runState)
+      throws IsClosedException {
     final Execution.WorkflowExecution.Phase phase = execution.getClosure().getPhase();
     final FlytePhase flytePhase = FlytePhase.fromProto(phase);
-    try {
-      switch (flytePhase) {
-        case SUCCEEDED:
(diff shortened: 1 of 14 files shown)


ckiosidis:
depends on https://github.com/spotify/styx/pull/842

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -62,4 +72,46 @@ public FlyteExecution createExecution(final String name, final FlyteExecConf fly
...
+          stateManager.receive(Event.terminate(runState.workflowInstance(), Optional.of(styxCode)));
+          break;
+      }
+    } catch (IsClosedException e) {
+      return;
+    }
+  }
+
+  private int flyteCodeToStyx(final String flyteCode) {
+      switch (flyteCode) {
```
narape, line 107: Can you put error in the name so it is more clear that we don't need to account for something to return 0?
```suggestion
  private int flyteErrorCodeToStyx(final String errorCode) {
      switch (errorCode) {
```

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -62,4 +72,46 @@ public FlyteExecution createExecution(final String name, final FlyteExecConf fly
...
+    emitFlyteEvents(execution, runState);
+  }
+
+  private void emitFlyteEvents(ExecutionOuterClass.Execution execution, RunState runState) {
+    final Execution.WorkflowExecution.Phase phase = execution.getClosure().getPhase();
+    final FlytePhase flytePhase = FlytePhase.fromProto(phase);
+    try {
+      switch (flytePhase) {
+        case SUCCEEDED:
+          stateManager.receive(Event.terminate(runState.workflowInstance(), Optional.of(0)));
```
narape, line 91: Can you put that 0 and the 20, 50 and 1 as constant with descriptive names? maybe we have those constants defined already somewhere

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -62,4 +72,46 @@ public FlyteExecution createExecution(final String name, final FlyteExecConf fly
...
+        case FAILED:
+        case ABORTED:
+        case TIMED_OUT:
+          final String flyteCode = execution.getClosure().getError().getCode();
+          final int styxCode = flyteCodeToStyx(flyteCode);
+          stateManager.receive(Event.terminate(runState.workflowInstance(), Optional.of(styxCode)));
+          break;
+      }
+    } catch (IsClosedException e) {
+      return;
```
narape, line 102: Please log a warning here as it is done in the Handlers

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -62,4 +72,46 @@ public FlyteExecution createExecution(final String name, final FlyteExecConf fly
       throw new CreateExecutionException(flyteExecConf, e);
     }
   }
+
+  public void poll(final String project, final String domain, final String name, RunState runState) {
```
narape, line 76: I think that we should create or own version of [WorkflowExecutionIdentifier](https://github.com/lyft/flyteidl/blob/fc787dc41985215ab025fbd55de14764fd7efa03/protos/flyteidl/core/identifier.proto#L39) and use it here instead of passing the three values independently

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -62,4 +72,46 @@ public FlyteExecution createExecution(final String name, final FlyteExecConf fly
...
+    Objects.requireNonNull(project);
+    Objects.requireNonNull(domain);
+    Objects.requireNonNull(name);
+    final ExecutionOuterClass.Execution execution =
+        flyteAdminClient.getExecution(project, domain, name);
```
narape, line 81: Probably we need to catch exception here, unless it would be handled in the upper layers

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlytePhase.java
```diff
@@ -0,0 +1,58 @@
...
+        .put(Execution.WorkflowExecution.Phase.ABORTED, FlytePhase.ABORTED)
+        .put(Execution.WorkflowExecution.Phase.TIMED_OUT, FlytePhase.TIMED_OUT)
+      .build();
+
+  public static FlytePhase fromProto(Execution.WorkflowExecution.Phase phase) {
+    final var flytePhase = phaseMapper.get(phase);
+    if (flytePhase != null) {
+      return flytePhase;
+    }
+    throw new NoSuchElementException();
```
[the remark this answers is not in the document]
ckiosidis: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: I created the FlyteExecution for that, i can reuse it

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: The equivalent check is ignored in the Kuberenetes handler here https://github.com/spotify/styx/blob/master/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java#L663

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: https://github.com/spotify/styx/blob/master/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/DockerRunnerHandler.java#L94
RuntimeExceptions are not handled in the equivalent layer of the current impl, so i thought it was probably fine to do the same here

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: There was something in RunState which I extended to have the rest of the exit codes as well

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: Should I add it anyway?

narape: DockerRunnerHandler and FlyteRunnerHandlers both logs on catching `IsClosedException`. I see that this have been done inconsistently but I sill prefer to err on the side of more login

styx-scheduler-service/src/test/java/com/spotify/styx/test/FlyteAdminClientRunnerTest.java
```diff
@@ -117,4 +135,92 @@ public void testThrowsCreateExecutionExceptionForUnknownException() {
...
+    assertThrows(
+        NullPointerException.class,
+        () ->    flyteRunner.poll(FlyteExecutionId.create("flyte-test", "testing", "test-run-state-cannot-be-null"), null)
+    );
+  }
+
+  @Test
+  public void testTransititionRunningToTerminateSuccessfulRun() throws Exception {
+    Workflow workflow = Workflow.create("id", configuration());
+    WorkflowInstance workflowInstance = WorkflowInstance.create(workflow.id(), "2016-03-14");
```
narape, line 158: These could be in constants as they are reused in several tests

styx-scheduler-service/src/test/java/com/spotify/styx/test/FlyteAdminClientRunnerTest.java
```diff
@@ -20,30 +20,44 @@
...
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
-import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.MockitoAnnotations;
 
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(JUnitParamsRunner.class)
 public class FlyteAdminClientRunnerTest {
```
narape, line 61: We are missing a test of how the handler would behave if the `FlyteAdminClient` throws an exception

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -20,22 +20,36 @@
 
 package com.spotify.styx.flyte;
 
+import static com.spotify.styx.state.RunState.MISSING_DEPS_EXIT_CODE;
+import static com.spotify.styx.state.RunState.UNKNOWN_ERROR_EXIT_CODE;
```
narape, line 24: So that you can use it bellow
```suggestion
import static com.spotify.styx.state.RunState.SUCCESS_EXIT_CODE;
import static com.spotify.styx.state.RunState.UNKNOWN_ERROR_EXIT_CODE;
```

styx-scheduler-service/src/test/java/com/spotify/styx/test/FlyteAdminClientRunnerTest.java
```diff
@@ -117,4 +135,92 @@ public void testThrowsCreateExecutionExceptionForUnknownException() {
...
+
+    final FlyteExecutionId flyteExecutionId =
+        FlyteExecutionId.create("flyte-test", "testing", "execution-name");
+    flyteRunner.poll(flyteExecutionId, runState);
+    verify(stateManager,  timeout(60_000)).receive(Event.terminate(workflowInstance, Optional.of(0)));
+  }
+
+  @Test
+  @Parameters({
+      "FAILED, USER:NotReady, 20",
```
narape, line 176: You can also use the constant of `RunState` here.

styx-scheduler-service/src/test/java/com/spotify/styx/test/FlyteAdminClientRunnerTest.java
```diff
@@ -117,4 +135,92 @@ public void testThrowsCreateExecutionExceptionForUnknownException() {
...
+                .project("flyte-test")
+                .domain("testing")
+                .name("test-lp")
+                .version("1")
+                .build())
+            .build()
+        )
+        .build();
+  }
+}
```
narape, line 226: New line.

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlytePhase.java
```diff
@@ -0,0 +1,57 @@
...
+  RUNNING,
+  SUCCEEDING,
+  SUCCEEDED,
+  FAILING,
+  FAILED,
+  ABORTED,
+  TIMED_OUT;
+
+  private final static Map<Execution.WorkflowExecution.Phase, FlytePhase> phaseMapper =
+    ImmutableMap.<Execution.WorkflowExecution.Phase, FlytePhase>builder()
```
narape, line 39: Just as a side note. Since Java 9 `Map.of()` achieve the same result as `ImmutableMap`. No need to change it as we depend on Guava anyways

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: right 👍 missed that

styx-scheduler-service/src/test/java/com/spotify/styx/test/FlyteAdminClientRunnerTest.java
ckiosidis: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlytePhase.java
ckiosidis: 👍

styx-scheduler-service/src/test/java/com/spotify/styx/test/FlyteAdminClientRunnerTest.java
ckiosidis: I think it's ok since the moment we will need to change the configuration or parametrise something, we will have to revert to what we see here.
How strongly do you feel about the refactoring?

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: I was looking at this and I think that the reason that `DockerRunnerHandler.java` is not handling exception is because [KubernetesDockerRunner](https://github.com/spotify/styx/blob/9ecb3f967ff86a0fe36c29d82b7d4026b679c6cd/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java#L107) is doing so. So it is not because it is handled un the upper layer but because it is handled in the lower layer. We should do the same as Styx shouldn't crash because a flyteadmin connection issue

styx-scheduler-service/src/test/java/com/spotify/styx/test/FlyteAdminClientRunnerTest.java
narape: Mark as solved as we are already discussing the same issue in other comment

ckiosidis: will fix 👍

styx-scheduler-service/src/test/java/com/spotify/styx/test/FlyteAdminClientRunnerTest.java
narape: Not super, but it looked that you needed to parameterized the response to the response of `getExecution()` the `workflow` and `workflowInstance` looks the same on both tests

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: yes i thought it's better to let the poll method handle the exception. Better also to keep everything in one place.
It seems that the KubDockerRunner is ignoring that one. But i think it's better to handle it explicitely.

styx-scheduler-service/src/test/java/com/spotify/styx/flyte/FlyteAdminClientRunnerTest.java
```diff
@@ -0,0 +1,295 @@
...
+    assertThrows(
+        NullPointerException.class,
+        () ->    flyteRunner.poll(FlyteExecutionId.create("flyte-test", "testing", "test-run-state-cannot-be-null"), null)
+    );
+  }
+
+  @Test
+  public void testTransititionRunningToTerminateSuccessfulRun() throws Exception {
+    Workflow workflow = Workflow.create("id", configuration());
+    WorkflowInstance workflowInstance = WorkflowInstance.create(workflow.id(), "2016-03-14");
```
narape, line 161: These two lines appears verbatim 6 times in these file. I think that we should defined as constant

narape (approved):
I still pointed out about the use of the possible use of the constant in the test now that with more test that were added we are repeating the lines 6 times.

sonjaer:
Should we use this in the FlyteRunnerHandler?

styx-scheduler-service/src/test/java/com/spotify/styx/flyte/FlyteAdminClientRunnerTest.java
ckiosidis: i moved them in a function instead 👍

PR #844, opened 2020-09-08
use polling to transitition to terminate

ckiosidis:

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
+++ b/styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
@@ -165,7 +165,26 @@ public class FlyteRunnerHandlerTest {
     flyteRunnerHandler.transitionInto(runState, eventRouter);
 
     verify(flyteRunner).poll(getFlyteExecutionId(FLYTE_EXECUTION_DESCRIPTION,
-        new StringBuilder(EXECUTION_ID).reverse().toString()), runState);
+        reverse.apply(EXECUTION_ID)), runState);
+  }
+
+  @Test
+  public void shouldReportRunErrorWhenCatchingExceptionDuringPolling()
+      throws FlyteRunner.PollingException, IsClosedException {
+    RunState runState = RunState.create(WORKFLOW_INSTANCE, State.RUNNING, StateData.newBuilder()
+        .executionId(EXECUTION_ID)
+        .executionDescription(FLYTE_EXECUTION_DESCRIPTION)
+        .build());
+    doThrow(new FlyteRunner.PollingException("Test polling exception"))
+        .when(flyteRunner)
+        .poll(any(), any());
+
+    flyteRunnerHandler.transitionInto(runState, eventRouter);
+
+    final var errMessage = "Test polling exception";
+
+    verify(eventRouter,  timeout(60_000)).receive(Event.runError(WORKFLOW_INSTANCE, errMessage),
+        runState.counter());
   }
 
   private FlyteExecutionId getFlyteExecutionId(ExecutionDescription executionDescription, String executionId) {
--- a/styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
+++ b/styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
@@ -164,7 +164,8 @@ public class FlyteRunnerHandlerTest {
 
     flyteRunnerHandler.transitionInto(runState, eventRouter);
 
-    verify(flyteRunner).poll(getFlyteExecutionId(FLYTE_EXECUTION_DESCRIPTION, EXECUTION_ID), runState);
+    verify(flyteRunner).poll(getFlyteExecutionId(FLYTE_EXECUTION_DESCRIPTION,
+        new StringBuilder(EXECUTION_ID).reverse().toString()), runState);
   }
 
   private FlyteExecutionId getFlyteExecutionId(ExecutionDescription executionDescription, String executionId) {
--- a/styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
+++ b/styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
@@ -20,12 +20,13 @@
 
 package com.spotify.styx.state.handlers;
 
-import static com.spotify.styx.state.handlers.FlyteRunnerHandler.STATIC_EXIT_CODE;
 import static com.spotify.styx.state.handlers.FlyteRunnerHandler.STATIC_RUNNER_ID;
 import static com.spotify.styx.testdata.TestData.EXECUTION_ID;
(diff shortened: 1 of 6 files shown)


styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
```diff
@@ -165,7 +155,24 @@ public void shouldNotThrowExceptionIfEvenRouterIsClosed(State state) throws Exce
...
+  @Test
+  public void shouldPollInRunning() throws FlyteRunner.PollingException {
+    RunState runState = RunState.create(WORKFLOW_INSTANCE, State.RUNNING, StateData.newBuilder()
+        .executionId(EXECUTION_ID)
+        .executionDescription(FLYTE_EXECUTION_DESCRIPTION)
+        .build());
+
+    flyteRunnerHandler.transitionInto(runState, eventRouter);
+
+    verify(flyteRunner).poll(getFlyteExecutionId(FLYTE_EXECUTION_DESCRIPTION, EXECUTION_ID), runState);
```
narape, line 169: The test is failing because `FlyteExecutionId` doesn't override `equals` and `hashCode` also it would be nice to override `toString`. Actually what it makes more sense is to convert `FlyteExecutionId` to AutoValue or Automatter

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
```diff
@@ -165,7 +155,24 @@ public void shouldNotThrowExceptionIfEvenRouterIsClosed(State state) throws Exce
         .build());
 
     flyteRunnerHandler.transitionInto(runState, eventRouter);
+  }
+
+  @Test
+  public void shouldPollInRunning() throws FlyteRunner.PollingException {
```
narape, line 161: Please also add another test for `shouldReportRunErrorWhenCatchingExceptionDuringPolling`

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
ckiosidis: yes i will fix

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
```diff
@@ -164,7 +164,8 @@ public void shouldPollInRunning() throws FlyteRunner.PollingException {
 
     flyteRunnerHandler.transitionInto(runState, eventRouter);
 
-    verify(flyteRunner).poll(getFlyteExecutionId(FLYTE_EXECUTION_DESCRIPTION, EXECUTION_ID), runState);
+    verify(flyteRunner).poll(getFlyteExecutionId(FLYTE_EXECUTION_DESCRIPTION,
+        new StringBuilder(EXECUTION_ID).reverse().toString()), runState);
```
narape, line 168: You can use the reverse function that it is already defined in line 63
```suggestion
        reverse.apply(EXECUTION_ID)), runState);
```

ckiosidis: yes i did forgot to push this change. It's already applied on my local branch

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
```diff
@@ -165,7 +154,43 @@ public void shouldNotThrowExceptionIfEvenRouterIsClosed(State state) throws Exce
...
+  public void shouldPollInRunning() throws FlyteRunner.PollingException {
+    RunState runState = RunState.create(WORKFLOW_INSTANCE, State.RUNNING, StateData.newBuilder()
+        .executionId(EXECUTION_ID)
+        .executionDescription(FLYTE_EXECUTION_DESCRIPTION)
+        .build());
+
+    flyteRunnerHandler.transitionInto(runState, eventRouter);
+
+    verify(flyteRunner).poll(getFlyteExecutionId(FLYTE_EXECUTION_DESCRIPTION,
+        reverse.apply(EXECUTION_ID)), runState);
```
tnsetting, line 169: Why do we need to use reversed `EXECUTION_ID`?

ckiosidis: The styx execution id `styx-run-xxxxxx` can not be used as is in Flyte, since Flyte has a limitation of (i think) 20 characters for the specific id.
So we generate it first and then transform it before passing it along in Flyte.

narape: Glad that you asked, it shows that it is not clear what are we doing here.
The issue is that we cannot use straight styx-run-ids as flyte-exec-names because flyte name are shorter. So a transformation function is needed. With the purpose of making the testing easier we pass a `Function<String, String>` as parameter to be able to use a simpler one (string reverse) for testing instead that the real one that it is more complicated.

To improve readability, we can create a constant:
```java
  private static final String FLYTE_EXEC_NAME = reverse.apply(EXECUTION_ID);
```
and then use the constant directly in the verify.

Side note: at the beginning we didn't use reverse but `Function.identity()` but this hid a bug that we were passing the styx-run-id instead fo the flyte exec name. The idea of using reverse was to do a simple transformation that it is easy to check.

narape: Does it makes sense @tnsetting?. By the way it took me like a second to recognize that this was your github account nian

tnsetting: @narape Yes. It is has been a while. 

That makes sense to me (didn't not realise the execution mapping part) and it sounds to use a constant to improve readability. 👍

tnsetting (commented):
It looks good for me overall.

styx-scheduler-service/src/test/java/com/spotify/styx/state/handlers/FlyteRunnerHandlerTest.java
```diff
@@ -165,7 +153,43 @@ public void shouldNotThrowExceptionIfEvenRouterIsClosed(State state) throws Exce
...
+        .executionId(EXECUTION_ID)
+        .executionDescription(FLYTE_EXECUTION_DESCRIPTION)
+        .build());
+    doThrow(new FlyteRunner.PollingException("Test polling exception"))
+        .when(flyteRunner)
+        .poll(any(), any());
+
+    flyteRunnerHandler.transitionInto(runState, eventRouter);
+
+    final var errMessage = "Test polling exception";
```
tnsetting, line 184: Should we create a constant for the `errMessage`?

narape (approved):

ckiosidis:
let's add the readability refactorings on another PR.

PR #847, opened 2020-09-10
fix error statemanager is not bound to environment

ckiosidis:

## Description

The statemanager is not bound to bound to the environment, so trying to resolve it in that way would not work.
Instead we pass along the actual instance since it's right there.

```
1) Explicit bindings are required and com.spotify.styx.state.StateManager is not explicitly bound.
  while locating com.spotify.styx.state.StateManager
```

## Checklist for PR author(s)

- [ ] All tests pass


narape (approved):
:+1:

PR #853, opened 2020-09-15
Terminate flyte executions on ERROR State

narape:

## Description
Terminate flyte executions when a styx workflows reaches STOP state.

## Motivation and Context
We want styx workflows instance executions reflects flyte executions and the other way around. 
At the moment Styx polling will make sure that styx will mark the styx wf instance execution as failed if the flyte execution fails. However there is no way to propagate halt or fail errors to flyte executions.
This is important because we shouldn't leak executions if the styx workflow instance execution has stopped. For example If I abort a styx execution I will expect my flyte execution to be aborted too.

I have included unit tests

## Checklist for PR author(s)

- [x] All tests pass


styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -44,6 +44,8 @@
 public class FlyteAdminClientRunner implements FlyteRunner {
 
   private static final Logger LOG = LoggerFactory.getLogger(FlyteAdminClientRunner.class);
+  @VisibleForTesting static final String TERMINATE_CAUSE = "Styx halt";
```
ckiosidis, line 47: should we make this a bit more descriptive? I feel that it's not going to be clear that there was triggered a terminate event

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -92,6 +94,27 @@ public String createExecution(RunState runState, final String execName, final Fl
...
+          flyteExecutionId.name(),
+          TERMINATE_CAUSE);
+    } catch (StatusRuntimeException e) {
+      if (e.getStatus().getCode() == Status.Code.NOT_FOUND) {
+        LOG.warn("Trying to terminate non existent flyte execution: {}", flyteExecutionId);
+        return;
+      } // TODO: Consider other status codes that could be interesting
+      LOG.warn("Couldn't terminate flyte execution: {}", flyteExecutionId, e);
+    } catch (Exception e) {
+      LOG.warn("Couldn't terminate flyte execution: {}", flyteExecutionId, e);
```
ckiosidis, line 114: is it enough to simply log a warn here and move on?

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/FlyteRunnerHandler.java
```diff
@@ -110,7 +112,7 @@ private void pollingExecution(RunState state, EventRouter eventRouter) {
     final String execName = styxExecIdToFlyteNameMapper.apply(executionId);
 
     try {
-      flyteRunner.poll(FlyteExecutionId.create(flyteExecConf.referenceId().project(), flyteExecConf.referenceId().domain(), execName), state);
+      flyteRunner.poll(getExecutionId(state), state);
```
ckiosidis, line 115: is the flyteExecCOnf used anywhere after this refactor? Can it be removed? What about the executionId and execName.
It looks to me that they are not

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -92,6 +94,27 @@ public String createExecution(RunState runState, final String execName, final Fl
     }
   }
 
+  @Override
+  public void terminateExecution(RunState runState, FlyteExecutionId flyteExecutionId) {
+    requireNonNull(runState, "runState");
+    requireNonNull(flyteExecutionId, "flyteExecutionId");
+    try {
+      flyteAdminClient.terminateExecution(
```
ckiosidis, line 102: what happens if the workflow is already terminated?

narape: It does nothing, at least in the cli. I can create a small program to validate what happens with the API and either I will make changes or add a comment describing that nothing happens if the workflow execution have been terminated

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/AbstractRunnerHandler.java
```diff
@@ -64,19 +67,29 @@ public final void transitionInto(RunState state, EventRouter eventRouter) {
         if(!applicableFn.test(maybeExecutionDescription.orElseThrow())) {
           return;
         }
+        safeTransitionInto(state, eventRouter);
+        break;
 
+      case ERROR:
```
ckiosidis, line 73: Thinking out loud here an ERROR would mean that there is only an ERROR in styx, right?
Flyte errors would cause a terminated state instead of an error state.
What if we see multiple error events coming in?

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: sounds good. i think it's good to know what happens just to make sure that we don't mistreat something

narape: I confirmed that using the the grpc client we gets the same result: it is ok to terminate a workflow more than once without throwing an execption. I will add a comment to make it clear

styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/FlyteRunnerHandler.java
narape: Yes, good catch. That was the reason of introducing the method to avoid repeating those lines

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: I will change the last one to log error but we will give an holistic review between styx and flyte failures in a different PR

ckiosidis (approved):
LGTM

PR #854, opened 2020-09-17
Treat ALREADY_EXISTS exception in createExecution as OK

narape:

## Description
Treat ALREADY_EXISTS exception in `createExecution` as OK
 
## Motivation and Context
In the rare cases that we are repeating the event that leads to create the same execution twice we will get ALREADY_EXISTS. This PR catch this scenario and return as is we created the execution successfully.

## styx-run-id -> flyte exec name problem
There is a problem in translating styx-run-id to flyte exec name: flyte exec names are shorter. So we did a mapping that has a very low probability of generating a collisions 
([see](https://github.com/spotify/styx/blob/288995d6aad3040886fa271af78700f693df0578/styx-scheduler-service/src/main/java/com/spotify/styx/state/handlers/StyxIdToFlyteExecNameMapper.java#L37)). A flyte exec name collision means that two styx-run-id bellowing to a different workflow+parameters produce the same flyte exec name. 
So we can tray to create an execution for workflow b and we get a ALREADY_EXIST because the collision (with a low probability). **We won't solve this issue** in this PR

I have included unit tests.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -67,7 +67,8 @@ public class FlyteAdminClientRunner implements FlyteRunner {
     final var launchPlanIdentifier = flyteExecConf.referenceId();
     final var execMode = runState.data().trigger()
         .map(this::toFlyteExecutionMode)
-        .orElse(ExecutionMode.UNRECOGNIZED);
+        .filter(mode -> mode != ExecutionMode.UNRECOGNIZED)
+        .orElseThrow(() -> new CreateExecutionException("Missing trigger or unknown in StateData: " + runState.data()));
 
     // TODO: verify if the execution already exist
     try {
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -86,9 +85,15 @@ public String createExecution(RunState runState, final String execName, final Fl
...
       return runnerId;
     } catch (StatusRuntimeException e) {
-      if (e.getStatus().getCode() == Status.Code.NOT_FOUND) {
-        throw new LaunchPlanNotFound(flyteExecConf, e);
-      } // TODO: Consider other status codes that could be interesting
+      switch (e.getStatus().getCode()) {
+        case ALREADY_EXISTS:
+          // TODO: 🤔 How do we make sure that we tell apart between low probabilities of collisions
+          //  on hashing of styx-run-id --> flyte exec and repetitions of same styx-id execution.
+          //  For now we always assume the later.
```
ckiosidis, line 92: How is this expected to be solved? I believe that we probably just let styx retry and the likelihood of the problem happening again is negligible.
Also what do you mean of `same styx-id execution` ?

ckiosidis (approved):
let's address the error scenarios in another pr

PR #857, opened 2020-09-20
Try reading token from gRPC context first

honnix:

## Description

In Styx client, if gRPC context carries an authorization token, use that one instead of generating one from credentials.

## Motivation and Context

Styx client user can propagate an already issued authorization token instead of letting the client generate one.

Unit test.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-client/src/main/java/com/spotify/styx/client/Constant.java
+++ b/styx-client/src/main/java/com/spotify/styx/client/Constant.java
@@ -1,0 +1,32 @@
+/*-
+ * -\-\-
+ * Spotify Styx API Client
+ * --
+ * Copyright (C) 2016 - 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.styx.client;
+
+import io.grpc.Context;
+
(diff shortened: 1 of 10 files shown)


styx-client/src/main/java/com/spotify/styx/client/Constant.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.styx.client;
+
+import io.grpc.Context;
+
+public final class Constant {
```
narape, line 25: This name is not very descriptive. What can of constant it contains? is the question that came up to my mind the moment that I read the name of the class. Here are some suggestions: `GrpcAuthConstant` would be more specific or `GrpcAuthUtil` we we are going to expand this further

styx-client/src/main/java/com/spotify/styx/client/StyxOkHttpClient.java
```diff
@@ -395,13 +410,16 @@ static StyxClient create(String apiHost, FutureOkHttpClient client, GoogleIdToke
...
   private CompletionStage<Response> execute(Request request) {
-    final Optional<String> authToken;
-    try {
-      authToken = auth.getToken(apiHost.toString());
-    } catch (IOException | GeneralSecurityException e) {
-      // Credential probably invalid, configured wrongly or the token request failed.
-      throw new ClientErrorException("Authentication failure: " + e.getMessage(), e);
-    }
+    var authToken = Optional
+        .ofNullable(AUTHORIZATION_KEY.get())
```
narape, line 414: Wow this is kind of magic. I had go and read about GRPC to understand why this makes sense

styx-client/src/main/java/com/spotify/styx/client/Constant.java
honnix: This is not meant to be grpc only constant. It's just for now we don't have anything else. If this feels unclear, could it be `GrpcContextKey`?

narape: That sounds good

narape (approved):

PR #858, opened 2020-09-21
Passing the parameter to create execution

tnsetting:

## Description
Passing the parameter (Aka. partition) to when creating a Flyte WF execution. It will fetch the launch plan from Flyte first to get all the default input. When it find a input parameter with name `parameter`, styx will replace this parameter with the real value from workflow instance.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/ParameterUtil.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/ParameterUtil.java
@@ -109,6 +109,18 @@ public final class ParameterUtil {
         ISO_LOCAL_DATE_HOUR.parse(dateHour)).atOffset(UTC));
   }
 
+  public static Instant parseDateMonth(String dateMonth) {
+    return Instant.from(LocalDate.from(
+        DateTimeFormatter.ISO_LOCAL_DATE.parse(dateMonth + "-01"))
+        .atStartOfDay(UTC));
+  }
+
+  public static Instant parseDateYear(String dateYear) {
+    return Instant.from(LocalDate.from(
+        DateTimeFormatter.ISO_LOCAL_DATE.parse(dateYear + "-01-01"))
+        .atStartOfDay(UTC));
+  }
+
   public static String toParameter(Schedule schedule, Instant instant) {
     switch (schedule.wellKnown()) {
       case DAILY:
--- a/styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
+++ b/styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
@@ -22,6 +22,7 @@ package com.spotify.styx.flyte.client;
 
 import static com.google.common.base.Strings.nullToEmpty;
 import static com.google.common.base.Verify.verifyNotNull;
+import static com.spotify.styx.flyte.client.FlyteInputsUtils.fillingParameterInInputs;
 
 import flyteidl.admin.Common;
 import flyteidl.admin.ExecutionOuterClass;
@@ -36,7 +37,7 @@ import org.slf4j.LoggerFactory;
 
 public class FlyteAdminClient {
 
-  private static final Logger log = LoggerFactory.getLogger(FlyteAdminClient.class);
+  private static final Logger LOG = LoggerFactory.getLogger(FlyteAdminClient.class);
   private static final String TRIGGERING_PRINCIPAL = "styx";
   private static final int USER_TRIGGERED_EXECUTION_NESTING = 0;
 
@@ -64,8 +65,9 @@ public class FlyteAdminClient {
       String name,
       IdentifierOuterClass.Identifier launchPlanId,
       ExecutionOuterClass.ExecutionMetadata.ExecutionMode executionMode,
-      Map<String, String> annotations) {
-    log.debug("createExecution {} {} {}", project, domain, launchPlanId);
+      Map<String, String> annotations,
+      String parameter) {
+    LOG.debug("createExecution {} {} {}", project, domain, launchPlanId);
 
     var metadata =
         ExecutionOuterClass.ExecutionMetadata.newBuilder()
@@ -73,6 +75,14 @@ public class FlyteAdminClient {
             .setPrincipal(TRIGGERING_PRINCIPAL)
             .setNesting(USER_TRIGGERED_EXECUTION_NESTING)
             .build();
+
+    var launchPlan =
+        stub.getLaunchPlan(
+            Common.ObjectGetRequest.newBuilder().
+                setId(launchPlanId).
+                build());
+
+    var inputs = launchPlan.getSpec().getDefaultInputs();
 
     var spec =
         ExecutionOuterClass.ExecutionSpec.newBuilder()
@@ -90,7 +100,7 @@ public class FlyteAdminClient {
                 .setProject(project)
                 .setName(name)
                 .setSpec(spec)
-                // TODO: Pass inputs
+                .setInputs(fillingParameterInInputs(inputs, parameter))
                 .build());
 
     verifyNotNull(
@@ -105,35 +115,31 @@ public class FlyteAdminClient {
   }
 
   public ExecutionOuterClass.Execution getExecution(String project, String domain, String name) {
-    log.debug("getExecution {} {} {}", project, domain, name);
-    var request = ExecutionOuterClass.WorkflowExecutionGetRequest
-        .newBuilder()
-        .setId(IdentifierOuterClass.WorkflowExecutionIdentifier
-            .newBuilder()
-            .setProject(project)
-            .setDomain(domain)
-            .setName(name)
-            .build())
-        .build();
+    LOG.debug("getExecution {} {} {}", project, domain, name);
+    var request =
+        ExecutionOuterClass.WorkflowExecutionGetRequest.newBuilder()
+            .setId(
+                IdentifierOuterClass.WorkflowExecutionIdentifier.newBuilder()
+                    .setProject(project)
+                    .setDomain(domain)
+                    .setName(name)
+                    .build())
+            .build();
     return stub.getExecution(request);
   }
 
   public ExecutionOuterClass.ExecutionTerminateResponse terminateExecution(
-      String project,
-      String domain,
-      String name,
-      String cause) {
-    log.debug("terminateExecution {} {} {}", project, domain, name);
+      String project, String domain, String name, String cause) {
+    LOG.debug("terminateExecution {} {} {}", project, domain, name);
 
     final var request =
-        ExecutionOuterClass.ExecutionTerminateRequest
-            .newBuilder()
-            .setId(IdentifierOuterClass.WorkflowExecutionIdentifier
-                .newBuilder()
-                .setProject(project)
-                .setDomain(domain)
-                .setName(name)
-                .build())
+        ExecutionOuterClass.ExecutionTerminateRequest.newBuilder()
+            .setId(
+                IdentifierOuterClass.WorkflowExecutionIdentifier.newBuilder()
+                    .setProject(project)
+                    .setDomain(domain)
+                    .setName(name)
+                    .build())
             .setCause(cause)
             .build();
 
@@ -146,7 +152,7 @@ public class FlyteAdminClient {
       int limit,
       String token,
       String filters) {
-    log.debug("listExecutions {} {} {} {}", project, domain, limit, token);
+    LOG.debug("listExecutions {} {} {} {}", project, domain, limit, token);
 
     final var request = Common.ResourceListRequest
         .newBuilder()
@@ -165,7 +171,7 @@ public class FlyteAdminClient {
   }
 
   public ProjectOuterClass.Projects listProjects() {
-    log.debug("listProjects");
+    LOG.debug("listProjects");
 
     return stub.listProjects(ProjectOuterClass.ProjectListRequest.getDefaultInstance());
   }
--- a/styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
+++ b/styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
@@ -22,7 +22,7 @@ package com.spotify.styx.flyte.client;
 
 import static com.google.common.base.Strings.nullToEmpty;
 import static com.google.common.base.Verify.verifyNotNull;
-import static com.spotify.styx.flyte.client.FlyteInputsUtils.fillingParameterInInputs;
+import static com.spotify.styx.flyte.client.FlyteInputsUtils.fillParameterInInputs;
 
 import flyteidl.admin.Common;
 import flyteidl.admin.ExecutionOuterClass;
@@ -100,7 +100,7 @@ public class FlyteAdminClient {
                 .setProject(project)
                 .setName(name)
                 .setSpec(spec)
-                .setInputs(fillingParameterInInputs(inputs, parameter))
+                .setInputs(fillParameterInInputs(inputs, parameter))
                 .build());
 
     verifyNotNull(
--- a/styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
+++ b/styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
@@ -27,6 +27,7 @@ import static com.spotify.styx.flyte.client.FlyteInputsUtils.fillParameterInInpu
 import flyteidl.admin.Common;
 import flyteidl.admin.ExecutionOuterClass;
 import flyteidl.admin.ProjectOuterClass;
+import flyteidl.admin.LaunchPlanOuterClass;
 import flyteidl.core.IdentifierOuterClass;
 import flyteidl.service.AdminServiceGrpc;
 import io.grpc.ManagedChannelBuilder;
@@ -76,11 +77,7 @@ public class FlyteAdminClient {
             .setNesting(USER_TRIGGERED_EXECUTION_NESTING)
             .build();
 
-    var launchPlan =
-        stub.getLaunchPlan(
-            Common.ObjectGetRequest.newBuilder().
-                setId(launchPlanId).
-                build());
+    var launchPlan = getLaunchPlan(launchPlanId);
 
     var inputs = launchPlan.getSpec().getDefaultInputs();
 
@@ -112,6 +109,14 @@ public class FlyteAdminClient {
         launchPlanId);
 
     return response;
+  }
+
+  LaunchPlanOuterClass.LaunchPlan getLaunchPlan(IdentifierOuterClass.Identifier launchPlanId) {
+    LOG.debug("getLaunchPlan {}", launchPlanId);
+    var request = Common.ObjectGetRequest.newBuilder()
+        .setId(launchPlanId)
+        .build();
+    return stub.getLaunchPlan(request);
   }
 
   public ExecutionOuterClass.Execution getExecution(String project, String domain, String name) {
--- a/styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
+++ b/styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
@@ -26,8 +26,11 @@ import com.google.protobuf.Timestamp;
 import com.spotify.styx.util.ParameterUtil;
 import flyteidl.core.Interface;
 import flyteidl.core.Literals;
+import java.time.DateTimeException;
 import java.time.Instant;
 import java.time.format.DateTimeParseException;
+import java.util.List;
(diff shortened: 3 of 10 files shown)


styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.protobuf.Timestamp;
+import com.spotify.styx.util.ParameterUtil;
+import flyteidl.core.Interface;
+import flyteidl.core.Literals;
+import java.time.Instant;
+import java.time.format.DateTimeParseException;
+
+
+public class FlyteInputsUtils {
+  private static final String PARAMETER_NAME = "parameter";
```
ckiosidis, line 34: can't this be styx_parameter?

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -0,0 +1,84 @@
...
+  static Literals.LiteralMap fillingParameterInInputs(Interface.ParameterMap parameterMap, String parameter) {
+    var literalMapBuilder = Literals.LiteralMap.newBuilder();
+    parameterMap
+        .getParametersMap()
+        .forEach(
+            (key, value) -> {
+              if (key.toLowerCase().equals(PARAMETER_NAME)) {
+                if (!value.getDefault().hasScalar()
+                    || !value.getDefault().getScalar().hasPrimitive()
+                    || value.getDefault().getScalar().getPrimitive().getValueCase().getNumber()
```
ckiosidis, line 57: should we throw exceptions instead of ignoring those? It will be also sligthly cleaner

ckiosidis: now it's a bit difficult to reason with the nested ifs

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -0,0 +1,84 @@
...
+  private FlyteInputsUtils() {
+    throw new UnsupportedOperationException();
+  }
+
+  static Literals.Literal buildLiteralForPartition(String value) {
+
+    var primitiveBuilderForString = Literals.Primitive.newBuilder().setDatetime(toTimestamp(value));
+    return Literals.Literal.newBuilder()
+        .setScalar(Literals.Scalar.newBuilder().setPrimitive(primitiveBuilderForString.build()).build())
+        .build();
```
ckiosidis, line 45: is the parameter set as a string and not a datetime?

ckiosidis: maybe the name is confusing `primitiveBuilderForString`

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -0,0 +1,81 @@
...
+            (key, value) -> {
+              if (key.toLowerCase().equals(PARAMETER_NAME)) {
+                if (value.getVar().getType().getSimple() != DATETIME) {
+                  literalMapBuilder.putLiterals(key, value.getDefault());
+                } else {
+                  literalMapBuilder.putLiterals(key, buildLiteralForPartition(parameter));
+                }
+              } else {
+                literalMapBuilder.putLiterals(key, value.getDefault());
+              }
```
ckiosidis, line 62: would something like that be a bit clearer:

```suggestion
              if (key.toLowerCase().equals(PARAMETER_NAME)) {
                literalMapBuilder.putLiterals(key, value.getDefault());
                return;
              }

              if (value.getVar().getType().getSimple() != DATETIME) {
                  literalMapBuilder.putLiterals(key, value.getDefault());
                  return;
              
              literalMapBuilder.putLiterals(key, buildLiteralForPartition(parameter));
              
              }
```

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
ckiosidis: would it make sense to use the one we've been using so far? like `STYX_PARAMETER` ?

styx-flyte-client/src/test/java/com/spotify/styx/flyte/client/FlyteInputsUtilsTest.java
```diff
@@ -0,0 +1,71 @@
...
+import java.time.Instant;
+import org.junit.Test;
+
+
+public class FlyteInputsUtilsTest {
+  static final Instant INSTANT = Instant.now();
+  static final String PARAMETER = INSTANT.toString();
+
+  @Test
+  public void shouldFillingParameterToInputs() {
```
ckiosidis, line 41: can we add a test where we check the default values as well?

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
tnsetting: You know when I wrote this I still have the big for loop in my mind. 😄

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -0,0 +1,81 @@
...
+    parameterMap
+        .getParametersMap()
+        .forEach(
+            (key, value) -> {
+              if (!key.toLowerCase().equals(PARAMETER_NAME)) {
+                literalMapBuilder.putLiterals(key, value.getDefault());
+                return;
+              }
+              if (value.getVar().getType().getSimple() != DATETIME) {
+                literalMapBuilder.putLiterals(key, value.getDefault());
```
sonjaer, line 59: Not that important but,
Should this be allowed? Should we allow to not have styx_parameter as a launch plan input?

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -0,0 +1,81 @@
...
+              }
+            });
+
+    return literalMapBuilder.build();
+  }
+
+  private static Timestamp toTimestamp(String value) {
+    Instant instant;
+    try {
+      instant = ParameterUtil.parseDate(value);
```
sonjaer, line 71: Not that important either but maybe it would be nice to write some functions to check which formatter to use

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
```diff
@@ -70,6 +72,14 @@ public static FlyteAdminClient create(String target, boolean insecure) {
             .setNesting(USER_TRIGGERED_EXECUTION_NESTING)
             .build();
 
+    var launchPlan =
+        stub.getLaunchPlan(
```
sonjaer, line 76: Should we write a getLaunchPlan function?

sonjaer: Should we handle exceptions from this call?

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
```diff
@@ -83,7 +93,7 @@ public static FlyteAdminClient create(String target, boolean insecure) {
                 .setProject(project)
                 .setName(name)
                 .setSpec(spec)
-                // TODO: Pass inputs
+                .setInputs(fillingParameterInInputs(inputs, parameter))
```
sonjaer, line 96: Should we handle exceptions from this?

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  static Literals.LiteralMap fillingParameterInInputs(Interface.ParameterMap parameterMap, String parameter) {
+    var literalMapBuilder = Literals.LiteralMap.newBuilder();
+    parameterMap
+        .getParametersMap()
+        .forEach(
+            (key, value) -> {
+              if (!key.toLowerCase().equals(PARAMETER_NAME)) {
+                literalMapBuilder.putLiterals(key, value.getDefault());
```
sonjaer, line 55: Should we check if the value is required to be filled and log+throw an exception if it is?

styx-flyte-client/src/test/java/com/spotify/styx/flyte/client/FlyteInputsUtilsTest.java
```diff
@@ -0,0 +1,98 @@
...
```
sonjaer, line 43: Rename to maybe: shouldFillParameterInInputs

styx-flyte-client/src/test/java/com/spotify/styx/flyte/client/FlyteInputsUtilsTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    assertThat(INSTANT.getEpochSecond(), equalTo(inputs.getLiteralsMap()
+        .get(PARAMETER_NAME)
+        .getScalar()
+        .getPrimitive()
+        .getDatetime().getSeconds()));
+  }
+
+  @Test
+  public void shouldOnlyFillingParameterToInputs() {
```
sonjaer, line 65: Rename also to: shouldOnlyFillParameterInInputs

styx-flyte-client/src/test/java/com/spotify/styx/flyte/client/TestAdminService.java
```diff
@@ -50,6 +53,24 @@ public void createExecution(final ExecutionOuterClass.ExecutionCreateRequest req
...
+  public void getLaunchPlan(final flyteidl.admin.Common.ObjectGetRequest request,
+                            final StreamObserver<LaunchPlanOuterClass.LaunchPlan> responseObserver) {
+    responseObserver.onNext(LaunchPlanOuterClass.LaunchPlan
+        .newBuilder()
+        .setId(request.getId())
+        .setSpec(LaunchPlanOuterClass.LaunchPlanSpec
+            .newBuilder()
+            .setDefaultInputs(Interface.ParameterMap.newBuilder()
+                .putParameters("parameter", Interface.Parameter.newBuilder()
+                    .setDefault(buildLiteralForPartition("2020-09-21"))
```
sonjaer, line 66: Maybe we don't need to pass any inputs here, or we could pass "styx_parameter" or PARAMETER_NAME constant

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
sonjaer: nvm, I saw that this was handled in the function calling this

styx-flyte-client/src/test/java/com/spotify/styx/flyte/client/FlyteInputsUtilsTest.java
```diff
@@ -0,0 +1,98 @@
...
+    var inputName = "xxx";
+    var parameterMap = Interface.ParameterMap.newBuilder()
+        .putParameters(inputName, Interface.Parameter.newBuilder()
+            .setVar(Interface.Variable.newBuilder()
+                .setType(Types.LiteralType.newBuilder().
+                    setSimple(Types.SimpleType.DATETIME)
+                    .build())
+                .build())
+            .setDefault(buildLiteralForPartition(PARAMETER))
```
sonjaer, line 74: Maybe we want to avoid reusing this for clarity

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
sonjaer, line 96: maybe rename to fillParameterInInputs

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java
sonjaer, line 76: Should we write a getLaunchPlan function in the FlyteAdminClient to be consistent?

ckiosidis: yes sounds good

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -0,0 +1,81 @@
...
+              } else {
+                literalMapBuilder.putLiterals(key, buildLiteralForPartition(parameter));
+              }
+            });
+
+    return literalMapBuilder.build();
+  }
+
+  private static Timestamp toTimestamp(String value) {
```
sonjaer, line 69: Wondering if we can rewrite this function to look a bit nicer somehow?

sonjaer: And if we can write unit tests for it

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
sonjaer: I went ahead and changed this to handle this [oneof behavior](https://github.com/lyft/flyteidl/blob/793b09d190148236f41ad8160b5cec9a3325c16f/protos/flyteidl/core/interface.proto#L38) since @ckiosidis agreed. Let me know what you think. @tnsetting @ckiosidis

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -56,7 +56,16 @@ private FlyteInputsUtils() {
         .forEach(
             (key, value) -> {
               if (!key.toLowerCase().equals(PARAMETER_NAME)) {
-                literalMapBuilder.putLiterals(key, value.getDefault());
+                if (value.hasDefault()) {
```
ckiosidis, line 59: have we verified that this will be populated in the LP?

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
sonjaer: I also changed this since @ckiosidis agreed @tnsetting @ckiosidis

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
sonjaer: Yes I believe so. 
`
* @return Whether the default field is set.
*/
public boolean hasDefault() {
    return behaviorCase_ == 2;
}
`

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -0,0 +1,96 @@
...
+                  literalMapBuilder.putLiterals(key, value.getDefault());
+                  return;
+                }
+                if (value.getRequired()) {
+                  var message = "LP inputs must have default values. Missing default value for key:" + key;
+                  LOG.error(message);
+                  throw new UnsupportedOperationException(message);
+                }
+                LOG.debug("Ignore key: {}", key);
```
tnsetting, line 68: Should be an exception here since an input field should always have some value.

sonjaer: If it is not required I think it would be fine to ignore it as we don't allow user defined values yet?

sonjaer: that makes sense

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
sonjaer: Simplified after discussion. We only allow inputs with default parameters for now.

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
sonjaer: Addressed this

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
sonjaer: Fixed this

tnsetting (commented):
LGTM

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -73,24 +76,30 @@ private FlyteInputsUtils() {
...
+        ParameterUtil::parseDateYear,
+        Instant::parse
+    );
+    for (Function<String, Instant> function : functions) {
       try {
-        instant = ParameterUtil.parseDateHour(value);
-      } catch (DateTimeParseException ignored2) {
-        instant = Instant.parse(value);
-      }
+        Instant instant = function.apply(value);
```
ckiosidis, line 93: why apply all the functions here?

ckiosidis (approved):

styx-common/src/main/java/com/spotify/styx/util/ParameterUtil.java
```diff
@@ -109,6 +109,18 @@ public static Instant parseDateHour(String dateHour) {
         ISO_LOCAL_DATE_HOUR.parse(dateHour)).atOffset(UTC));
   }
 
+  public static Instant parseDateMonth(String dateMonth) {
+    return Instant.from(LocalDate.from(
+        DateTimeFormatter.ISO_LOCAL_DATE.parse(dateMonth + "-01"))
+        .atStartOfDay(UTC));
```
narape, line 115: We can reduce duplication by basing this on the other function
```suggestion
    return parseDate(dateMonth + "-01");
```

styx-common/src/main/java/com/spotify/styx/util/ParameterUtil.java
```diff
@@ -109,6 +109,18 @@ public static Instant parseDateHour(String dateHour) {
...
+  }
+
+  public static Instant parseDateYear(String dateYear) {
+    return Instant.from(LocalDate.from(
+        DateTimeFormatter.ISO_LOCAL_DATE.parse(dateYear + "-01-01"))
+        .atStartOfDay(UTC));
```
narape, line 121: same here
```suggestion
    return parseDate(dateYear + "-01-01");
```

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -73,24 +76,30 @@ private FlyteInputsUtils() {
...
     return literalMapBuilder.build();
   }
 
-  private static Timestamp toTimestamp(String value) {
-    Instant instant;
-    try {
-      instant = ParameterUtil.parseDate(value);
-    } catch (DateTimeParseException ignored1) {
+  static Timestamp toTimestamp(String value) {
+    List<Function<String, Instant>> functions = List.of(
```
narape, line 84: I would say that it would be better to have this code inside a method in `ParameterUtil`

```java
  public static Instant parseBest(String value) {
  }
```
Then we will add a lot of unit testing for this method and future refactoring or optimizations would be highly localized

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
narape: She is going from less specific to more specific because we don't know which of this applies

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
sonjaer: Will do this in a following PR

PR #866, opened 2020-09-23
Refactor parameter to timestamp function

sonjaer:

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/util/ParameterUtil.java
+++ b/styx-common/src/main/java/com/spotify/styx/util/ParameterUtil.java
@@ -27,7 +27,9 @@ import static java.time.ZoneOffset.UTC;
 import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
 import static java.time.temporal.ChronoField.MONTH_OF_YEAR;
 
+import com.google.protobuf.Timestamp;
 import com.spotify.styx.model.Schedule;
+import java.time.DateTimeException;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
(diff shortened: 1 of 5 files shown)


styx-common/src/main/java/com/spotify/styx/util/ParameterUtil.java
```diff
@@ -110,15 +117,11 @@ public static Instant parseDateHour(String dateHour) {
   }
 
   public static Instant parseDateMonth(String dateMonth) {
-    return Instant.from(LocalDate.from(
-        DateTimeFormatter.ISO_LOCAL_DATE.parse(dateMonth + "-01"))
-        .atStartOfDay(UTC));
+    return parseDate(dateMonth + "-01");
```
ckiosidis, line 120: doesn't this work without the suffix "-01" ?

ckiosidis: nvm

ckiosidis (approved):

styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteInputsUtils.java
```diff
@@ -46,7 +40,7 @@ private FlyteInputsUtils() {
   static Literals.Literal buildLiteralForPartition(String value) {
 
     var primitive =
-        Literals.Primitive.newBuilder().setDatetime(toTimestamp(value)).build();
+        Literals.Primitive.newBuilder().setDatetime(parseBest(value)).build();
```
honnix, line 43: Can we get the `schedule` here from workflow definition? If yes we don't need to do this kind of best matching parsing which is both time consuming and error prone.

PR #868, opened 2020-09-24
Terminate dangling styx related executions

narape:

## Description
Terminate dangling styx related executions. A dangling executions is one running in Flyte with styx annotation
whose RunState is not longer active in Styx.

## Motivation and Context
A user could emit a halt command and force a RunState being on Error and it won't be active.

I have included unit tests. Will test in staging.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -639,7 +639,7 @@ public class StyxScheduler implements AppInit {
     }
 
     var flyteAdminClient = getFlyteAdminClient(config, runnerId);
-    return new FlyteAdminClientRunner(runnerId, flyteAdminClient, stateManager);
+    return FlyteRunner.flyteAdmin(runnerId, flyteAdminClient, stateManager);
   }
 
   private static FlyteAdminClient getFlyteAdminClient(Config rootConfig, String runnerId) {
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -57,9 +57,8 @@ import com.spotify.styx.api.SchedulerResource;
 import com.spotify.styx.api.ServiceAccountUsageAuthorizer;
 import com.spotify.styx.api.WorkflowActionAuthorizer;
 import com.spotify.styx.docker.DockerRunner;
-import com.spotify.styx.docker.RunnerId;
 import com.spotify.styx.docker.Fabric8KubernetesClient;
-import com.spotify.styx.flyte.FlyteAdminClientRunner;
+import com.spotify.styx.docker.RunnerId;
 import com.spotify.styx.flyte.FlyteRunner;
 import com.spotify.styx.flyte.client.FlyteAdminClient;
 import com.spotify.styx.model.Event;
@@ -106,13 +105,10 @@ import com.spotify.styx.util.StorageFactory;
 import com.spotify.styx.util.Time;
 import com.spotify.styx.util.TriggerUtil;
 import com.typesafe.config.Config;
-import flyteidl.service.AdminServiceGrpc;
 import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.NamespacedKubernetesClient;
 import io.fabric8.kubernetes.client.utils.HttpClientUtils;
-import io.grpc.ManagedChannel;
-import io.grpc.ManagedChannelBuilder;
 import java.io.IOException;
 import java.security.GeneralSecurityException;
 import java.time.Duration;
@@ -649,15 +645,9 @@ public class StyxScheduler implements AppInit {
     }
 
     var config = flyteAdminRootConfig.getConfig(runnerId);
-    var builder =
-        ManagedChannelBuilder.forAddress(config.getString(FLYTEADMIN_HOST), config.getInt(FLYTEADMIN_PORT));
-    if (config.getBoolean(FLYTEADMIN_INSECURE)) {
-      builder.usePlaintext();
-    }
-    final ManagedChannel channel = builder.build();
-
-    var stub = AdminServiceGrpc.newBlockingStub(channel);
-    return new FlyteAdminClient(stub);
+    final var target = config.getString(FLYTEADMIN_HOST) + ":" + config.getInt(FLYTEADMIN_PORT);
+    final var insecure = config.getBoolean(FLYTEADMIN_INSECURE);
+    return FlyteAdminClient.create(target, insecure);
   }
 
   @VisibleForTesting
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -168,6 +168,11 @@ public class FlyteAdminClientRunner implements FlyteRunner {
     return trigger.accept(TriggerToExecutionModeVisitor.INSTANCE);
   }
 
+  @Override
+  public void close() {
+    // Nothing to close... yet
+  }
+
   private static class TriggerToExecutionModeVisitor implements TriggerVisitor<ExecutionMode> {
     private static final TriggerToExecutionModeVisitor INSTANCE = new TriggerToExecutionModeVisitor();
 
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -221,7 +221,7 @@ public class FlyteAdminClientRunner implements FlyteRunner {
       // TODO: add tracing
       tryTerminateDanglingFlyteExecutions();
     } catch (Throwable t) {
-      LOG.warn("Error while terminating dangling flyte executions", t);
+      LOG.error("Error while terminating dangling flyte executions", t);
     }
 
   }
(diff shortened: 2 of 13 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -152,7 +209,95 @@ public void poll(FlyteExecutionId flyteExecutionId, RunState runState)
...
+      LOG.warn("Error while terminating dangling flyte executions", t);
+    }
+
+  }
+
+  private void tryTerminateDanglingFlyteExecutions() {
+    var projects = flyteAdminClient.listProjects();
+
+    var allFutures = new ArrayList<CompletableFuture<Void>>();
+    for (var project : projects.getProjectsList()) {
```
tnsetting, line 233: Do we want to get all executions every time? I think it is going to be a huge when the system is running a while.

styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -639,7 +639,7 @@ static FlyteRunner createFlyteRunner(String runnerId, Config config, StateManage
     }
 
     var flyteAdminClient = getFlyteAdminClient(config, runnerId);
-    return new FlyteAdminClientRunner(runnerId, flyteAdminClient, stateManager);
+    return FlyteRunner.flyteAdmin(runnerId, flyteAdminClient, stateManager);
```
ckiosidis, line 638: 👍

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteRunner.java
```diff
@@ -43,6 +45,12 @@ static FlyteRunner noop() {
     return new NoopFlyteRunner();
   }
 
+  static FlyteRunner flyteAdmin(final String runnerId,
```
ckiosidis, line 48: should we call this create?

ckiosidis: actually there is a create method in the FlyteAdminClient, can we refactor to use it as well? https://github.com/spotify/styx/blob/master/styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java#L51

narape: `FlyteRunner` contains several factory methods: `flyteAdmin`, `noop` and `routing`, each of them returning a different class

ckiosidis: i see 👍

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: I was thinking trying out filters to see if I can retrieve only running executions with styx annotations. But I was planning to do that in a different PR.

Other than that, what is your proposal? just doing a few projects every "tick"? we will need to keep state of what we have done and what is still missing. 

On the other hand, we already [iterate over all running Pods](https://github.com/spotify/styx/blob/6a79ce45715076074123ca5662458077bbe6a939/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java#L568) in KubernetesRunner.

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -152,7 +209,95 @@ public void poll(FlyteExecutionId flyteExecutionId, RunState runState)
...
+    scheduleWithJitter(this::terminateDanglingFlyteExecutions, scheduledExecutor, terminateDanglingFlyteExecInterval);
+  }
+
+  @VisibleForTesting
+  void terminateDanglingFlyteExecutions() {
+    try {
+      // TODO: add tracing
+      tryTerminateDanglingFlyteExecutions();
+    } catch (Throwable t) {
+      LOG.warn("Error while terminating dangling flyte executions", t);
```
ckiosidis, line 224: let's log an error here?

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteRunner.java
narape: > actually there is a create method in the FlyteAdminClient, can we refactor to use it as well? https://github.com/spotify/styx/blob/master/styx-flyte-client/src/main/java/com/spotify/styx/flyte/client/FlyteAdminClient.java#L51

I did in the next commit

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -152,7 +209,95 @@ public void poll(FlyteExecutionId flyteExecutionId, RunState runState)
...
+  private void tryTerminateDanglingFlyteExecutions() {
+    var projects = flyteAdminClient.listProjects();
+
+    var allFutures = new ArrayList<CompletableFuture<Void>>();
+    for (var project : projects.getProjectsList()) {
+      for (var domain : project.getDomainsList()) {
+        String paginationToken = null;
+        do {
+          var executions =
+              flyteAdminClient.listExecutions(project.getId(), domain.getId(), 100, paginationToken, "");
```
ckiosidis, line 238: do we care about all projects and *all* domains equally?

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
tnsetting: The executions are a bit different than Pods as k8s Pods will be cleaned up but executions will not so it will be accumulated. 

I think if the executions we are getting are sorted by time order, we can keep track the time of last execution that has been cleaned up and we don't need to going further in history.

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -152,7 +209,95 @@ public void poll(FlyteExecutionId flyteExecutionId, RunState runState)
...
+          paginationToken = executions.getToken();
+        } while (!"".equals(paginationToken));
+      }
+    }
+    allFutures.forEach(CompletableFuture::join);
+  }
+
+  private void tryTerminateDanglingFlyteExecution(AnnotatedFlyteExecutionId annotatedId) {
+    var workflowInstance = getWorkflowInstance(annotatedId);
+    var runState = stateManager.getActiveState(workflowInstance);
```
ckiosidis, line 257: are these operations  becoming a load concern?

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: :shrug:  What is your idea? doing production first, then staging and next development? Or are we sure that styx will only execute launch plans in production? SO far we are not enforcing that (I think)

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -152,7 +209,95 @@ public void poll(FlyteExecutionId flyteExecutionId, RunState runState)
...
+
+    var shouldTerminate = runState.isEmpty() // state reached terminal state and aren't  active anymore
+                          || !isFlyteExecRelatedToRunState(annotatedId, runState.orElseThrow());
```
ckiosidis, line 260: why throw an exception here? If there is no runstate from the statemanager, what does it mean? Should we just terminate it?

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: We are doing [the same](https://github.com/spotify/styx/blob/6a79ce45715076074123ca5662458077bbe6a939/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java#L589) in `KubernetesDockerRunner`

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: It would never throw an exception, this is a OR and java doesn't evaluate the right side if the left one is `true`. so if `runState.isEmpty()` is `true` then `shouldTerminate` is `true` and the right hand side doesn't get evaluated. On the other hand, if `runState.isEmpty()` is 'false` then `runState.orElseThrow()` will never throw exception.

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
tnsetting: So for each flyte workflow we keep track of the last clean up execution time and update it if we cleanup some executions. What do you think?

narape: Interesting idea, do we need to keep a time for every workflow? wouldn't last cleanup date would be enough?

ckiosidis (approved):
LGTM
We can always make it more performant but I we have room to do that in the future if it proves to be a problem.
Limiting both the executions that we check, or maybe going the other way around and not going through the whole project/domain execution list could be solving any potential performance struggles

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
ckiosidis: yes, let's keep it like that for now. until we something comes up. It's good to avoid "early optimisation" i guess

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: We talk in a video call and we agreed to reduce the number of execution retrieved from Flyte in a different PR.
As a first alternative would be to leverage filters to see if we can get only running executions or in the worst case execution not older than x time (1.5 days maybe).

PR #886, opened 2020-10-28
Port forwarding to flyteadmin for e2e test

tnsetting:

## Description
Setup portforwarding to flyteadmin in the test cluster for e2e test by using kubectl. So in the e2e test, styx just connect to flyteadmin on 127.0.0.1:8089. So there is no need to change flyte system deployment to open up the system to the internet.

## Motivation and Context
For e2e with flyte component

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,6 @@ jobs:
             chmod +x grpc_health/grpc_health_probe
             echo 'dc5c8cd7b33ef503f49eb639af6017f33d340884  grpc_health/grpc_health_probe' > .grpc_checksum
             sha1sum -c .grpc_checksum
-            nohup google-cloud-sdk/bin/kubectl -n flyte port-forward deployment/flyteadmin 8089:8089 &
       - run:
           name: Tests
           environment:
@@ -63,6 +62,8 @@ jobs:
             # Maven heap size
             MAVEN_OPTS: -Xmx256m -Xms256m
           command: |
+            google-cloud-sdk/bin/kubectl -n flyte port-forward deployment/flyteadmin 8089:8089 &
+            # it takes some time for the portforwarding to be ready
             grpc_health/grpc_health_probe -addr 127.0.0.1:8089  -connect-timeout 8s
             source google-cloud-sdk/path.bash.inc
             mvn -B verify -P disable-static-analysis
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,14 @@ jobs:
             google-cloud-sdk/bin/gcloud container clusters get-credentials styx-e2e-test --zone europe-west4-c --project styx-oss-test
       # Run tests without static analysis
       - run:
+          name: "Install grpc_health_probe"
+          command: |
+            mkdir grpc_health
+            curl --silent --fail --show-error --location --output grpc_health/grpc_health_probe "https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.3.2/grpc_health_probe-linux-amd64"
+            chmod +x grpc_health/grpc_health_probe
+            echo 'dc5c8cd7b33ef503f49eb639af6017f33d340884  grpc_health/grpc_health_probe' > .grpc_checksum
+            sha1sum -c .grpc_checksum
+      - run:
           name: Tests
           environment:
             # Configure datastore emulator heap size
@@ -56,7 +64,15 @@ jobs:
           command: |
             google-cloud-sdk/bin/kubectl -n flyte port-forward deployment/flyteadmin 8089:8089 &
             # it takes some time for the portforwarding to be ready
-            sleep 5
+            max_retry=10
+            counter=0
+            until ./grpc_health/grpc_health_probe -addr 127.0.0.1:8089
+            do
+            	sleep 1
+            	[[ counter -eq $max_retry ]] && echo "Failed!" && exit 1
+            	 echo "Trying again. Try #$counter"
+            	((counter++))
+            done
             source google-cloud-sdk/path.bash.inc
             mvn -B verify -P disable-static-analysis
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
               google-cloud-sdk/install.sh -q
               google-cloud-sdk/bin/gcloud components install beta cloud-datastore-emulator -q
               google-cloud-sdk/bin/gcloud components install kubectl -q
-              google-cloud-sdk/bin/gcloud config set project testing
+              google-cloud-sdk/bin/gcloud config set project styx-oss-test
             fi
       - save_cache:
           key: v4-deps-{{ checksum "pom.xml" }}-{{ checksum "gcloud-sdk-url.txt" }}
@@ -41,7 +41,7 @@ jobs:
             echo "export GOOGLE_APPLICATION_CREDENTIALS=\"$(realpath credentials.json)\"" >> $BASH_ENV
             google-cloud-sdk/bin/gcloud auth activate-service-account --key-file=credentials.json
       - run:
-          name: "Set up k8s cluster"
+          name: "Set up GKE cluster credentials"
           command: |
             google-cloud-sdk/bin/kubectl version --client
             google-cloud-sdk/bin/gcloud container clusters get-credentials styx-e2e-test --zone europe-west4-c --project styx-oss-test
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -26,8 +26,13 @@ jobs:
               tar xzf google-cloud-sdk-*.tar.gz
               google-cloud-sdk/install.sh -q
               google-cloud-sdk/bin/gcloud components install beta cloud-datastore-emulator -q
-              google-cloud-sdk/bin/gcloud config set project testing
             fi
+      - run:
+          name: "Install kubectl "
+          command: |
+            curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
```
narape, line 33: I would like to verify that the downloaded file has the proper shasum as I saw honnix doing in another repo
```Dockerfile
RUN curl --silent --fail --show-error --location --output /bin/grpc_health_probe "https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.3.2/grpc_health_probe-linux-amd64" && \
--
  | chmod +x /bin/grpc_health_probe && \
  | echo 'dc5c8cd7b33ef503f49eb639af6017f33d340884  /bin/grpc_health_probe' > .grpc_checksum && \
  | sha1sum -c .grpc_checksum
```
But we won't able to fetch the latest kubectl but ping it to a version

.circleci/config.yml
```diff
@@ -38,6 +43,14 @@ jobs:
...
             echo "$GCLOUD_SERVICE_KEY" > credentials.json
             echo "export GOOGLE_APPLICATION_CREDENTIALS=\"$(realpath credentials.json)\"" >> $BASH_ENV
+            google-cloud-sdk/bin/gcloud auth activate-service-account --key-file=credentials.json
+      - run:
+          name: "Set up k8s cluster"
+          command: |
+            kubectl version --client
+            google-cloud-sdk/bin/gcloud config set project styx-oss-test
+            google-cloud-sdk/bin/gcloud container clusters get-credentials styx-e2e-test --zone europe-west4-c
+            # google-cloud-sdk/bin/gcloud config set project testing
```
narape, line 53: this is no longer necessary

.circleci/config.yml
honnix: Does `gcloud components install kubectl` work?

.circleci/config.yml
tnsetting: Agree. Forget to remove it.

.circleci/config.yml
```diff
@@ -38,6 +43,14 @@ jobs:
           command: |
             echo "$GCLOUD_SERVICE_KEY" > credentials.json
             echo "export GOOGLE_APPLICATION_CREDENTIALS=\"$(realpath credentials.json)\"" >> $BASH_ENV
+            google-cloud-sdk/bin/gcloud auth activate-service-account --key-file=credentials.json
+      - run:
+          name: "Set up k8s cluster"
+          command: |
+            kubectl version --client
+            google-cloud-sdk/bin/gcloud config set project styx-oss-test
```
honnix, line 51: We don't need to change default settings. All gcloud commands accept `--project`.

.circleci/config.yml
```diff
@@ -26,8 +26,13 @@ jobs:
...
               google-cloud-sdk/install.sh -q
               google-cloud-sdk/bin/gcloud components install beta cloud-datastore-emulator -q
-              google-cloud-sdk/bin/gcloud config set project testing
             fi
+      - run:
+          name: "Install kubectl "
+          command: |
+            curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
+            chmod +x ./kubectl
+            mv ./kubectl /usr/local/bin/kubectl
```
honnix, line 35: I'm actually surprised this will be cached and brought to the next step.

.circleci/config.yml
```diff
@@ -47,6 +60,10 @@ jobs:
             # Maven heap size
             MAVEN_OPTS: -Xmx256m -Xms256m
           command: |
+            kubectl -n flyte port-forward deployment/flyteadmin 8089:8089 &
+            # it takes some time for the portforwarding to be ready
```
honnix, line 64: Do a spin check with timeout here to make sure it starts instead of guess-sleep?

.circleci/config.yml
```diff
@@ -47,6 +60,10 @@ jobs:
...
+            sleep 5
+            # make sure we can access flyteadmin
```
honnix, line 66:
```suggestion
```

.circleci/config.yml
```diff
@@ -26,8 +26,13 @@ jobs:
               tar xzf google-cloud-sdk-*.tar.gz
               google-cloud-sdk/install.sh -q
               google-cloud-sdk/bin/gcloud components install beta cloud-datastore-emulator -q
-              google-cloud-sdk/bin/gcloud config set project testing
```
honnix, line 29: Why did we do this?

.circleci/config.yml
honnix: If it does, we don't need this step as it can be in previous step when installing gcloud.

.circleci/config.yml
tnsetting: Yes. In that case we can install `grpc_health_probe` to make sure it starts.

.circleci/config.yml
tnsetting: I will check if the gcloud install kubeclt works

.circleci/config.yml
honnix: We don't restart/redeploy flyte every time we run e2e test, right? The spin check I mentioned here is to make sure port forwarding is ready. But using grpc_health_probe is more reliable.

.circleci/config.yml
tnsetting: 👍

.circleci/config.yml
tnsetting: No idea..

tnsetting: It looks like we have to set it to use the datastore emulator

honnix: That's strange because I don't need this locally and CI passed for your previous commit. Maybe just legacy code.

tnsetting: Maybe you have to set the project to something. Because in the other commit I set the project to flyte-oss-test but now I don't

honnix: Then let's set it to the real thing `styx-oss-test`?

.circleci/config.yml
```diff
@@ -38,7 +39,21 @@ jobs:
           command: |
             echo "$GCLOUD_SERVICE_KEY" > credentials.json
             echo "export GOOGLE_APPLICATION_CREDENTIALS=\"$(realpath credentials.json)\"" >> $BASH_ENV
+            google-cloud-sdk/bin/gcloud auth activate-service-account --key-file=credentials.json
+      - run:
+          name: "Set up k8s cluster"
```
honnix, line 44:
```suggestion
          name: "Set up GKE cluster credentials"
```

It managed to surprise me when seeing the name. 😮

.circleci/config.yml
```diff
@@ -26,6 +26,7 @@ jobs:
               tar xzf google-cloud-sdk-*.tar.gz
               google-cloud-sdk/install.sh -q
               google-cloud-sdk/bin/gcloud components install beta cloud-datastore-emulator -q
+              google-cloud-sdk/bin/gcloud components install kubectl -q
               google-cloud-sdk/bin/gcloud config set project testing
```
honnix, line 30:
```suggestion
              google-cloud-sdk/bin/gcloud config set project styx-oss-test
```

.circleci/config.yml
```diff
@@ -47,6 +62,17 @@ jobs:
             # Maven heap size
             MAVEN_OPTS: -Xmx256m -Xms256m
           command: |
+            google-cloud-sdk/bin/kubectl -n flyte port-forward deployment/flyteadmin 8089:8089 &
+            # it takes some time for the portforwarding to be ready
+            max_retry=10
+            counter=0
+            until ./grpc_health/grpc_health_probe -addr 127.0.0.1:8089
```
honnix, line 69: It seems we can use grpc_health_probe timeout directly than doing a loop here.

.circleci/config.yml
honnix: I misunderstood how CircleCI works. All the `runs` happen in the same container, not multiple ones.

tnsetting: Yes. It looks like so and this makes sense. But what I don't understand is that I need to do the port forwarding in the same step as running the test.

.circleci/config.yml
```diff
@@ -47,6 +62,9 @@ jobs:
             # Maven heap size
             MAVEN_OPTS: -Xmx256m -Xms256m
           command: |
+            google-cloud-sdk/bin/kubectl -n flyte port-forward deployment/flyteadmin 8089:8089 &
+            # it takes some time for the portforwarding to be ready
+            ./grpc_health/grpc_health_probe -addr 127.0.0.1:8089  -connect-timeout 8s
```
honnix, line 67: nit
```suggestion
            grpc_health/grpc_health_probe -addr 127.0.0.1:8089  -connect-timeout 8s
```

.circleci/config.yml
honnix: Probably. You can `nohup kubectl ... &` if you want to try.

tnsetting: it doesn't work :(

honnix (approved):

PR #907, opened 2020-12-14
Ignore expected exceptions

honnix:

## Description

Ignore expected exceptions and send in expected counter (instead of always accepting).

## Motivation and Context

Emitting an event may fail due to for example exhausted counter capacity, another schedule instance took the transaction, or the scheduler is being shutting down. In those case, it should be throw an exception.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
@@ -677,8 +677,6 @@ class KubernetesDockerRunner implements DockerRunner {
       }
 
       try {
-        // TODO: spoofing counter values like this can give unexpected results, e.g. if we emit two events here the
-        // first one might be discarded and the second one accepted.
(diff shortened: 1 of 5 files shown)


tnsetting (approved):

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -677,8 +677,6 @@ private void emitPodEvents(Pod pod, RunState runState) {
       }
 
       try {
-        // TODO: spoofing counter values like this can give unexpected results, e.g. if we emit two events here the
```
honnix, line 680: This comment has been addressed because if the first event gets rejected, the whole method would return without further trying.

tnsetting:
The `codecov/patch` error in build does not make sense as there is a test case for `CounterCapacityException` and `IsClosedException`.

honnix:
Strange, it was green a few commits back.

honnix:
Hmm, I think I broke it somewhere.

PR #912, opened 2020-12-29
Dequeue older partition first

honnix:

## Description

Dequeue older partition of the same workflow first.

## Motivation and Context

Currently dequeue order is completely shuffled without respecting partition age. This PR keeps shuffling as much as possible while prioritising older partition of the same workflow.

Unit test

## Checklist for PR author(s)

- [x] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -324,10 +324,10 @@ public class Scheduler {
    */
   @VisibleForTesting
   List<WorkflowInstance> shuffleInstances(Set<WorkflowInstance> activeInstances) {
-    var groups = activeInstances
+    final Map<WorkflowId, SortedSet<WorkflowInstance>> groups = activeInstances
         .stream()
         .collect(groupingBy(WorkflowInstance::workflowId, LinkedHashMap::new,
-            toCollection(() -> (SortedSet<WorkflowInstance>) new TreeSet<>(comparing(WorkflowInstance::parameter)))));
+            toCollection(() -> new TreeSet<>(comparing(WorkflowInstance::parameter)))));
 
     return merge(groups.values(), activeInstances.size());
   }
@@ -351,7 +351,7 @@ public class Scheduler {
     var merged = new WorkflowInstance[size];
 
     var from = 0;
-    for (SortedSet<WorkflowInstance> group : instanceGroups) {
+    for (var group : instanceGroups) {
       var subIndices = indices.subList(from, from + group.size()).stream().sorted().collect(toList());
       var j = 0;
       for (var instance : group) {
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -321,7 +321,8 @@ public class Scheduler {
   /**
    * Shuffle instances. For the same workflow ID, older instances will be appear early in the list.
    */
-  private List<WorkflowInstance> shuffleInstances(Set<WorkflowInstance> activeInstances) {
+  @VisibleForTesting
+  List<WorkflowInstance> shuffleInstances(Set<WorkflowInstance> activeInstances) {
     var groups = activeInstances
         .stream()
         .collect(groupingBy(WorkflowInstance::workflowId,
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
@@ -67,6 +67,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
```diff
@@ -150,8 +159,10 @@ public void setUp() throws Exception {
     when(stateManager.getActiveState(any())).then(a ->
         Optional.ofNullable(activeStates.get(a.<WorkflowInstance>getArgument(0))));
 
+    when(random.nextInt(anyInt())).then(a -> a.<Integer>getArgument(0) - 1);
```
honnix, line 162: This is dirty, but otherwise hard to test.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -309,6 +318,53 @@ private void processInstance(StyxConfig config, Map<String, Resource> resources,
...
   }
 
+  /**
+   * Shuffle instances. For the same workflow ID, older instances will appear early in the list.
+   */
+  @VisibleForTesting
+  List<WorkflowInstance> shuffleInstances(Set<WorkflowInstance> activeInstances) {
+    var groups = activeInstances
+        .stream()
+        .collect(groupingBy(WorkflowInstance::workflowId, LinkedHashMap::new,
```
narape, line 328: Using `LinkedHashMap` to have predictable iteration to simplify test

styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
narape: I factored out the shuffling to avoid using Random

styx-scheduler-service/src/test/java/com/spotify/styx/SchedulerTest.java
```diff
@@ -261,31 +263,51 @@ public void shouldExecuteRetryIfDelayIsReset() {
...
+            time.get().minus(j, ChronoUnit.SECONDS)));
+      }
+    }
 
-    populateActiveStates(RunState.create(INSTANCE_1, State.QUEUED, stateData, time.get()));
+    scheduler.tick();
 
-    List<WorkflowInstance> workflowInstances = new ArrayList<>();
+    var instancesOlderFirst = activeStates.keySet().stream()
+        .sorted(WorkflowInstance.KEY_COMPARATOR.thenComparing(WorkflowInstance::parameter));
```
honnix, line 282: I think I would prefer duplicating the group by and then sorting logic than this double sorting.

honnix: Hit enter too fast. To me the reason is to avoid doing the translation between group by + linked hashmap + sorting + merging and double sorting when reading the code. But I see the simplification here too. So not a strong opinion.

narape: I didn't want to use the grouping because I felt that we were reusing the code from production code. But I will remove the ordering and create for loops with the verification inside to be explicit that we are testing that older goes first

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
honnix: I guess this comment could live with the code because it would be hard to recall why order matters.

honnix:
Maybe we should really disable these flaky system tests as they are largely covered by e2e tests.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -204,8 +214,7 @@ private void processInstances(StyxConfig config, Map<String, Resource> resources
     var resourceExhaustedCache = new ConcurrentHashMap<String, Boolean>();
 
     // Shuffle the instances in order to process them in random order and reduce contention with other schedulers etc
-    var shuffledInstances = new ArrayList<>(activeInstances);
-    Collections.shuffle(shuffledInstances);
+    var shuffledInstances = shuffleInstances(activeInstances);
```
ckiosidis, line 217: this shuffle seems unnecessary now with the shuffle in the shuffleInstances method

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
```diff
@@ -309,6 +318,53 @@ private void processInstance(StyxConfig config, Map<String, Resource> resources,
...
+   */
+  @VisibleForTesting
+  List<WorkflowInstance> shuffleInstances(Set<WorkflowInstance> activeInstances) {
+    var groups = activeInstances
+        .stream()
+        .collect(groupingBy(WorkflowInstance::workflowId, LinkedHashMap::new,
+            toCollection(() -> new TreeSet<>(comparing(WorkflowInstance::parameter)))));
+
+    var shuffledGroups = new ArrayList<SortedSet<WorkflowInstance>>(groups.values());
+    shuffler.shuffle(shuffledGroups);
```
ckiosidis, line 332: this shuffle seems unnecessary now with the shuffle in the shuffleInstances method

honnix: Maybe or maybe not. I guess it doesn't hurt with more randomness.

ckiosidis (approved):
LGTM.
Added one comment for the extra (unnecessary) shuffle

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
narape: I am with babis here, the shuffle at merge is directing which workflow goes to what index, a previous shuffle won't improve the randomness here, so it is unnecessary

honnix:
Another thing to consider is, users will come up with good reasons that newer partitions should be handled first. Still, even in that case, having this kind of determinism is still better than fully shuffled solution.

styx-scheduler-service/src/main/java/com/spotify/styx/Scheduler.java
honnix: 👍

honnix:
🎉

PR #915, opened 2021-01-25
Ignore 400 errors until google fix their api

sonjaer:

Ignore 400 errors until google fix their api. Users may sometimes clean up keys associated with SAs. When deleting a non-existent key we receive 400 from google which cause the workflow to fail with:

```
java.util.concurrent.ExecutionException: com.google.api.client.googleapis.json.GoogleJsonResponseException: 400 Bad Request
{
  "code" : 400,
  "errors" : [ {
    "domain" : "global",
    "message" : "Precondition check failed.",
    "reason" : "failedPrecondition"
  } ],
  "message" : "Precondition check failed.",
  "status" : "FAILED_PRECONDITION"
}
```

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
@@ -68,8 +68,8 @@ public class ServiceAccountKeyManager {
           .execute();
       return true;
     } catch (GoogleJsonResponseException e) {
-      // TODO: handle 403 correctly once google fixes their API
-      if (e.getStatusCode() == 403 || e.getStatusCode() == 404) {
+      // TODO: handle 403 and 400 correctly once google fixes their API
+      if (e.getStatusCode() == 403 || e.getStatusCode() == 400 || e.getStatusCode() == 404) {
         return false;
       }
       throw e;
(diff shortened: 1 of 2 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/ServiceAccountKeyManager.java
```diff
@@ -68,8 +68,8 @@ public boolean keyExists(String keyName) throws IOException {
           .execute();
       return true;
     } catch (GoogleJsonResponseException e) {
-      // TODO: handle 403 correctly once google fixes their API
-      if (e.getStatusCode() == 403 || e.getStatusCode() == 404) {
+      // TODO: handle 403 and 400 correctly once google fixes their API
+      if (e.getStatusCode() == 403 || e.getStatusCode() == 400 || e.getStatusCode() == 404) {
```
honnix, line 72: It is more accurate to check status to be FAILED_PRECONDITION in case of 400 because 400 could mean different things.

tnsetting:
The issue seems to be resolved on google side. Let wait for another day to see it is really the case and we can close the PR.

sonjaer:
Closing since google fixed it

PR #922, opened 2021-03-23
Redo labels and annotations on FlyteAdminClientRunner

narape:

## Description
Now we send STYX_EXECUTION_ID in labels but also send styx
variables also parts of annotations

## Motivation and Context
In kubernetes world, Labels are meant for identifier info so STYX_EXECUTION_ID seems appropriate for that. 
However labels impose constrains not only on the key but also in the value. 

On the other hand. Annotations are mean to be used for user/tool specific metadata and also do not impose 
restriction on value format. So it seems appropriate to set all this STYX related variables.

I have included unit tests.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
@@ -22,12 +22,14 @@ package com.spotify.styx.flyte;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.spotify.styx.ScheduledExecutionUtil.scheduleWithJitter;
+import static com.spotify.styx.docker.LabelValue.normalize;
 import static com.spotify.styx.flyte.FlyteEventTranslator.translate;
 import static com.spotify.styx.util.CloserUtil.register;
 import static com.spotify.styx.util.GrpcContextUtil.currentContextExecutorService;
 import static com.spotify.styx.util.GuardedRunnable.guard;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.CompletableFuture.runAsync;
+import static java.util.stream.Collectors.toUnmodifiableMap;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
@@ -36,6 +38,7 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.spotify.styx.flyte.client.FlyteAdminClient;
 import com.spotify.styx.model.Event;
 import com.spotify.styx.model.FlyteExecConf;
+import com.spotify.styx.model.TriggerParameters;
 import com.spotify.styx.model.WorkflowInstance;
 import com.spotify.styx.state.RunState;
 import com.spotify.styx.state.StateData;
@@ -73,6 +76,7 @@ import org.slf4j.LoggerFactory;
 public class FlyteAdminClientRunner implements FlyteRunner {
 
   private static final Logger LOG = LoggerFactory.getLogger(FlyteAdminClientRunner.class);
+
   @VisibleForTesting static final String TERMINATE_CAUSE_PREFIX = "Styx workflow instance execution reached state: ";
   @VisibleForTesting static final String TERMINATE_CAUSE = "Styx workflow instance execution is not active";
   @VisibleForTesting static final String STYX_WORKFLOW_INSTANCE_ANNOTATION = "styx-workflow-instance";
@@ -84,6 +88,12 @@ public class FlyteAdminClientRunner implements FlyteRunner {
       .setDaemon(true)
       .setNameFormat("flyte-scheduler-thread-%d")
       .build();
+  private static final String STYX_COMPONENT_ID = "STYX_COMPONENT_ID";
+  private static final String STYX_WORKFLOW_ID = "STYX_WORKFLOW_ID";
+  private static final String STYX_PARAMETER = "STYX_PARAMETER";
+  private static final String STYX_EXECUTION_ID = "STYX_EXECUTION_ID";
+  private static final String STYX_TRIGGER_ID = "STYX_TRIGGER_ID";
+  private static final String STYX_TRIGGER_TYPE = "STYX_TRIGGER_TYPE";
 
   private final String runnerId;
   private final FlyteAdminClient flyteAdminClient;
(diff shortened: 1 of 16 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -129,20 +139,35 @@
...
+        .map(FlyteAdminClientRunner::getFilteredTriggerParams)
+        .orElseGet(Map::of);
+
+    var labels = Map.of(STYX_EXECUTION_ID, normalize(styxVariables.get(STYX_EXECUTION_ID)));
+    final var annotations = ImmutableMap.<String, String>builder()
+        .putAll(styxVariables)
+        // just to keep compatibility with removing dangling executions
+        // TODO: base dangling logic on STYX_WORKFLOW_ID, STYX_PARAMETER and STYX_EXECUTION_ID
+        //   and the remove these annotations
+        .put(STYX_WORKFLOW_INSTANCE_ANNOTATION, styxVariables.get(STYX_WORKFLOW_ID) + "#" + styxVariables.get(STYX_PARAMETER))
```
tnsetting, line 164: Would there be cases where `STYX_PARAMETER` in this format `2021-01-01T00:00:00`? Do we need to `normalize` the string as well?

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -129,20 +139,35 @@
...
         .filter(mode -> mode != ExecutionMode.UNRECOGNIZED)
         .orElseThrow(() -> new CreateExecutionException("Missing trigger or unknown in StateData: " + runState.data()));
 
-    var extraDefaultInputs = getExtraDefaultInputs(runState.workflowInstance(), runState.data());
+    var styxVariables = getStyxVariables(runState.workflowInstance(), runState.data());
+    var triggeredParams = runState.data().triggerParameters()
+        .map(FlyteAdminClientRunner::getFilteredTriggerParams)
+        .orElseGet(Map::of);
+
+    var labels = Map.of(STYX_EXECUTION_ID, normalize(styxVariables.get(STYX_EXECUTION_ID)));
```
tnsetting, line 158: I guess `normalize` function won't do anything so we will get the same string back, right?

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -129,20 +139,35 @@
...
+    final var annotations = ImmutableMap.<String, String>builder()
+        .putAll(styxVariables)
+        // just to keep compatibility with removing dangling executions
+        // TODO: base dangling logic on STYX_WORKFLOW_ID, STYX_PARAMETER and STYX_EXECUTION_ID
+        //   and the remove these annotations
+        .put(STYX_WORKFLOW_INSTANCE_ANNOTATION, styxVariables.get(STYX_WORKFLOW_ID) + "#" + styxVariables.get(STYX_PARAMETER))
+        .put(STYX_EXECUTION_ID_ANNOTATION, styxVariables.get(STYX_EXECUTION_ID))
```
tnsetting, line 165: Should be also normalize the `id` to be consistent with line 164?

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -129,20 +139,35 @@
...
 
-    var extraDefaultInputs = getExtraDefaultInputs(runState.workflowInstance(), runState.data());
+    var styxVariables = getStyxVariables(runState.workflowInstance(), runState.data());
+    var triggeredParams = runState.data().triggerParameters()
+        .map(FlyteAdminClientRunner::getFilteredTriggerParams)
+        .orElseGet(Map::of);
+
+    var labels = Map.of(STYX_EXECUTION_ID, normalize(styxVariables.get(STYX_EXECUTION_ID)));
+    final var annotations = ImmutableMap.<String, String>builder()
+        .putAll(styxVariables)
```
tnsetting, line 160: Do we need to normalize the variables as well?

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -336,29 +362,29 @@ private boolean isFlyteExecRelatedToRunState(AnnotatedFlyteExecutionId annotated
...
+    triggerId.ifPresent(value -> mapBuilder.put(STYX_TRIGGER_ID, value));
+    triggerType.ifPresent(value -> mapBuilder.put(STYX_TRIGGER_TYPE, value));
 
     return mapBuilder.build();
   }
 
+  // returns trigger parameters but avoid any STYX_XXX ones to prevent collisions
+  private static Map<String, String> getFilteredTriggerParams(TriggerParameters triggerParameters) {
+    return triggerParameters.env().entrySet().stream()
+                .filter(entry -> !entry.getKey().startsWith("STYX_"))
```
tnsetting, line 384: Do you think it make sense to convert the key to lower case and check start with `styx_`

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: @tnsetting actually annotation values don't have a restriction on the format ([see](https://github.com/spotify/styx/pull/922)). Annotation keys they do have restrictions and in the past I mistakenly missread the restrictions for the keys as the restrictions for the values.

Following the link you will see an example that they use `"https://hub.docker.com/"` as value and you can notice `:` and `/`

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: Yes I can use just the string and add a comment that normalize in this case is not necessary because the way we generate styx_execution ids. I can add a test to make sure that we don't break that invariant

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: Same answer here, it is not needed

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
narape: Good point, I can do that

styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteAdminClientRunner.java
```diff
@@ -336,29 +363,29 @@ private boolean isFlyteExecRelatedToRunState(AnnotatedFlyteExecutionId annotated
...
+    stateData.executionId().ifPresent(value -> mapBuilder.put(STYX_EXECUTION_ID, value));
+    triggerId.ifPresent(value -> mapBuilder.put(STYX_TRIGGER_ID, value));
+    triggerType.ifPresent(value -> mapBuilder.put(STYX_TRIGGER_TYPE, value));
 
     return mapBuilder.build();
   }
 
+  // returns trigger parameters but avoid any STYX_XXX ones to prevent collisions
+  private static Map<String, String> getFilteredTriggerParams(TriggerParameters triggerParameters) {
+    return triggerParameters.env().entrySet().stream()
```
tnsetting, line 384: Something not related to this PR but it looks we are using the environment variables i the `triggerParameters`. This is fine for now but it looks a bit confusion. Maybe this is something we need to re-evaluate later on.

tnsetting (approved):

PR #925, opened 2021-03-25
[WIP] Adding ActionAuhoriser class

Elisiac:
This pull requests adds an ActionAuthorizer class that can be used to implement an extra authorization system for destructive action perform on workflow though POST, DELETE AND PATCH api commands.
The default class provides a NOP class that does... well... nothing as expected but allow users to provide their own implementation for it.


Elisiac:
Pushed the PR from upstream directly instead of a fork

PR #930, opened 2021-03-26
Adding ActionAuhoriser class

Elisiac:
This pull requests adds an ActionAuthorizer class that can be used to implement an extra authorization system for destructive action perform on workflow though POST, DELETE AND PATCH api commands.
The default class provides a NOP class that does... well... nothing as expected but allow users to provide their own implementation for it.

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
@@ -132,6 +132,7 @@ public final class WorkflowResource {
           throw new ResponseException(response);
         }
         var workflow = workflowOpt.orElseThrow();
+        workflowActionAuthorizer.authorizeDeleteWorkflowAction(workflow);
         workflowActionAuthorizer.authorizeWorkflowAction(ac, workflow);
         tx.deleteWorkflow(workflowId);
         return workflow;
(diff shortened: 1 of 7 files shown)


styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -161,6 +162,7 @@ public WorkflowResource(Storage storage, WorkflowValidator workflowValidator,
 
     final Workflow workflow = Workflow.create(componentId, workflowConfig);
 
+    workflowActionAuthorizer.authorizeCreateOrUpdateWorkflowAction(workflow);
```
tnsetting, line 165: I am not sure about the double actions as we invoke the `workflowActionAuthorizer` object twice with `authorizeCreateOrUpdateWorkflowAction` and `authorizeWorkflowAction`. Would it be simpler to just call `authorizeCreateOrUpdateWorkflowAction` with inputs `ac` and `workflow` and have the logic for normal `serviceAccountUsageAuthorizer` and `actionAuthorizer`?

Elisiac: @tnsetting  I actually though about it. From a first glance it make sense to refactor the call directly into the `workflowActionAuthorizer` but the `authorizeWorkflowAction` method as it is written today is identical for all 3 calls (Create/update, Patch and Delete). Thus, one need to either use a specific function (like I did) or pass a hardcoded argument into the `authorizeWorkflowAction` function to indicate which extra action to take for the call. In the sake of clarity I would instead rename the `authorizeWorkflowAction` method to something more explicit like `authorizeUserForWorkflow`.
That way we have clear distinction between authorizing the user itself / service account on one side and the action taken on the other.

ckiosidis (approved):

ckiosidis (changes requested):
could you refactor the NOP to NOOP for consistency?

ckiosidis (approved):

PR #949, opened 2021-05-23
[Hotfix] Temporarily disable key deletion

honnix:
# Hey, I just made a Pull Request!

## Description

This PR temporarily disables key deletion when the key(s) are found in k8s secret but cannot verified to exist in GCP.

## Motivation and Context

This is a hotfix to work around a possible issue. **Note** that this PR should be reverted later.

## Have you tested this? If so, how?
 

Unit test

## Checklist for PR author(s)

- [x] Changes are covered by unit test
- [ ] All tests pass
- [ ] Code coverage check passes
- [ ] Error handling is tested
- [ ] Errors are handled at the appropriate layer
- [ ] Errors that cannot be handled where they occur are propagated
- [ ] (optional) Changes are covered by system test
- [ ] Relevant documentation updated
- [ ] This PR has NO breaking change to public API
- [ ] This PR has breaking change to public API and it is documented

## Checklist for PR reviewer(s)

- [ ] This PR has been incorporated in release note for the coming version
- [ ] Risky changes introduced by this PR have been all considered

PR #953, opened 2021-05-27
Revert "[Hotfix] Temporarily disable key deletion"

honnix:
Reverts spotify/styx#949

Wait until we can merge to avoid having a non deployable main branch.


tnsetting (approved):

honnix:
@tnsetting We will need to revert https://github.com/spotify/styx/pull/951 as well.

PR #967, opened 2021-12-14
parse config env-vars and inject them to docker container

ckiosidis:

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
@@ -129,6 +129,7 @@ import java.util.concurrent.ThreadFactory;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import okhttp3.OkHttpClient;
 import okhttp3.Protocol;
 import org.apache.hadoop.hbase.client.Connection;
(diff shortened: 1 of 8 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java
```diff
@@ -152,6 +153,8 @@
   private static final String STYX_STATE_MANAGER_TICK_INTERVAL = "styx.state-manager.tick-interval";
   private static final String STYX_SCHEDULER_THREADS = "styx.scheduler-threads";
   private static final String STYX_ENVIRONMENT = "styx.environment";
+  private static final String STYX_EXECUTION_ENV_VARS = "styx.execution-env-vars";
```
narape, line 156:
```suggestion
  private static final String STYX_EXECUTION_ENV_VARS = "styx.XXX.execution-env-vars";
```
I think that it would be good to add some prefix, like docker or similar to signal that this is a config not related to flyte executions

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -185,15 +187,18 @@
         register(closer, Objects.requireNonNull(scheduledExecutor), "kubernetes-scheduled-executor");
     this.executor = currentContextExecutorService(
         register(closer, new ForkJoinPool(K8S_POD_PROCESSING_THREADS), "kubernetes-executor"));
+    this.executionEnvVars = Objects.requireNonNull(executionEnvVars);
```
narape, line 190:
```suggestion
    this.executionEnvVars = Objects.requireNonNull(executionEnvVars, "executionEnvVars");
```
we get a more descriptive error message this way

styx-scheduler-service/src/main/java/com/spotify/styx/docker/KubernetesDockerRunner.java
```diff
@@ -370,6 +377,13 @@ static EnvVar envVar(String name, String value) {
     env.put(TRIGGER_TYPE, runSpec.trigger().map(TriggerUtil::triggerType).orElse(null));
     env.put(ENVIRONMENT, styxEnvironment);
     env.put(LOGGING, "structured");
+    executionEnvVars.forEach((key, value) -> {
+      if (env.containsKey(key)) {
+        LOG.error("Key already exists in execution environment variables {}. Key will be skipped", key);
```
narape, line 382:
```suggestion
        LOG.info("Key already exists in execution environment variables {}. Key will be skipped", key);
```
we should only log error if we are going to avoid or fail the execution, but here we are just ignoring and continuing

PR #1000, opened 2022-08-02
Check role first before checking admin

honnix:

## Description

Check role before checking administrators.

Also a minor fix to omit logging stacktrace for info level that was not covered by #999.

## Motivation and Context

In most cases the principle would not be configured administrators, so the group member check would fail or return false, which is wasteful. Also the directory API call could be slow for unknown reason.

By swapping the two checks we should be able to save many unnecessary API calls.

Unit tests get this covered.

## Checklist for PR author(s)

- [x] All tests pass


narape (approved):

honnix:
There is a bad regression because there are cases `checkRole` would throw, for example a service account no longer exists. So the problem is for those cases, even admins cannot do anything.

PR #1002, opened 2022-08-05
Backfills start filter

ckiosidis:
 
## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
@@ -31,7 +31,6 @@ import static com.spotify.styx.util.TimeUtil.nextInstant;
 import static java.util.stream.Collectors.toList;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.google.cloud.Timestamp;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Streams;
 import com.google.common.io.Closer;
@@ -61,6 +60,7 @@ import com.spotify.styx.state.RunState;
 import com.spotify.styx.state.StateData;
 import com.spotify.styx.storage.Storage;
 import com.spotify.styx.storage.StorageTransaction;
+import com.spotify.styx.util.ParameterUtil;
 import com.spotify.styx.util.RandomGenerator;
 import com.spotify.styx.util.ReplayEvents;
 import com.spotify.styx.util.ResourceNotFoundException;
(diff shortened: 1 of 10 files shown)


styx-api-service/src/main/java/com/spotify/styx/api/BackfillResource.java
```diff
@@ -164,12 +165,13 @@ private BackfillsPayload getBackfills(RequestContext rc) {
     final Optional<String> workflowOpt = rc.request().parameter("workflow");
     final boolean includeStatuses = rc.request().parameter("status").orElse("false").equals("true");
     final boolean showAll = rc.request().parameter("showAll").orElse("false").equals("true");
+    final Instant start = rc.request().parameter("start").map(ParameterUtil::parseDate).orElse(Instant.MIN);
```
narape, line 168: Can you add a test that shows that we pick the value from the request and also that we use Instant.Min when it is not in the request

styx-client/src/main/java/com/spotify/styx/client/StyxBackfillClient.java
```diff
@@ -133,7 +134,8 @@ CompletionStage<Backfill> backfillEditConcurrency(String backfillId,
   CompletionStage<BackfillsPayload> backfillList(Optional<String> componentId,
                                                  Optional<String> workflowId,
                                                  boolean showAll,
-                                                 boolean includeStatus);
+                                                 boolean includeStatus,
+      Optional<Instant> start);
```
honnix, line 138: This change is not backward compatible from API perspective. We will either need to help existing users of this call migrate, if there is any; or add a new method.

narape: Agreed, you could:
* Add `@Deprecated` to this method
* Create a new overloaded method with the new parameter
* The `StyxOkHttpClient` class that implements this interface could then 
```java
  @Override
  public CompletionStage<BackfillsPayload> backfillList(Optional<String> componentId,
                                                        Optional<String> workflowId,
                                                        boolean showAll,
                                                        boolean includeStatus) {
    return(componentId, workflowId, showAll, includeStatus, Optional.empty());
  }
```

Please also follow the indentation scheme already present on the file and if a class/interface has a javadoc, you should add also a javadoc of the new method/parameter that your are adding

max-pavoni: @narape thanks, that makes sense. But shouldn't we also add a new overloaded method inside `StyxOkHttpClient`? Since it also implements `StyxClient`.

narape (approved):

PR #1021, opened 2023-02-01
add trigger id to trigger response

benkonz:
adds trigger id to trigger response so that consumers of the trigger endpoint can track a trigger's execution. I wasn't able to run the unit tests due to a permission issue, however the tests up to that point all passed. Just to confirm, could whoever reviews this verify the tests still pass?

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
@@ -41,6 +41,7 @@ import com.spotify.styx.api.Middlewares.AuthContext;
 import com.spotify.styx.model.Event;
 import com.spotify.styx.model.TriggerParameters;
 import com.spotify.styx.model.TriggerRequest;
+import com.spotify.styx.model.TriggerResponse;
 import com.spotify.styx.model.Workflow;
 import com.spotify.styx.model.WorkflowInstance;
 import com.spotify.styx.serialization.Json;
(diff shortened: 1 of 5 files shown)


styx-scheduler-service/src/main/java/com/spotify/styx/api/SchedulerResource.java
```diff
@@ -232,9 +233,15 @@ private Response<TriggerRequest> triggerAndWait(TriggerRequest triggerRequest,
     } catch (Exception e) {
       return handleException(e);
     }
-
-    // todo: change payload to a struct returning the triggerId as well so the user can refer to it
-    return Response.forPayload(triggerRequest);
+    TriggerResponse response;
+    if (triggerRequest.triggerParameters().isPresent()) {
```
benkonz, line 237: would calling `TriggerResponse.of(triggerRequest.workflowId(), triggerRequest.parameter(),
       parameters, triggerId);` here be sufficient? Would a null `triggerParametres()` and `TriggerParameters.zero()` output the same?

bergman: From the looks of the generated code for the automatter builder, passing `null` should behave the same way as not calling the setter on the builder (`Optional.empty()`), so something like this should work:

```java
TriggerResponse response =
    TriggerResponse.of(triggerRequest.workflowId(), triggerRequest.parameter(),
        triggerRequest.triggerParameters().orElse(null), triggerId);
``` 

If you tweaked the `of` factory method to accept `Optional<TriggerParameters>` you could even pass it directly through. I'm not a big fan of `Optional` for parameters though. I don't know. What do you think? :)

`TriggerParameters.zero()` is an internal thing and should probably not be passed either in or out, so better to go with `Optional.empty()`.

bergman: I will try to, but otherwise will bring to the rest of the team. It's been a few years since I last worked on Styx 😬

benkonz: thanks!!

bergman: `mvn test` at least runs fine on my machine without errors, but need to verify if that runs all the tests. Will try to look into that tomorrow.

narape:
Hi @benkonz, our CI/CD doesn't work for forks (something with permissions). I have added you as collaborator for this repo, so you can move your branch from your forked repo to this repo, so the build could happen

cc @bergman

benkonz:
@narape @bergman I opened up a new PR on top of master instead of a fork https://github.com/spotify/styx/pull/1024, we can continue the discussion there, but it still seems like the CI isn't running?

PR #1024, opened 2023-02-08
Add trigger id to trigger response

benkonz:
adds trigger id to trigger response so that consumers of the trigger endpoint can track a trigger's execution for the /trigger POST api endpoint

### DIFF
--- a/styx-common/src/main/java/com/spotify/styx/model/TriggerResponse.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/TriggerResponse.java
@@ -1,3 +1,23 @@
+/*-
+ * -\-\-
+ * Spotify Styx Common
+ * --
+ * Copyright (C) 2016 - 2023 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
 package com.spotify.styx.model;
 
 import io.norberg.automatter.AutoMatter;
--- a/styx-common/src/main/java/com/spotify/styx/model/TriggerResponse.java
+++ b/styx-common/src/main/java/com/spotify/styx/model/TriggerResponse.java
@@ -1,0 +1,32 @@
+package com.spotify.styx.model;
+
+import io.norberg.automatter.AutoMatter;
+
+@AutoMatter
+public interface TriggerResponse extends TriggerRequest {
+  String triggerId();
+
+
+
+  static TriggerResponseBuilder builder() {
+    return new TriggerResponseBuilder();
+  }
+
+  static TriggerResponse of(WorkflowId workflowId, String parameter, String triggerId) {
+    return builder()
+        .workflowId(workflowId)
+        .parameter(parameter)
+        .triggerId(triggerId)
+        .build();
+  }
+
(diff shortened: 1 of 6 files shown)


benkonz:
@narape @bergman tomorrow, when you get the chance, could you give this a look?

styx-common/src/main/java/com/spotify/styx/model/TriggerResponse.java
```diff
@@ -0,0 +1,42 @@
...
+
+package com.spotify.styx.model;
+
+import io.norberg.automatter.AutoMatter;
+
+@AutoMatter
+public interface TriggerResponse extends TriggerRequest {
+  String triggerId();
+
+  static TriggerResponseBuilder builder() {
```
honnix, line 29: This is not used outside of this class, so maybe remove it?

styx-common/src/main/java/com/spotify/styx/model/TriggerResponse.java
```diff
@@ -0,0 +1,42 @@
...
+    return new TriggerResponseBuilder();
+  }
+
+  static TriggerResponse of(WorkflowId workflowId, String parameter,
```
honnix, line 33: Maybe a factory method `TriggerResponseBuilder from(TriggerRequest triggerRequest)` could be more useful. A small thing, doesn't matter.

styx-scheduler-service/src/test/java/com/spotify/styx/api/SchedulerResourceTest.java
```diff
@@ -335,13 +336,17 @@ public void testTriggeredWorkflowGeneratesTrigger() throws Exception {
...
+        OBJECT_MAPPER.readValue(response.payload().get().toByteArray(), TriggerResponse.class);
 
     assertThat(response.status(), is(Status.OK));
     final Instant expectedInstant = Instant.parse("2014-12-31T23:00:00.000Z");
     verify(triggerListener).event(eq(HOURLY_WORKFLOW), triggerCaptor.capture(), eq(expectedInstant), eq(expectedParameters));
-    assertThat(TriggerUtil.triggerType(triggerCaptor.getValue()), is("adhoc"));
-    assertThat(TriggerUtil.triggerId(triggerCaptor.getValue()), startsWith("ad-hoc-cli-"));
+    final Trigger trigger = triggerCaptor.getValue();
+    final String triggerId = TriggerUtil.triggerId(trigger);
+    assertThat(TriggerUtil.triggerType(trigger), is("adhoc"));
```
honnix, line 347: Maybe taking the opportunity to change `assertThat` to `import static org.hamcrest.MatcherAssert.assertThat;`? `org.junit.Assert.assertThat` is deprecated.

honnix (commented):
Minor comments. I don't remember how the client and CLI lib work, have you checked whether we need to change anything there?

honnix:
> Minor comments. I don't remember how the client and CLI lib work, have you checked whether we need to change anything there?

NVM I see we don't return anything from client lib.

bergman:
> > Minor comments. I don't remember how the client and CLI lib work, have you checked whether we need to change anything there?
> 
> NVM I see we don't return anything from client lib.

Yeah, I had a look and did not find any, so this will mainly affect any direct integration towards to API. Hope they can accept unknown fields in their deserialisation 😬

styx-common/src/main/java/com/spotify/styx/model/TriggerResponse.java
bergman: I think I would rather not leak the builder, but it's not a strong objection 🤷‍♀️

honnix (approved):

styx-scheduler-service/src/test/java/com/spotify/styx/api/SchedulerResourceTest.java
bergman: There are ~60 other uses of this import. I fixed this one and can fix the rest in another PR.

honnix: Yeah I was only referring the file we anyway need to touch in this PR. The rest can wait.

PR #1025, opened 2023-02-08
Increase upper bound

sonjaer:

We have users that want more than 24h due to an ML pipeline running for a long time that doesn't make sense to split.
Ideally this should be made configurable, but we can separate the upper bound from the max value.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-api-service/src/main/java/com/spotify/styx/StyxApi.java
+++ b/styx-api-service/src/main/java/com/spotify/styx/StyxApi.java
@@ -84,7 +84,9 @@ public class StyxApi implements AppInit {
   private static final String DEFAULT_SCHEDULER_SERVICE_BASE_URL = "http://localhost:8080";
 
   private static final String STYX_RUNNING_STATE_TTL_CONFIG = "styx.stale-state-ttls.running";
+
   private static final Duration DEFAULT_STYX_RUNNING_STATE_TTL = Duration.ofHours(24);
+  private static final Duration MAX_STYX_RUNNING_STATE_TTL = Duration.ofHours(30);
 
   private final String serviceName;
   private final StorageFactory storageFactory;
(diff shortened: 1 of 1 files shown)


honnix:
I'm not sure I understand this change. This is already configurable, what's the purpose of the max value?

sonjaer:
> I’m not sure I understand this change. This is already configurable, what’s the purpose of the max value?

I was going to look at this more tomorrow so its not ready. Basically wanted to allow users to set more than 24h timeout from the api but still keep the default timeout of 24h for those that don't set.

honnix:
`running_timeout` is added for this customization purpose, only that it cannot pass the max value (24h) due to API validation as you have figured out. If we want to allow that, we can stop validating the value because at execution time it is not checked (https://github.com/spotify/styx/blob/4a7175c1ae3d3715907e4bf63e99436dbe11e28c/styx-scheduler-service/src/main/java/com/spotify/styx/state/StateUtil.java#L118-L122). Adding yet another max value might be confusing, and if we do want to take this path, this must be configurable.

This is a somewhat a big decision to make, so let's agree first before doing more implementation. Today we have users hitting 24h limit, and tomorrow we might have those hitting 30h limit.

sonjaer:
how come the default becomes 24h hours if running_timeout is unset? Where is that code?

sonjaer:
> `running_timeout` is added for this customization purpose, only that it cannot pass the max value (24h) due to API validation as you have figured out. If we want to allow that, we can stop validating the value because at execution time it is not checked (
> 
> https://github.com/spotify/styx/blob/4a7175c1ae3d3715907e4bf63e99436dbe11e28c/styx-scheduler-service/src/main/java/com/spotify/styx/state/StateUtil.java#L118-L122
> 
> ). Adding yet another max value might be confusing, and if we do want to take this path, this must be configurable.
> This is a somewhat a big decision to make, so let's agree first before doing more implementation. Today we have users hitting 24h limit, and tomorrow we might have those hitting 30h limit.

Thanks for explaning!

how come the default becomes 24h hours if running_timeout is unset? Where is that code?

sonjaer:
> This is a somewhat a big decision to make, so let's agree first before doing more implementation. Today we have users hitting 24h limit, and tomorrow we might have those hitting 3

Thanks for explaining! But so where does that default timeout comes from if the running_timeout is unset that scheduler uses terminate workflows after 24h, i.e.   the `timeout` variable from the code you linked?

honnix:
> > This is a somewhat a big decision to make, so let's agree first before doing more implementation. Today we have users hitting 24h limit, and tomorrow we might have those hitting 3
> 
> Thanks for explaining! But so where does that default timeout comes from if the running_timeout is unset that scheduler uses terminate workflows after 24h, i.e. the `timeout` variable from the code you linked?

Yes, that is correct, and that value comes from https://github.com/spotify/styx/blob/a257d290c52f5ebf41b38066aac0b593f34ce2ff/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java#L412-L413.

PR #1057, opened 2023-03-13
Update API docs after new endpoint was added

sonjaer:

## Description
Update API docs after new endpoint was added


.circleci/config.yml
```diff
@@ -0,0 +1,2 @@
+
+workflows
```
sonjaer, line 2:
```suggestion
```

.circleci/config.yml
narape, line 1: What is this file doing? why is needed? and why a no-op job?

.circleci/config.yml
```diff
@@ -0,0 +1,15 @@
+version: 2.1
+jobs:
+  no-op:
+    docker:
+      - image: circleci/python:3.9
```
sonjaer, line 5:
```suggestion
      - image: non-existing-image
```

.circleci/config.yml
```diff
@@ -0,0 +1,15 @@
+version: 2.1
+jobs:
+  no-op:
+    docker:
+      - image: non-existing-image
```
sonjaer, line 5:
```suggestion
      - image: circleci/python:3.9
```

.circleci/config.yml
sonjaer: this is for github pages to show api html doc, and we don't care about running the build here

circle ci complains if the file is missing or empty so this is the only way I found works

narape (approved):

PR #1040, opened 2023-03-03
Fix some snyk reports

narape:
# Hey, I just made a Pull Request!

## Description
Bump libraries spotted by snyk
Here are the same versions as  
* https://github.com/spotify/styx/pull/1036
* https://github.com/spotify/styx/pull/1032
* https://github.com/spotify/styx/pull/1031
* https://github.com/spotify/styx/pull/1030
* https://github.com/spotify/styx/pull/1029

Plus enforcer and error prone pleasing

## Motivation and Context
Prevent bit rot and address possible security issues

## Have you tested this? If so, how?
Same test suit

## Checklist for PR author(s)

- [X] Changes are covered by unit test
- [X] All tests pass
- [X] Code coverage check passes
- [ ] Error handling is tested
- [ ] Errors are handled at the appropriate layer
- [ ] Errors that cannot be handled where they occur are propagated
- [ ] (optional) Changes are covered by system test
- [ ] Relevant documentation updated
- [ ] This PR has NO breaking change to public API
- [ ] This PR has breaking change to public API and it is documented

## Checklist for PR reviewer(s)

- [ ] This PR has been incorporated in release note for the coming version
- [ ] Risky changes introduced by this PR have been all considered

PR #1061, opened 2023-03-20
Bump grpc and protobuf libs

honnix:

## Description

Bump grpc and protobuf libs.

## Motivation and Context

Bump those versions so we can compile on Apple silicon.

Note that this is absolutely not the ideal change because of the ever-growing version override. Maybe we should switch to https://github.com/googleapis/java-cloud-bom at certain point.

Existing tests.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,12 +27,13 @@
     <metrics.version>4.2.15</metrics.version>
     <opencensus.version>0.28.0</opencensus.version>
     <google.client.lib.version>1.30.8</google.client.lib.version>
-    <google.http.client.lib.version>1.34.2</google.http.client.lib.version>
+    <google.http.client.lib.version>1.41.0</google.http.client.lib.version>
     <auto-matter.version>0.26.1</auto-matter.version>
     <auto-value.version>1.10.1</auto-value.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -649,11 +650,6 @@
         <artifactId>error_prone_annotations</artifactId>
         <version>${errorprone.version}</version>
       </dependency>
-      <dependency>
-        <groupId>com.google.protobuf</groupId>
```
honnix, line 653: This is covered by `protobuf-bom`.

honnix: This was added in https://github.com/spotify/styx/pull/1040. Maybe unintentionally? @narape

pom.xml
```diff
@@ -674,6 +670,26 @@
         <artifactId>checker-qual</artifactId>
         <version>3.19.0</version>
       </dependency>
+      <dependency>
```
honnix, line 673: All about version resolution as enforced by the enforcer.

honnix:
This is not going to work well without spending a lot of time figuring out the correct combination of deps. We will need to come back to this later. So for now #1062 seems to be more reasonable to unblock us.

PR #1056, opened 2023-03-10
Add endpoint with enabled state

sonjaer:
# Hey, I just made a Pull Request!

## Description

There is no endpoint for quickly getting all workflows enabled state. This PR adds it.

## Motivation and Context

Currently you need to make one request per workflow which takes a long time. Since we are making some changes internally it would be good to store this information to be able to recover it.

## Have you tested this? If so, how?
I have included unit tests

## Checklist for PR author(s)

- [x] Changes are covered by unit test
- [x] All tests pass
- [x] Code coverage check passes
- [x] Error handling is tested
- [x] Errors are handled at the appropriate layer
- [x] Errors that cannot be handled where they occur are propagated
- [ ] (optional) Changes are covered by system test
- [ ] Relevant documentation updated
- [x] This PR has NO breaking change to public API
- [x] This PR has breaking change to public API and it is documented

## Checklist for PR reviewer(s)

- [ ] This PR has been incorporated in release note for the coming version
- [ ] Risky changes introduced by this PR have been all considered

PR #1065, opened 2023-03-26
Add param to endpoint

sonjaer:

Reverts/changes https://github.com/spotify/styx/pull/1056 

## Description

There are two endpoints now:

* /workflows/full
* /workflows/

If a component is named full it would collide with the endpoint getting workflow states. This PR changes the path of the endpoint to be:

*  /workflows?full=true

## Alt
* https://github.com/spotify/styx/pull/1064

## Checklist for PR author(s)

- [x] All tests pass


doc/api.apib
```diff
@@ -48,7 +48,9 @@ Resources related to the Workflows registered in Styx.
     + version: `v3` (enum[string]) - API version
         + Members
             + `v3`
+    + full: `true` (optional, boolean) - Return the state of the workflow
 
+#### Without full
```
sonjaer, line 53:
```suggestion
#### Without ?full=true
```

doc/api.apib
```diff
@@ -48,7 +48,9 @@ Resources related to the Workflows registered in Styx.
     + version: `v3` (enum[string]) - API version
         + Members
             + `v3`
+    + full: `true` (optional, boolean) - Return the state of the workflow
 
+#### Without ?full=true
```
sonjaer, line 53:
```suggestion
#### Without `full`
```

styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
```diff
@@ -830,7 +830,7 @@ public void shouldReturnWorkflowsWithState() throws Exception {
     sinceVersion(Api.Version.V3);
 
     Response<ByteString> response = awaitResponse(
-            serviceHelper.request("GET", path("/full")));
+            serviceHelper.request("GET", path("?full=true")));
```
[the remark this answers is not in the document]
sonjaer: yes it is this test case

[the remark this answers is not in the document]
sonjaer: do you mean if we have without states? That one is called shouldReturnWorkflows

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -219,31 +217,32 @@ private WorkflowConfiguration readFromJsonWithDefaults(ByteString payload)
...
 
+      List<Workflow> workflows = filteredWorkflows.stream().map(WorkflowWithState::workflow).collect(Collectors.toList());
+      return Response.forPayload(workflows);
     } catch (IOException e) {
       throw new RuntimeException("Failed to get workflows", e);
     }
   }
 
+  private boolean shouldIncludeState(Optional<String> includeStates) {
+    return includeStates.filter(Boolean::parseBoolean).isPresent();
```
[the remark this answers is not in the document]
sonjaer: it is not sufficient to just have ?full, I added a test to show it here https://github.com/spotify/styx/pull/1065/files#diff-a92738177ebe75548687fdc45ffa62a2b696244ac699584214fa96cb7e14cf35R823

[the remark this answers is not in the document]
honnix: It currently does not work because of `Boolean::parseBoolean` I guess. But we could make it work, and have a value-less query parameter in this case seems to be more reasonable than having `full=true` or `full=false`.

styx-api-service/src/main/java/com/spotify/styx/api/WorkflowResource.java
```diff
@@ -219,31 +217,32 @@ private WorkflowConfiguration readFromJsonWithDefaults(ByteString payload)
...
 
+      List<Workflow> workflows = filteredWorkflows.stream().map(WorkflowWithState::workflow).collect(Collectors.toList());
+      return Response.forPayload(workflows);
     } catch (IOException e) {
       throw new RuntimeException("Failed to get workflows", e);
     }
   }
 
+  private boolean includeStates(Request request) {
+    Optional<String> full = getFilterParams(request, INCLUDE_STATES);
```
honnix, line 242: `getFilterParams(request, INCLUDE_STATES).isPresent()` should work without checking the actual value.

styx-api-service/src/test/java/com/spotify/styx/api/WorkflowResourceTest.java
```diff
@@ -814,27 +815,39 @@ public void shouldFailInvalidWorkflowImage() throws Exception {
...
   @Test
   public void shouldReturnWorkflows() throws Exception {
     sinceVersion(Api.Version.V3);
-
-    Response<ByteString> response = awaitResponse(
-        serviceHelper.request("GET", path("")));
-
-    assertThat(response, hasStatus(withCode(Status.OK)));
-    assertJson(response, "[*]", hasSize(2));
+    var paths = List.of("", "?full", "?full=false");
```
honnix, line 823: Please use parameterized test instead of a for loop.

honnix: For this rather old code base, we have been using https://github.com/spotify/styx/blob/4a7175c1ae3d3715907e4bf63e99436dbe11e28c/styx-scheduler-service/src/test/java/com/spotify/styx/docker/LabelValueTest.java#L122-L126 instead.

sonjaer: ok I couldn't make it work with parametrized test, thought it was due to the parent class already using parameters, but will try that

sonjaer: it gives me:

```
org.junit.runners.model.InvalidTestClassError: Invalid test class 'com.spotify.styx.api.WorkflowResourceTest':
  1. Method shouldReturnWorkflows should have no parameters
  ```

sonjaer: doesn't make sense to have multiple parameters anymore after changing full=true to full

honnix (approved):

PR #1084, opened 2023-05-24
Fix flyte error translation

### LINKED ISSUE #1083: 🐛  BUGFIX: Handle Flyte error codes that come from Dynamic Workflows
# Description 
While working with the deployment of the dynamic workflow, it was found that the workflow would result return an error code of `RetriesExhaused|User:NotReady` when there was a dependency missing instead of `User:NotReady`, which is the typical error code when a dependency was missing. Styx uses these error codes returned by Flyte to determine what status the Styx workflow instance should be, and if it is User:NotReady, the system will return a 20 error code for a missing dependency. ([relevant code](https://github.com/spotify/styx/blob/69aa2e7b707d5b7e5e79ab4119667813fe45ea66/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteEventTranslator.java#L44)) With the Flyte team's help, the issue could be tracked to a set of locations in the Flyte propeller code. 

# Ideal Behavior 
Styx returns a Missing Dependencies error code when the error code contains `User:NotReady` 

# Current Behavior 
Styx returns an unknown error error code when the error code is not exactly `User:NotReady`

# Possible Cause 
Within the Flyte propeller code base, it was found that dynamic workflows will raise a RetryableFailure status if any dynamically generated nodes fail ([relevant code](https://github.com/flyteorg/flytepropeller/blob/33ca38cf0c1c311e4b2eb46d803a92c95e8c103f/pkg/controller/nodes/dynamic/handler.go#L206)). Once this status is raised for the dynamic workflow, the Flyte propeller will prepend the error code with `RetriesExhaused|` before the dynamic node's original error code ([relevant code](https://github.com/flyteorg/flytepropeller/blob/33ca38cf0c1c311e4b2eb46d803a92c95e8c103f/pkg/controller/nodes/executor.go#L494)).

The impact is that any dynamic workflow cannot raise a `User:NotReady` in a way Styx can identify. This will result in erroneously labeling workflows as having unknown errors when the team may be raising error codes known to the Styx service but not recognized due to the RetriesExhaused string prepended to it.

# Suggested Remediation 
A possible remediation to this to allow dynamic workflows to raise specific Styx errors is to remove the `RetriesExhaused|` String before matching it to any of the known error codes.

brandon-segal:

## Description
Add a step to clean error codes in the FlyteEventTranslator class of the RetriesExhaused prefix.

## Motivation and Context
fixes #1083 

Added Tests to the FlyteEventTranslator test case to include one item for each scenario with the added prefix to ensure that the code is behaving as expected

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteEventTranslator.java
+++ b/styx-scheduler-service/src/main/java/com/spotify/styx/flyte/FlyteEventTranslator.java
@@ -38,7 +38,7 @@ import org.slf4j.LoggerFactory;
 
 public class FlyteEventTranslator {
   private static final Logger LOG = LoggerFactory.getLogger(FlyteEventTranslator.class);
-
+  private static final String RETRIES_EXHAUSTED = "RetriesExhausted|";
   private static final String PERSISTED_CODE = "USER:Persisted";
   private static final String NOT_RETRYABLE_CODE = "USER:NotRetryable";
   private static final String NOT_READY_CODE = "USER:NotReady";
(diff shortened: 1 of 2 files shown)


andresgomezfrr:
Closed in favor of https://github.com/spotify/styx/pull/1085

